PR #3, opened 2017-05-04
Pipe stdout directly to the terminal when building

developius:
What it says on the tin but it essentially attaches the command's `stdout` to the terminal and does some funky referencing in case we get errors. This means you can see the command's output as it comes in, instead of waiting for it all in one go. Much nicer for slower `docker build`'s etc.

Tested with both failing and passing commands, all works as expected.

Thoughts? 💡🔎

### DIFF
--- a/app.go
+++ b/app.go
@@ -68,13 +68,17 @@ func main() {
 		fmt.Println(strings.Join(builder, " "))
 		targetCmd := exec.Command(builder[0], builder[1:]...)
 		targetCmd.Dir = tempPath
-		cmdOutput, cmdErr := targetCmd.CombinedOutput()
+		targetCmd.Stdout = os.Stdout
+
+		var stderr bytes.Buffer
+		targetCmd.Stderr = &stderr
+		targetCmd.Start()
+		cmdErr := targetCmd.Wait()
 
 		if cmdErr != nil {
-			fmt.Printf("Error: %s\n" + cmdErr.Error())
+			fmt.Printf("Error: %s\n", cmdErr)
+			fmt.Printf(stderr.String())
 		}
-
-		fmt.Println(string(cmdOutput))
 
 		fmt.Printf("Image: %s built.\n", image)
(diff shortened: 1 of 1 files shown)


app.go
```diff
@@ -68,14 +68,18 @@ func main() {
 		fmt.Println(strings.Join(builder, " "))
 		targetCmd := exec.Command(builder[0], builder[1:]...)
 		targetCmd.Dir = tempPath
-		cmdOutput, cmdErr := targetCmd.CombinedOutput()
+		targetCmd.Stdout = os.Stdout
+
+		var stderr bytes.Buffer
+		targetCmd.Stderr = &stderr
+		targetCmd.Start()
```
alexellis: Do you need to call Start()?

app.go
```diff
@@ -68,14 +68,18 @@ func main() {
...
+		cmdErr := targetCmd.Wait()
 
 		if cmdErr != nil {
-			fmt.Printf("Error: %s\n" + cmdErr.Error())
+			fmt.Printf("Error: %s\n", cmdErr)
+			fmt.Printf(stderr.String())
```
alexellis: Isn't this already piped to the screen? since targetCmd.stderr was connected to stderr?

alexellis:
Looks like a useful change, thanks.

app.go
developius: It's not piped, I only set up stdin to do that - I can change it to both if that would be better?

app.go
developius: Doesn't look like it

developius: Yeah, otherwise you get `Error: exec: not started`

alexellis:
If you could `git rebase HEAD~2 -s` then squash one of those commits down, that'll let you give a new commit message for both? `-s` will sign off the commit.

developius:
Resolved by 560c04f0531bb24bbed4fd05fd002918aed0c90c

PR #12, opened 2017-06-23
Use raw sample.yml URL in example

johnmccabe:
Prior to this commit the example URL was downloading the Github page that the `sample.yml` was on rather than the raw yml.


alexellis:
I think this is up to date now. Thanks for raising the fix 👍

PR #21, opened 2017-07-23
Added copyTree function and replaced the template copyFiles calls

rgee0:

Main Change:
Created a copyTree function to address [issue#18](https://github.com/alexellis/faas-cli/issues/18).
The function takes a similar approach to copyFiles but on finding a folder creates an equivalent folder in the build dir and then calls itself, this time passing in the source folder and newly created target folder.  This means that all files and folder to _n-levels_ will be copied.
A helper function has also been created; `pathExists()` is used within `copyTree` to try to offset errors where the destination may already exist.
Finally, replaced the two `copyFile` calls for the template folders with a single call to `copyTree` and amended the TODO comment accordingly. 

Test:
Created 10 levels of files and folders beneath a template folder
Ran `./faas-cli -action build -yaml ./samples.yml`
Confirmed that the images built successfully.
Confirmed that the expected files and folders were extant in the build folder following completion of build.

Other small edits:
Standardised some of the messages returned to the user.
Some diffs being shown are a result of vscode re-formatting.

Misc:
Used `git commit -s` as per the contribution guide

### DIFF
--- a/app.go
+++ b/app.go
@@ -27,6 +27,7 @@ import (
 
 const providerName = "faas"
 const defaultNetwork = "func_functions"
+
 var GitCommit string
 
 func main() {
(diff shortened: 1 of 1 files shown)


alexellis:
Thanks for this!

rgee0:
I left `copyFiles` in for situations where only files needed copying, so we have a choice:

Assume every copy will need to do the whole tree from its start path. Remove `copyFiles` and rename `copyTree`To `copyFiles`.

Leave `copyFiles` for situations where only the path's files require to be copied. Rename `copyTree` To `copyFilesTree`.

alexellis:
I was thinking either `copyFiles` which is always depth-recursive or `copyFiles` plus a flag `recursive` for when you don't want to go down the file-structure?

rgee0:
Cool. We'll go with the additional param since that alllows is greater control

alexellis:
Thanks for this 💥

Issue #23, opened 2017-07-29
Templating should use the --squash parameter for docker build


alexellis:
This will result in potentially smaller images.

alexellis:
Fixed in #24 - thanks a lot Richard!

PR #24, opened 2017-08-02
Added squash flag & associated fixes

rgee0:

### **Main Change:**
Added a -squash flag to address  #23.
Fixed issue where supplying -no-cache flag would cause builds to fail
Changed http_proxy test to include a test for https_proxy (typo in the OR)
Refactored the generation of the build string to alleviate need for multiple operations in a proxy environment.... Previously the `builder` var would be initialised once outside the proxy test and then reassigned within the proxy test with additional params.  By buffering the docker build flags throughout the flag checks we can then just initialise `builder` once post-checks with code as succinct as the original initialisation.
**EDIT** Added a function `buildFlagString` to build the build flags string, as suggested.

### **Test:**
**Pre:**
On Master ran `./faas-cli -action build -yaml ./samples.yml -no-cache`
and confirmed bug:
`docker build  --no-cache-t alexellis2/faas-urlping .`
`unknown flag: --no-cache-t`

**Post:**
Ran `./faas-cli -action build -yaml ./samples.yml`
and confirmed the command run was `docker build -t alexellis2/faas-urlping .`
Ran `./faas-cli -action build -yaml ./samples.yml -no-cache`
and confirmed the command run was `docker build --no-cache -t alexellis2/faas-urlping .`
Ran `./faas-cli -action build -yaml ./samples.yml -squash`
and confirmed the command run was `docker build --squash -t alexellis2/faas-urlping .`
Ran `./faas-cli -action build -yaml ./samples.yml -no-cache -squash`
and confirmed the command run was `docker build --no-cache --squash -t alexellis2/faas-urlping .`

Also tested `-flag=true` & `-flag=false` and confirmed expected outcomes.

### **Misc:**
Used git commit -s as per the contribution guide

### DIFF
--- a/app.go
+++ b/app.go
@@ -45,6 +45,7 @@ func main() {
 	var yamlFile string
 	var yamlFileShort string
 	var version bool
+	var squash bool
 
 	flag.StringVar(&handler, "handler", "", "handler for function, i.e. handler.js")
 	flag.StringVar(&image, "image", "", "Docker image name to build")
@@ -59,6 +60,7 @@ func main() {
 	flag.StringVar(&yamlFile, "yaml", "", "use a yaml file for a set of functions")
 	flag.StringVar(&yamlFileShort, "f", "", "use a yaml file for a set of functions (same as -yaml)")
 	flag.BoolVar(&version, "version", false, "show version and quit")
+	flag.BoolVar(&squash, "squash", false, "use Docker's squash flag for potentially smaller images (currently experimental)")
 
 	flag.Parse()
 
@@ -118,7 +120,7 @@ func main() {
 					function.Name = k
 					// fmt.Println(k, function)
 					fmt.Printf("Building: %s.\n", function.Name)
-					buildImage(function.Image, function.Handler, function.Name, function.Language, nocache)
+					buildImage(function.Image, function.Handler, function.Name, function.Language, nocache, squash)
 				}
 			}
 		} else {
@@ -134,7 +136,7 @@ func main() {
 				fmt.Println("Please provide the deployed -name of your function.")
 				return
 			}
-			buildImage(image, handler, functionName, language, nocache)
+			buildImage(image, handler, functionName, language, nocache, squash)
 		}
 		break
 	case "deploy":
@@ -244,7 +246,7 @@ func pushImage(image string) {
 	execBuild("./", []string{"docker", "push", image})
 }
 
-func buildImage(image string, handler string, functionName string, language string, nocache bool) {
+func buildImage(image string, handler string, functionName string, language string, nocache bool, squash bool) {
 
 	switch language {
 	case "node", "python":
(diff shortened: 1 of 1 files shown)


alexellis:
> Refactored the generation of the build string to alleviate need for multiple operations in a proxy environment.

Can you be specific about this?

app.go
```diff
@@ -244,24 +246,31 @@ func pushImage(image string) {
...
 
 	switch language {
 	case "node", "python":
 		tempPath := createBuildTemplate(functionName, handler, language)
 
 		fmt.Printf("Building: %s with Docker. Please wait..\n", image)
 
-		cacheFlag := ""
+		//lets create a buffer for any build flags
+		var flagBuf bytes.Buffer
```
alexellis: Could this be a string instead?

alexellis:
Squashing looks good - just a couple of questions/comments.

app.go
rgee0: It could.  What I've read though suggests that using byte buffer is better performance wise when there is any kind of variability in length - strings are immutable.  Less of a consideration with 3 functions and four concats per function, granted, but I'd hope users would be using the CLI for batching many more functions.

rgee0:
Updated the original main change section with further detail about the `builder` refactor

app.go
alexellis: I thought that would be the thinking. This code is building a very short string so I think I prefer the string concatenation and readability.

alexellis: We should probably extract a method to generate this string (whether it uses a buffer internally or a string) so it can be unit tested.

app.go
```diff
@@ -244,24 +246,31 @@ func pushImage(image string) {
...
+		}
+		if squash {
+			flagBuf.WriteString("--squash ")
 		}
 
-		builder := strings.Split(fmt.Sprintf("docker build %s-t %s .", cacheFlag, image), " ")
-		if len(os.Getenv("http_proxy")) > 0 || len(os.Getenv("http_proxy")) > 0 {
-			builder = strings.Split(fmt.Sprintf("docker build %s--build-arg http_proxy=%s --build-arg https_proxy=%s -t %s .", cacheFlag, os.Getenv("http_proxy"), os.Getenv("https_proxy"), image), " ")
+		if len(os.Getenv("http_proxy")) > 0 || len(os.Getenv("https_proxy")) > 0 {
+			flagBuf.WriteString("--build-arg http_proxy=" + os.Getenv("http_proxy") + " ")
```
alexellis: Should this be changed to sprintf since we are concatenating strings here and in other places use `fmt.Sprintf`?

rgee0: Well, ultimately, yes, now that we have to go back to strings.

app.go
```diff
@@ -244,24 +246,31 @@ func pushImage(image string) {
...
+			flagBuf.WriteString("--build-arg https_proxy=" + os.Getenv("https_proxy") + " ")
 		}
 
+		builder := strings.Split(fmt.Sprintf("docker build %s-t %s .", flagBuf.String(), image), " ")
 		fmt.Println(strings.Join(builder, " "))
```
rgee0: Whats the idea here?  Seems to split on spaces then join on spaces?

rgee0:
Made changes as discussed, committed and squashed.

alexellis:
Looks very clean and concise. Merged.

PR #26, opened 2017-08-09
Add Ruby template

alexellis:
Basic Ruby template

Now has gem support via Gemfile.


alexellis:
What do you think @errm? Will this work for the kind of stuff you are doing at Reevoo?

alexellis:
Merging initial support for Ruby language.

PR #46, opened 2017-08-22
Add Custom Builds : fixes #29

### LINKED ISSUE #29: Should have - "Custom" template type
## Expected Behaviour

It should be possible to "build" a function which isn't using one of the known / supported template types. An example would be the shrink_image example in the repo.

Snippet:

```yaml
  shrink-image:
    lang: custom
    handler: ./samples/image-magick
    image: functions/resizer
    fprocess: "convert - -resize 50% fd:1"
```

The context is whatever is in the `handler` folder and the Dockerfile in the root of the directory will be used for the build.

## Current Behaviour

Build manually and remove the `lang` element then mark the item as `skip_build`

## Possible Solution

When setting the language to `custom` then a `docker build` is invoked in the handler folder directly.

## Context

This will add versatility - especially for languages which don't suit dynamic templating like Golang.

dudymas:
## Description

The builder handles a new language type, specified by the string 'custom'
This type treats the 'handler' as a directory with a dockerfile and builds it.

## Motivation and Context

This is for issue #29 

## How Has This Been Tested?

This code has been manually tested using rebuilt faas-cli binaries on a windows machine.

## Types of changes

## Checklist:

(note: I spaced out some string concats in the builder. Holler if I shouldn't do that)

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,24 @@ $ uname -a | curl http://localhost:8080/function/nodejs-echo--data-binary @-
 
 > For further instructions on the manual CLI flags (without using a YAML file) read [manual_cli.md](https://github.com/alexellis/faas-cli/blob/master/MANUAL_CLI.md)
 
+#### Adding custom function support
+
+If you want to add a function that does not have a template (such as `go`),
+you only need to place a dockerfile within a directory such that:
+
+1. The dockerfile builds successfully with appropriate settings:
+   * The `fprocess` env var should be the path + flags for an executable binary
+   * The binary should accept stdin and print to stdout
+2. You specify the directory that contains the dockerfile as a 'handler':
+   * For the yaml, just put in the path. You can see this in `samples.yml`
+  for the golang-echo function
+   * On the command line, use the `-handler` flag
+3. You specify the the `language` as `custom`:
+   * For the yaml, specify `lang: custom` under the function config
+   * On the command line, use the flag `-lang custom`
+
+You can look at [the golang-echo sample for reference](./sample/golang-echo)
+
 ## FaaS-CLI Developers / Contributors
 
 ### Updating the brew formula
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ faas-cli-armhf
 master.zip
 *.png
 
+obj
--- a/builder/build.go
+++ b/builder/build.go
@@ -9,28 +9,54 @@ import (
 )
 
 // BuildImage construct Docker image from function parameters
-func BuildImage(image string, handler string, functionName string, language string, nocache bool, squash bool) {
+func BuildImage(image string, handler string, functionName string, languageTemplate string, nocache bool, squash bool) {
+	var tempPath string
+	var builder []string
+	flagStr := buildFlagString(nocache, squash, os.Getenv("http_proxy"), os.Getenv("https_proxy"))
 
-	switch language {
+	switch languageTemplate {
 	case "node", "python", "ruby", "csharp":
-		tempPath := createBuildTemplate(functionName, handler, language)
-
-		fmt.Printf("Building: %s with Docker. Please wait..\n", image)
-
-		flagStr := buildFlagString(nocache, squash, os.Getenv("http_proxy"), os.Getenv("https_proxy"))
-
-		builder := strings.Split(fmt.Sprintf("docker build %s-t %s .", flagStr, image), " ")
-		fmt.Println(strings.Join(builder, " "))
-		ExecCommand(tempPath, builder)
+		tempPath = createBuildTemplate(functionName, handler, languageTemplate)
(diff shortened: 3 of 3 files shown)


dudymas:
I don't mind squashing this. I've left commits separated to expedite review and rework

README.md
```diff
@@ -98,6 +98,24 @@ $ uname -a | curl http://localhost:8080/function/nodejs-echo--data-binary @-
 
 > For further instructions on the manual CLI flags (without using a YAML file) read [manual_cli.md](https://github.com/alexellis/faas-cli/blob/master/MANUAL_CLI.md)
 
+#### Adding custom language support
+
+If you want to add a function that does not have a template (such as `go`),
+you only need to place a dockerfile with in a directory such that:
+
+1. The dockerfile builds successfully with appropriate settings:
```
alexellis: dockerfile -> Dockerfile

README.md
```diff
@@ -98,6 +98,24 @@ $ uname -a | curl http://localhost:8080/function/nodejs-echo--data-binary @-
 
 > For further instructions on the manual CLI flags (without using a YAML file) read [manual_cli.md](https://github.com/alexellis/faas-cli/blob/master/MANUAL_CLI.md)
 
+#### Adding custom language support
```
alexellis: Not just custom language, this could be a pre-existing binary. For instance take the samples here -> https://github.com/alexellis/faas/tree/master/sample-functions

and here -> https://github.com/alexellis/faas/blob/master/docker-compose.yml

sample/golang-echo/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.9-alpine
```
alexellis: What's the current LTS version?

dudymas: 3.5 ... setting that nowish

dudymas: of course, all the older stable container tags are unsupported AND vulnerable. For now, I'm setting this to `golang:1.9-alpine3.6`
would that be okay. If not, got a preferred golang container for this sample? Or I could just scrap it for fsharp instead

alexellis: Docker have official Golang images, are you certain that the LTS Golang version is vulnerable? Can you link to something showing that?

README.md
dudymas: I updated the wording of this to:
> Adding custom function support

will that work?

sample/golang-echo/Dockerfile
dudymas: https://hub.docker.com/r/library/golang/tags/ only lists the alpine3.6 versions as good. And I misunderstood what you meant by LTS. I thought you were talking about which version of alpine has best support.

alexellis: I meant LTS of Golang. By the way - where are you seeing the bad flags on the images? How about 1.8-alpine3.6?

alexellis: Here's a different idea - can we do the sample in a different PR? .. just the custom support in this one and then we can get it merged.

dudymas: that's a solid idea... picking this fellow out now

alexellis:
Hi @dudymas,

Please can you correct the case of dockerfile->Dockerfile?

Can you show the output from the build action with the a Golang sample like this one?

https://github.com/alexellis/faas/tree/master/sample-functions/DockerHubStats

At the same time would you like to update that sample to use a multi-stage build? (it will need ca-certs / ca-certificates as a dependency for the web scraping)

README.md
```diff
@@ -98,6 +98,24 @@ $ uname -a | curl http://localhost:8080/function/nodejs-echo--data-binary @-
...
+you only need to place a dockerfile within a directory such that:
+
+1. The dockerfile builds successfully with appropriate settings:
+   * The `fprocess` env var should be the path + flags for an executable binary
+   * The binary should accept stdin and print to stdout
+2. You specify the directory that contains the dockerfile as a 'handler':
+   * For the yaml, just put in the path. You can see this in `samples.yml`
+  for the golang-echo function
+   * On the command line, use the `-handler` flag
+3. You specify the the `language` as `custom`:
```
johnmccabe: PR is looking good @dudymas, the only concern I have is that `custom` is a bit too broad for what is in effect a Dockerfile specific template.

Changing the language from `custom` to `dockerfile` would make its behaviour unambiguous.

alexellis:
@dudymas thanks for working on this. I've added some changes in PR https://github.com/alexellis/faas-cli/pull/52 which mean we can get this into the code faster. I made a couple of changes: Dockerfile is the language instead of custom and there is no temporary build folder in use (since there is no template).

dudymas:
No worries. I had a busy week, and I know how fast this PR was put together ;-)

Issue #59, opened 2017-09-02
Issue - function sub-folders are not copied recursively


alexellis:
## Expected Behaviour

All / any assets such as a "src" or "samples" folder should be copied into the final image.

## Current Behaviour

Only files in the root directory are copied in

## Possible Solution

Seems as simple as setting a flag to "true" in the builder.

alexellis:
@rgee0 https://github.com/alexellis/faas-cli/compare/recurse_function_copy?expand=1

PR #61, opened 2017-09-02
Added recusion to handler copy + debugPrint fn

rgee0:

Addresses issue #59.  

## Description
Changes the recursive arg to true on the handler `copyFiles` call so that sub directories within the handler are also copied into the build directory. 
Extracted a `debugPrint` function to avoid code repetition as The new directory branch in `copyFiles` now messages out only if the system is creating a new directory and `debug` is enabled.
Corrected a typo which meant that `err` was being tested rather than `newDirErr` when a new directory is being created.
Changed FaaS to OpenFaaS where seen.

## Motivation and Context
#59
- [X] There is a raised an issue to propose this change ([required](https://github.com/alexellis/faas/blob/master/CONTRIBUTING.md))

## How Has This Been Tested?
Created 10 levels of files and folders beneath a handler folder
Ran ./faas-cli -action build -yaml ./samples.yml
Confirmed that the images built successfully.
Confirmed that the expected files and folders were extant in the build folder following completion of build

## Types of changes

## Checklist:

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -50,7 +50,7 @@ func BuildImage(image string, handler string, functionName string, language stri
 	}
 }
 
-// createBuildTemplate creates temporary build folder to perform a Docker build with Node template
+// createBuildTemplate creates temporary build folder to perform a Docker build with language template
 func createBuildTemplate(functionName string, handler string, language string) string {
 	tempPath := fmt.Sprintf("./build/%s/", functionName)
 	fmt.Printf("Clearing temporary build folder: %s\n", tempPath)
@@ -72,7 +72,7 @@ func createBuildTemplate(functionName string, handler string, language string) s
 	copyFiles("./template/"+language, tempPath, true)
 
 	// Overlay in user-function
-	copyFiles(handler, tempPath+"function/", false)
+	copyFiles(handler, tempPath+"function/", true)
 
 	return tempPath
 }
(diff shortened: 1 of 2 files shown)


alexellis:
Thanks for the fixes. If all usages are recursive then the flag is redundant and can be removed.

alexellis:
Can we get any test coverage for this scenario?

builder/build.go
```diff
@@ -161,3 +160,10 @@ func buildFlagString(nocache bool, squash bool, httpProxy string, httpsProxy str
 
 	return buildFlags
 }
+
+func debugPrint(message string) {
+
+	if val, exists := os.LookupEnv("debug"); exists && (val == "1" || val == "true") {
```
johnmccabe: Could add a `--debug` flag for turning this on across all commands, something for a seperate PR perhaps.

johnmccabe:
have built and run a few template builds locally and it looks good 👍

builder/build.go
rgee0: Yes, that was kind of the theory behind pulling it out.  Suggest a separate PR since this one is a 'bug' fix

alexellis:
Thanks for helping to integration test. I would like some automated tests in "Golang" for this.. doesn't have to be super extensive. Can either of you take that on?

PR #64, opened 2017-09-03
Remove master.zip file after building function with faas-cli build

ems5311:
This change will remove the master.zip file after the `fetchTemplates()` function is done creating the templates from the downloaded zip file.

## Description

Use the `os.Remove()` function to remove the zip file, whose name is now stored in a const variable, after making sure that the zip file is still present with `os.Stat()`

## Motivation and Context

- [ ] I have raised an issue to propose this change ([required](https://github.com/alexellis/faas/blob/master/CONTRIBUTING.md))
Noticed the zip file hanging around, was confused as to its necessity to remain the directory. However, as long as the `templates` directory is present in your YAML build file location, the `master.zip` file will not be used. Therefore, it is not necessary to keep it around.

## How Has This Been Tested?

Tried building the node, python, csharp, ruby, and Dockerfile templates, and checked that the master.zip file was removed, but the function worked when deployed.

## Types of changes

## Checklist:

### DIFF
--- a/commands/fetch_templates.go
+++ b/commands/fetch_templates.go
@@ -18,9 +18,10 @@ import (
 // fetchTemplates fetch code templates from GitHub master zip file.
 func fetchTemplates() error {
 
+	const zipFileName = "./master.zip"
 	err := fetchMasterZip()
 
-	zipFile, err := zip.OpenReader("./master.zip")
+	zipFile, err := zip.OpenReader(zipFileName)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


alexellis:
@ems5311 thanks for looking at this. Can you suggest an automated test for checking this behaviour?

If we can easily mock the filesystem or http call then we can save on bandwidth/time during test runs.

commands/fetch_templates.go
```diff
@@ -51,6 +52,11 @@ func fetchTemplates() error {
 		}
 	}
 
+	log.Printf("Cleaning up zip file...")
+	if _, err := os.Stat(zipFileName); err == nil {
```
alexellis: If there is an error we should probably print that.

ems5311:
forgot to rebase before pushing, need to create new PR from new branch

Issue #60, opened 2017-09-02
Proposal - filter functions in stacks


alexellis:
## Expected Behaviour

If a stack has 10 functions but I only want to build one (by name) - then I should be able to do this:

```
$ faas-cli build -f stack.yml --filter "payroll"
```

Applies to:

* delete, build, deploy, push

alexellis:
@johnmccabe - feel free to add examples / alternative.

alexellis:
Thanks to @ems5311 we now have this and it will be out in the next release.

PR #78, opened 2017-09-05
Added basic filter functionality with regex search in stack.ParseYAML

ems5311:
Added basic filter functionality with regex search in stack.ParseYAML

## Description

This change adds filter functionality by introducing a new Persistent Flag which is called `--filter`. You can provide a regexp for this filter, and it will select functions from your YAML file (specified with `-f`) that are a regex match.

This works by removing everything from the parse YAML map `Functions` which is a member of the `stack.Services` struct.

## Motivation and Context

- [ ] I have raised an issue to propose this change ([required](https://github.com/alexellis/faas/blob/master/CONTRIBUTING.md))
#60 

## How Has This Been Tested?

Manually tested `faas-cli build` and `faas-cli deploy` with the `--filter` flag for various options.
EDIT: Tests now added

## Types of changes

## Checklist:


build.sh
```diff
@@ -5,3 +5,4 @@ docker build --build-arg http_proxy=$http_proxy --build-arg https_proxy=$https_p
  docker cp faas-cli:/root/faas-cli . && \
  docker rm -f faas-cli
 
+go test ./...
```
alexellis: What does this do? `./...`

stack/stack_test.go
```diff
@@ -0,0 +1,64 @@
...
+		title:        "Filter for all functions: '.*'",
+		filter:       ".*",
+		functions:    []string{"nodejs-echo", "ruby-echo", "shrink-image", "url-ping"},
+	},
+}
+
+func Test_parseYaml(t *testing.T) {
+	for _, test := range ParseYamlTests {
+		t.Run(test.title, func(t *testing.T) {
+			parsedYaml, err := ParseYAML("../samples.yml", test.filter)
```
alexellis: Brilliant to see the unit tests 💯 So that this test doesn't become brittle - let's create a test .yml file and keep it alongside the Golang code. Even better if you add an override which takes []bytes or string rather than a filename - then we can hard-code the test case into the file and use multiple inputs.

build.sh
ems5311: Runs all unit tests in the entire project, directory by directory. If it encounters a directory without tests, it just does nothing. It the dir has tests, it runs them.

stack/stack_test.go
```diff
@@ -0,0 +1,92 @@
...
+	{
+		title:        "Filter functions only containing '.+-.+'",
+		filter:       ".+-.+",
+		functions:    []string{"nodejs-echo", "ruby-echo", "url-ping"},
+	},
+	{
+		title:        "Filter for all functions: '.*'",
+		filter:       ".*",
+		functions:    []string{"imagemagick", "nodejs-echo", "ruby-echo", "url-ping"},
+	},
```
johnmccabe: You should add some negative tests, for example what happens when the filter results in an empty function slice.

stack/stack.go
```diff
@@ -43,11 +50,21 @@ func ParseYAML(yamlFile string) (*Services, error) {
...
+		for k, function := range services.Functions {
+			function.Name = k
+			match, _ := regexp.MatchString(filter, function.Name)
+			if !match {
+				delete(services.Functions, function.Name)
+			}
+		}
+	}
+
 	return &services, err
```
johnmccabe: Shouldn't the error be `nil` here?

stack/stack.go
```diff
@@ -32,6 +32,13 @@ func ParseYAML(yamlFile string) (*Services, error) {
...
 	}
+	return ParseYamlData(fileData, filter)
+}
+
+// ParseYamlData parse YAML data into a stack of "services".
+func ParseYamlData(fileData []byte, filter string) (*Services, error) {
+	var services Services
+	var err error
 
 	err = yaml.Unmarshal(fileData, &services)
```
johnmccabe: can just use a short assignment here `:=` and dump the `err` forward declaration on line 41.

stack/stack_test.go
johnmccabe: Also what happens when the regex is invalid.

stack/stack.go
```diff
@@ -43,11 +50,21 @@ func ParseYAML(yamlFile string) (*Services, error) {
 		return nil, fmt.Errorf("'%s' is the only valid provider for this tool - found: %s", providerName, services.Provider.Name)
 	}
 
+	if len(filter) > 0 {
+		for k, function := range services.Functions {
+			function.Name = k
+			match, _ := regexp.MatchString(filter, function.Name)
```
johnmccabe: should handle the error here and exit with a message warning the user why the command has failed

stack/stack.go
ems5311: Do you mean I should set the `err` to `nil` explicitly before returning it?

johnmccabe: for clarity yes

stack/stack_test.go
ems5311: When the function results in no functions being matched, the same behavior will be exhibited as if you had called `faas-cli <verb> -f <YAML>` on a YAML file which does not have a `functions:` object. What it does with the `build` action is prints the following error message.

`Please provide a valid -image name for your Docker image.`

I don't think it's within the scope of this PR to fix the behavior of commands when they parse a YAML file and there are no functions.

If the regex is invalid, the err variable will be set (I will change it to storing and checking the `err` variable after the call to `regexp.MatchString()`)

ems5311: Actually, I was wrong about the error behavior. I found a case that causes the `err` var to be set to non-`nil`:

```
faas-cli deploy -f samples.yml --filter "*"
2017/09/06 06:32:49 Error parsing regex passed in through --filter argument: error parsing regexp: missing argument to repetition operator: `*`
2017/09/06 06:32:49 error parsing regexp: missing argument to repetition operator: `*`
```

If you issue a command with quotes around the `*` variable without an argument, the go `regexp` parser errors out. I think this is another reason why filter should be wildcard matching rather than regexp matching. I think we can use [this class for wildcard matching](https://github.com/ryanuber/go-glob/blob/master/glob.go).

ems5311:
@johnmccabe Addressed your comments as best I could in the latest commit. Added a few test cases.

stack/stack.go
```diff
@@ -43,11 +49,26 @@ func ParseYAML(yamlFile string) (*Services, error) {
...
+				fmt.Printf("Error parsing regex passed in through --filter argument: %v", err)
+				return nil, err
+			}
+			if !match {
+				delete(services.Functions, function.Name)
+			}
+		}
+	}
+
+	err = nil
```
johnmccabe: Better to just return nil, don't bother assigning a value to err
```
return &services, nil
```

stack/stack.go
```diff
@@ -43,11 +49,26 @@ func ParseYAML(yamlFile string) (*Services, error) {
 		return nil, fmt.Errorf("'%s' is the only valid provider for this tool - found: %s", providerName, services.Provider.Name)
 	}
 
+	if len(filter) > 0 {
+		for k, function := range services.Functions {
+			function.Name = k
+			match, err := regexp.MatchString(filter, function.Name)
+			if err != nil {
+				fmt.Printf("Error parsing regex passed in through --filter argument: %v", err)
```
johnmccabe: Minor tweak I'd change the error to `Invalid regex passed to filter`

stack/stack.go
```diff
@@ -43,11 +49,26 @@ func ParseYAML(yamlFile string) (*Services, error) {
...
+	}
+
+	err = nil
 	return &services, err
 }
 
-// fetchYAML pulls in file from remote location such as GitHub raw file-view
-func fetchYAML(address *url.URL) ([]byte, error) {
+// fetchYaml pulls in file from remote location such as GitHub raw file-view
+func fetchYaml(address *url.URL) ([]byte, error) {
```
johnmccabe: should be fetchYAML (in comments and fn calls etc)

stack/stack_test.go
johnmccabe: @ems5311  we've had a chat about this, and think your glob suggestion is a good one, we'd still like to keep the regex handling however so how about two flags:
- `--filter=` using your glob suggestion for a simple filtering experience
- `--regex=` using the current regex handling that you have here for a richer filtering

Thanks for all your work on this btw, its looking great so far !!!

stack/stack.go
ems5311: blah, missed this one. Good catch

johnmccabe:
Can you also add an example to each of the command declarations, for example in `commands/deploy.go`, in the `deployCmd` var:
```
	Example: `  faas-cli deploy -f https://domain/path/myfunctions.yml
  faas-cli deploy -f ./samples.yml
  faas-cli deploy -f ./samples.yml --replace=false
  faas-cli deploy -f ./samples.yml --filter="*gif*"
  faas-cli deploy -f ./samples.yml --regex="fn[0-9]_.*"
```

alexellis:
Have you covered all of John's feedback?

ems5311:
Now I have addressed all of John's feedback, also added checks for if `--filter` and `--regex` is passed in, and added a test for that. Added some negative tests for `--regex`. `--filter` does not return any `error`, so I think the current test coverage is OK.

I have squashed these changes into one commit.

alexellis:
That sounds perfect, thank you so much! 

I'll push a dev-build, if you can get 1-2 people to test it on the Slack channel I'll merge.

alexellis:
https://github.com/alexellis/faas-cli/releases/tag/0.4.13-alpha

alexellis:
Thank you @ems5311 - I'll merge and we'll need to test this extensively in the next pre-release build.

Issue #85, opened 2017-09-06
Proposal: third-party templates via GitHub


alexellis:
## Expected Behaviour

Anyone should be able to download a community-supported 3rd party template for Perl, PHP, Kotlin etc.

## Current Behaviour

PRs to main project.

## Possible Solution

I'm open to input on this too.

```
$ faas-cli add-template https://github.com/alexellis/kotlin-template

Downloading https://github.com/alexellis/kotlin-template/master/blob/master.zip
Expanding kotlin/Dockerfile to template/kotlin/Dockerfile
...
```

Repo structure:

./template/kotlin/Dockerfile
./template/kotlin/handler.java
./README.md

Initial version of this can be a "dumb" shortcut to download a zip file and unzip into a directory.

Out of scope for v1:

* Listing templates installed locally
* Online directory of templates

Mentions: @johnmccabe and others for the suggestion. @nicholasjackson how does Hashicorp do this?

Other files which need changes:
https://github.com/alexellis/faas-cli/pull/82#issuecomment-327607545

`fprocess` needs to come out of the Golang code and sit in a `template.yml` file within i.e. `./template/kotlin/template.yml`

```
language: kotlin
fprocess: kotlin handler.kotlin
```

johnmccabe:
Derek add label: proposal

rgee0:
Questions:

ems5311:
I'm happy to take on this issue if that's okay. Should have a first-run of it committed by tonight (west coast time).

ems5311:
Derek add label: skill/intermediate

alexellis:
@rgee0 - maybe we should "cowardly refuse" to overwrite files?

@ems5311 absolutely, please do. I'd see this as re-using some of the work around "fetchTemplates".

alexellis:
Derek add label: skill/intermediate

alexellis:
@ems5311 added some more detail in the original comment.

ems5311:
@alexellis, Should this change expect there to be a separate `template.yml` file in each template folder? And would the schema of that file be just `language:` and `fprocess:`?

alexellis:
In terms of decoupling the Golang code I think yes. What's your take on it?

aafrey:
[...]

PR #87, opened 2017-09-08
Create `faas-cli add-template --url <URL>` command

ericstoekl:
1. Create `faas-cli add-template --url <URL>` command
2. Add template.yml files for each language in the ./template dir.
3. Read fProcess from ./template/<lang> in deploy command, instead of using hard-coded value
4. Adds mock-server test

## Description

Adds the `addTemplate` command and a basic test for it. Changes the way that `fProcess` command is acquired: from a hard-coded value in `proxy/deploy.go` to a new YAML file in each language-specific dir in the `template/` dir, for example `template/csharp/template.yml`.

## Motivation and Context

- [X] I have raised an issue to propose this change ([required](https://github.com/alexellis/faas/blob/master/CONTRIBUTING.md))
#85 

## How Has This Been Tested?

Tests added in `addTemplate_test.go`

## Types of changes

## Checklist:


alexellis:
Looking forward to looking at this. One quick note. It would be
"add-template" for the command and add_template for filenames.

ems5311:
Updated file names, improved help for `add-templates` command

alexellis:
For reference let's update the description of this PR with the naming "add-template".

commands/add_template.go
```diff
@@ -0,0 +1,52 @@
...
+directory from the root of the repo, if it exists.`,
+	Example: "faas-cli add-template --url https://github.com/alexellis/faas-cli",
+	Run: runAddTemplate,
+}
+
+func runAddTemplate(cmd *cobra.Command, args []string) {
+	URL = strings.TrimRight(URL, "/")
+	URL = URL + "/archive/master.zip"
+
+	err := os.Setenv("templateUrl", URL)
```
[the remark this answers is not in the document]
ems5311: Sure. I can change `fetchTemplates()` in `fetch_temlates.go` to accept a param of the custom URL.

ems5311:
Question about this PR: should I add some logic to scan through the repository that is passed in to search for the template dir? Or should we look for third party templates in the root path of a repo? Currently `fetchTemplates()` only looks for the template dir in the `faas-cli-master/` directory, because as the code is written right now, it only expects to be grabbing a template dir from the alexellis/faas-cli repo. The implication of this is that the repository containing 3rd party templates needs to be named `faas-cli`. We could have a 'custom URL' flag that `fetchTemplates()` takes to tell that the repo may be named anything. Or maybe if a custom URL is passed into `fetchTemplates()`, it just uses whatever repo name the URL points to.

https://github.com/alexellis/faas-cli/blob/3a2e81471c69275571d71f84c158bb787fb1ad7b/commands/fetch_templates.go#L31

ems5311:
Thanks to @itscaro for huge help with this update.

alexellis:
@ericstoekl what's remaining to get this in?

itscaro:
@alexellis It's ready to be merge ;-)

ericstoekl:
@alexellis , I agree with Minh-Quan, it looks ready for merging.

alexellis:
>This branch has conflicts that must be resolved

itscaro:
@alexellis Conflicts are resolved 👍

commands/add_template.go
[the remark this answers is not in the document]
itscaro: The updated version no longer uses env var.

ericstoekl:
Resolved merge conflicts

ericstoekl:
@itscaro , revisiting this PR, what is the point of the `.cache` directory again? In the current state of the code, after `fetchTemplates()` is run, the zip file contained in `.cache/` is deleted, so an empty `.cache/` directory remains. Is there a point to having this directory? Would it make more sense to just not create the `.cache/` directory, because we're going to just remove the zip file at the end anyway?

ericstoekl:
Looks good :+1:

alexellis:
Derek add label: status/testing

alexellis:
Can you demo this on contributors' call?

commands/fetch_templates.go
```diff
@@ -15,111 +15,154 @@ import (
...
+				if !canWriteLanguage(language, overwrite) {
+					continue
+				}
+			}
+		} else {
+			// template/
+			continue
+		}
+
 		if strings.Index(relativePath, "template") == 0 {
```
ericstoekl: Is this `if` statement needed? If `"template"` is not in the `relativePath`, we would have `continue`d  on lines 43-46 above:

```
	relativePath := z.Name[strings.Index(z.Name, "/")+1:]
+		if strings.Index(relativePath, "template/") != 0 {
+			// Process only directories inside "template" at root
+			continue
+		}
```

stack/language_template.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+// ParseYAML parse a YAML file into a LanguageTemplate struct.
+func ParseYAMLForLanguageTemplate(yamlFile string) (*LanguageTemplate, error) {
```
ericstoekl: We're going to need a test for this

guide/TEMPLATE.md
```diff
@@ -0,0 +1,53 @@
...
+    └── template.yml
+```
+
+## Download external repository
+
+In order to build functions using 3rd party templates, you need to add 3rd templates before the build step, with the following command:
+
+```
+./faas-cli add-template https://github.com/alexellis/faas-cli
+```
```
rgee0: To avoid potential confusion for new users, should this be a genuine third party template repo, even if its just a copy of our current templates in an unfamiliar repo?

guide/TEMPLATE.md
```diff
@@ -0,0 +1,53 @@
...
+
+If you need to update the downloaded repository, just add the flag `--overwrite` to the download command:
+
+```
+./faas-cli add-template https://github.com/alexellis/faas-cli --override
+```
```
rgee0: Same idea here

alexellis: I guess it would be helpful if it was a different repo than the one we use for the project templates

alexellis: Incidentally this is a good way to update templates ^

rgee0: Yes - probably worth spelling that out in the guide too.

ericstoekl: I appreciate the input. Would this repo work?

https://github.com/ericstoekl/faas-custom-templates

Or would a copy of faas-cli work better? (https://github.com/ericstoekl/faas-cli)

itscaro: I'm for either https://github.com/ericstoekl/faas-custom-templates or https://github.com/itscaro/openfass-template-php, to show the way to add a community-supported languages.

rgee0: I think the former.  Really what we are trying to demonstrate is separation of repos so that its obvious to the user that they can point it at a repo that doesn't follow the structure of faas-cli. 

(also, the latter wouldn't work that well for ppl who have forked)

rgee0:
@ericstoekl I think that so far the approach has been from the perspective of a user starting with a set of official templates and overriding with some their own.  What happens if a user has started a new function repo locally and pulled down custom templates of a subset of the languages available?

The user has their own go and ruby templates for example.  No other templates are available locally.

The user then runs `faas-cli new --list` and sees:
```
$ faas-cli new --list
Languages available as templates:
- node
- python
- python3
- ruby
- csharp
- Dockerfile
- go
```

Seeing this list the user runs:

`$ faas-cli new thisisthename --lang node`

This creates `./thisisthename` which is empty.

There are a couple of points here, I think:

1) `faas-cli new --list` could become more intelligent and display only the language templates currently available.

2) If the user tries to create a new function using a language not currently available locally there needs to be a clear route of action.  That could be politely decline because the template isn’t available, or something like download the official templates (to somewhere other than `./template`) and move only the required template (or all the official templates without clobbering the user’s) into the working template dir, or something else.

rgee0:
I can see that `add-template` was the chosen flag from the start but I wonder whether something more intuitive & phonetically robust might be better.  Something that developers naturally associate with like `fetch` / `pull` or similar.  This would also fit with the short/sharp verbs currently in use

itscaro:
There is "faas-cli push" for "docker push", imo, I would avoid using fetch/pull other non-docker operations.

Maybe adding a `template` group of commands?
`faas-cli template add`
or
`faas-cli template pull`

still I prefer `add`, to stay in the line with `git remote add`

itscaro:
@rgee0 

`
mkdir new-env && cd new-env

..\faas-cli new --list
Languages available as templates:
- Dockerfile


Or alternatively create a folder containing a Dockerfile, then pick
the "Dockerfile" lang type in your YAML file.

..\faas-cli new test --lang xcom
2017/10/08 13:26:51 No templates found in current directory.
2017/10/08 13:26:51 HTTP GET https://github.com/openfaas/faas-cli/archive/master.zip
2017/10/08 13:26:52 Writing 231Kb to template-openfaas-faas-cli.zip

2017/10/08 13:26:52 Attempting to expand templates from template-openfaas-faas-cli.zip
2017/10/08 13:26:53 Fetched 9 template(s) from https://github.com/openfaas/faas-cli
2017/10/08 13:26:53 Cleaning up zip file...
xcom is unavailable or not supported.

`

`
rm -rf template && mkdir template

..\faas-cli new test --lang xcom
xcom is unavailable or not supported.
`

johnmccabe:
I think we need to be very careful about making ad-hoc evolutions to the cli here, as it will increasingly loose coherence - the **most** important thing is that it should "just make sense", in that a user should be able to pretty much guess right first time what command invocations are required to do what they want.

Do you feel thats what we have here - looking at the cli verbs as a whole does it make sense/flow?

johnmccabe:
The initial port to the verb based system was carried out with the explicit intention to not deviate from the structure of the old CLI. As we add more functionality I think its reached the point where we need to reevaluate the desired UX for the medium/long term.

alexellis:
What suggestions do you have Richard/John?

rgee0:
I think @johnmccabe covered it very well. If we compare our list of verbs to other prominent projects/products, their top level verb contexts tend to be mixed and so it might be that our current approach is already suited to the longer term and we simply need to ensure we maintain a standard.

Arguably we deviated from a de-facto standard when `new` was introduced - probably should have been `create` or `generate` or `concoct` or `invent`.  This should probably be using `fetch`, or `pull`, or `acquire`, or `import`.

As a rule I've always shied away from anything I would need to convey verbally as 'something hyphen something'.

ericstoekl:
@itscaro Did you make some changes and not commit them yet? The `new --list` and creation of non-existant language still is unfixed for me in `2d74b67d` (most recent commit to `addTemplateCmd`):

```
$ ls template/
bash  perl
$ fc new --list
Languages available as templates:
- node
- python
- python3
- ruby
- csharp
- Dockerfile
- go

Or alternatively create a folder containing a Dockerfile, then pick
the "Dockerfile" lang type in your YAML file.
$ fc new csharp-test2 --lang csharp
csharp is unavailable or not supported.
Folder: csharp-test2 created.
2017/10/08 16:16:35 open ./template/csharp/function/: no such file or directory
$ ls csharp-test2/
$ # empty dir
$ fc version
Commit: 2d74b67d217a7795831398f1c3a2933630ccead5
Version: dev

```

ericstoekl:
With regards to the naming issue, I like `faas-cli new template`, would make it a subsidiary of `faas-cli new`. @johnmccabe is this possible/easy to do with Cobra CLI?

itscaro:
@ericstoekl Since the `new` command now takes positional arg as function name, I think we should not use `new template`

ericstoekl:
@itscaro Thanks for the update. That's a good point about `new template`, prob not a good idea because we use `new <positionalArg>` already so it would get confusing.

How about `faas-cli import` then? The `--help` would state that it's for importing *languages*. I don't think there would be too much confusion in that case between `new`, `pull`, `import`...

ericstoekl:
Just pushed a commit to improve the `new_function_tests` file. Not sure why it's not showing up in the comments thread/being built by travis CI, but you can see it in the Commits tab.

alexellis:
I want to revisit this and see what we need to do to merge the work.

I like the suggestion via @itscaro about a sub-command

```
faas-cli template pull https://....
.....
# Writes a line to ~/.openfaas/config.json
```

ericstoekl:
Latest functionality for `faas-cli template pull`:

https://asciinema.org/a/ieyznKidFlvimpeiS5BFMG3ep

1. Now uses `faas-cli template pull`
2. If you use an incorrect verb for `faas-cli template <VERB>`, it will error and print a list of supported verbs (currently only `pull`).

commands/fetch_templates.go
```diff
@@ -18,113 +18,160 @@ import (
...
-	if _, err = os.Stat(ZipFileName); err != nil {
 
-		if len(templateUrl) == 0 {
-			templateUrl = "https://github.com/openfaas/faas-cli/archive/" + ZipFileName
-		}
+	templateURL = strings.TrimRight(templateURL, "/")
+	templateURL = templateURL + "/archive/master.zip"
 
+	templateURLSplit := strings.Split(templateURL, "/")
+	templateURLSplit = templateURLSplit[len(templateURLSplit)-4 : len(templateURLSplit)-2]
```
alexellis: I don't like the magic numbers -4 -2

What's this doing? maybe break out a function and document the indexes

commands/fetch_templates.go
```diff
@@ -137,3 +184,34 @@ func createPath(relativePath string, perms os.FileMode) error {
 	err := os.MkdirAll(dir, perms)
 	return err
 }
+
+// canWriteLanguage tells whether the language can be processed or not
+// if overwrite is activated, the directory template/language/ is removed before to keep it in sync
```
alexellis: @rgee0 @johnmccabe  I don't know if overwrite should delete anything. It should probably dump over the top. Not sure?

commands/fetch_templates.go
```diff
@@ -137,3 +184,34 @@ func createPath(relativePath string, perms os.FileMode) error {
...
+func canWriteLanguage(language string, overwrite bool) bool {
+
+	if len(language) > 0 {
```
alexellis: I'm struggling to read/decipher this function - think it can be made simpler?

commands/invoke.go
```diff
@@ -51,7 +51,7 @@ func runInvoke(cmd *cobra.Command, args []string) {
 	functionName = args[0]
 
 	if len(yamlFile) > 0 {
-		parsedServices, err := stack.ParseYAMLFile(yamlFile, regex, filter)
+		parsedServices, err := stack.ParseYAMLFileForStack(yamlFile, regex, filter)
```
alexellis: Why ParseYAMLFileForStack vs ParseYAMLFile?

commands/new_function.go
```diff
@@ -45,14 +45,20 @@ language or type in --list for a list of languages available.`,
 
 func runNewFunction(cmd *cobra.Command, args []string) {
 	if list == true {
+		var available_templates string
+
+		if f, err := ioutil.ReadDir(templateDirectory); err != nil {
```
alexellis: can we change `f` to something else please?

commands/new_function.go
```diff
@@ -45,14 +45,20 @@ language or type in --list for a list of languages available.`,
...
+			available_templates = "There is no available languages"
```
alexellis: "No language templates were found. Please run 'faas-cli template pull'."

commands/new_function.go
```diff
@@ -45,14 +45,20 @@ language or type in --list for a list of languages available.`,
...
+		} else {
+			for _, file := range f {
+				if file.IsDir() {
+					available_templates += fmt.Sprintf("- %s\n", file.Name())
```
alexellis: Build up an array in a separate function then return this - and have the formatting/printing separate.

commands/fetch_templates.go
rgee0: Delete is a strong action. I do think it important that the user knows exactly what the current state of their templates is and that they aren't able to 'accidentally' overwrite.

commands/new_function_test.go
```diff
@@ -5,37 +5,180 @@ package commands
...
+		faasCmd.SetArgs(cmdParameters)
 		faasCmd.Execute()
+	})
+
+	// Validate new function output
+	if !strings.HasPrefix(stdOut, ListOptionOutput) {
+		t.Fatalf("Output is not as expected: %s\n", stdOut)
+	}
 
+	if err := os.Chdir("../.."); err != nil {
```
alexellis: I don't think I like this because it's cryptic. Please can you explain what's happening?

guide/TEMPLATE.md
alexellis: That might be a good idea, but not a copy. Also s/alexellis/openfaas/g

guide/TEMPLATE.md
alexellis: Please update with the new syntax

alexellis: Let's use the template from @itscaro

proxy/deploy.go
```diff
@@ -24,16 +24,8 @@ func DeployFunction(fprocess string, gateway string, functionName string, image
...
-	} else if language == "node" {
-		fprocessTemplate = "node index.js"
-	} else if language == "ruby" {
-		fprocessTemplate = "ruby index.rb"
-	} else if language == "csharp" {
-		fprocessTemplate = "dotnet ./root.dll"
-	} else if language == "python3" {
-		fprocessTemplate = "python3 index.py"
+	} else {
+		fmt.Printf("Command to be invoked for function %s not found.\n", functionName)
```
alexellis: When would this error fire off? There is no hard need to provide an fprocess at the deployment time. Some Dockerfiles will bake this in.

stack/common.go
```diff
@@ -0,0 +1,31 @@
+package stack
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/url"
+)
+
+func ParseYAML(
+	file string,
```
alexellis: This doesn't read very well split over so many lines. How does it look on one line?

If there is a function being passed in - perhaps extract it to an interface?

stack/common.go
```diff
@@ -0,0 +1,31 @@
...
+	parseFunc func([]byte, ...string) (interface{}, error),
+	parseFuncArgs ...string,
+) (interface{}, error) {
```
alexellis: Why does this return interface{}?

stack/language_template.go
```diff
@@ -0,0 +1,58 @@
...
+	if object, err := iParseYAMLDataForLanguageTemplate(fileData, args...); err != nil {
+		return nil, err
+	} else {
+		return object.(*LanguageTemplate), nil
+	}
+}
+
+// iParseYAMLDataForLanguageTemplate parses YAML data into language template
+// Use the alias ParseYAMLDataForLanguageTemplate
+func iParseYAMLDataForLanguageTemplate(fileData []byte, args ...string) (interface{}, error) {
```
alexellis: What's the "i"-prefix convention?

stack/stack.go
```diff
@@ -19,28 +19,39 @@ import (
...
+		regex,
+		filter,
+	); err != nil {
+		return nil, err
 	} else {
-		fileData, err = ioutil.ReadFile(yamlFile)
-		if err != nil {
-			return nil, err
-		}
+		return object.(*Services), nil
```
alexellis: I'm not sure why we need to use untyped interfaces?

stack/language_template.go
ericstoekl: Because it returns an interface.

commands/fetch_templates.go
ericstoekl: This is just to name the `master.zip` file as the name of the repository. It's not necessary so I'm removing it.

stack/common.go
ericstoekl: What do you mean by extract to an interface?

stack/language_template.go
ericstoekl: Actually I'm going to try doing this without the interface approach.

stack/common.go
ericstoekl: I think the interface may not be necessary for what we're doing here, I might refactor it to something more simple

itscaro: It was for mutualise this function for parsing Stack & Language

stack/language_template.go
itscaro: Actually "i" was for "internal", I wanted not to expose it and use the public method to do the assertion on return type.

That's why ParseYAMLDataForLanguageTemplate calls iParseYAMLDataForLanguageTemplate.

This re-factoring is because I wanted to extract ParseYAMLFile to ParseYAML, since the part of parsing can be re-used for both LanguageTemplate & Stack

itscaro:
@ericstoekl I add a verif when downloading an invalid url, the zip is removed if it's not valid.
The command prints out now the list of fetched languages, same as when it does not overwrite
I updated the guide too

stack/language_template.go
ericstoekl: I am thinking we will add this re-factoring in a future PR, after this one is merged. The plan for this PR is to add minimal functionality.

stack/common.go
ericstoekl: We will add this back in the next re-factoring PR

commands/invoke.go
ericstoekl: Changed this one back.

proxy/deploy.go
ericstoekl: reverting this

stack/common.go
ericstoekl: Actually removing this entire file for now

commands/fetch_templates.go
ericstoekl: I have updated it to no longer delete before `--overwrite`.

commands/new_function_test.go
ericstoekl: We need to execute these tests in `faas-cli/commands/testdir/new_function` directory, so we need to `cd` into it for the test to run, then cd back to the home directory afterwards (because we don't know what order the tests will execute in).

I've changed this to save the full directory path at the beginning of the function, then `cd` back into it instead of using the relative `../..`.

ericstoekl:
Squashed the commits into 2 separate ones - First commit is everything before Alex's comments on 10/25, second commit is the changes to address those comments.

After the latest commit, the functionality is all the same, but the code has been changed in many places for simplification.

BurtonR:
Ran the tests on my machine and they all passed. 
Pulled @ericstoekl Bash template from https://github.com/ericstoekl/faas-custom-templates and made a simple bash function.

Tried the command without parameters and found it pulls the templates from this repo's templates. Something I didn't expect, but it didn't overwrite changes to the existing templates, so I don't see a problem with that.

Nice work!

ericstoekl:
Work in this PR is continued here: #201

PR #93, opened 2017-09-12
Make gofmt fail builds

itscaro:
## Description

* `gofmt` to break local and CI builds 

## Motivation and Context

- [x] I have raised an issue to propose this change (#95)

## Types of changes

Build changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,13 +15,13 @@ before_script:
 
 script:
   - go get -d -v
-  - ./build.sh
+  - make build
   - ./build_samples.sh
   - ./test/version.sh ./faas-cli
 #  - ./deploy_samples.sh # requires a deployed FaaS
 
 before_deploy:
-  - ./build_redist.sh
+  - make build_redist
 
 deploy:
   provider: releases
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: required
 language: go
 
 go:
-  - 1.7.x
+  - 1.9.x
 install:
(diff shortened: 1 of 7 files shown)


alexellis:
Hi @itscaro,

This is an interesting idea, thanks for looking into it. We're still using Golang 1.8 across the board until our main dependencies also move to Golang 1.9. I.e. Docker and Kubernetes. The reason the pull request template requires everyone to raise a proposal for significant changes is so that folks don't end up working on things which we can't include yet.

I'd be happy to take the change for `gofmt` if you can re-push it in a generic way - i.e. run it via the build_redist.sh file?

itscaro:
Since the go in travis is only used to do ```go get``` that why I bump it up directly from 1.7 to 1.9

I moved the ```gofmt``` verification into ```build.sh``` and ```build_redist.sh```

itscaro:
@alexellis should we use Makefile?

https://github.com/itscaro/faas-cli/blob/d4a511d375b0180f9f8ebabad8af14013e1a6f8b/Makefile

alexellis:
So let's start at the beginning again.

Can you add your "exit 1" code to these files?

https://github.com/alexellis/faas-cli/blob/master/Dockerfile#L7

https://github.com/alexellis/faas-cli/blob/master/Dockerfile.redist

itscaro:
@alexellis is it ok like this?

alexellis:
Merging! Thank you @itscaro for your patience on this one. We really value your input on the project, keep it up.

Issue #13, opened 2017-06-25
Proposal: parallel builds for stacks


alexellis:
Docker builds can be run in parallel rather than serial.

Pros:
* much quicker with several functions

Cons
* will be harder to debug

itscaro:
Example output (not yet optimized, suggestions are welcomed)

```
.\faas-cli build -f .\samples.yml --parallel 2
>>> Building alexellis/ruby-echo for ruby:ruby-echo
[ruby:ruby-echo] Clearing temporary build folder: ./build/ruby-echo/
>>> Building functions/resizer for Dockerfile:shrink-image
[Dockerfile:shrink-image] Building functions/resizer
[ruby:ruby-echo] Preparing ./sample/ruby-echo/ ./build/ruby-echo/function
[ruby:ruby-echo] Building alexellis/ruby-echo
2017/09/15 11:25:04 docker build -t functions/resizer .
2017/09/15 11:25:04 docker build -t alexellis/ruby-echo .
2017/09/15 11:25:18 Image: functions/resizer built.
<<< Built image functions/resizer for Dockerfile:shrink-image
>>> Building alexellis/faas-url-ping for python:url-ping
[python:url-ping] Clearing temporary build folder: ./build/url-ping/
[python:url-ping] Preparing ./sample/url-ping/ ./build/url-ping/function
[python:url-ping] Building alexellis/faas-url-ping
2017/09/15 11:25:18 docker build -t alexellis/faas-url-ping .
2017/09/15 11:25:22 Image: alexellis/ruby-echo built.
<<< Built image alexellis/ruby-echo for ruby:ruby-echo
>>> Building alexellis/faas-nodejs-echo for node:nodejs-echo
[node:nodejs-echo] Clearing temporary build folder: ./build/nodejs-echo/
[node:nodejs-echo] Preparing ./sample/nodejs-echo/ ./build/nodejs-echo/function
[node:nodejs-echo] Building alexellis/faas-nodejs-echo
2017/09/15 11:25:22 docker build -t alexellis/faas-nodejs-echo .
2017/09/15 11:25:39 Image: alexellis/faas-url-ping built.
<<< Built image alexellis/faas-url-ping for python:url-ping
2017/09/15 11:25:43 Image: alexellis/faas-nodejs-echo built.
<<< Built image alexellis/faas-nodejs-echo for node:nodejs-echo
```

alexellis:
I have merged the parallel build work, have you been able to test?
[...]

PR #107, opened 2017-09-14
Build in non-blocking execution, send stdOut & stdErr to logs/*.log

itscaro:
## Motivation and Context

- [x] I have raised an issue to propose this change (#13)

## Types of changes

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,9 @@
 **/node_modules/
 build/
-faas-cli
+logs/
 !contrib/bash/faas-cli
+
+faas-cli
 faas-cli-darwin
 faas-cli-armhf
 faas-cli.exe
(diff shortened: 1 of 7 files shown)


alexellis:
Thank you for looking at this. I have a slightly different approach which I'd like to share with you and I think may be more scaleable. For instance - if we are building 100 functions at once we'll kill the machine - so we need to use a sync.WaitGroup to throttle to 2-8 for instance.

itscaro:
Yeap I was wondering to use sync too and am going to do this today or this weekend ;)

alexellis:
Parallel building with a depth limit - https://github.com/alexellis/faas-cli/compare/parallel_build?expand=1

Issue #124, opened 2017-09-26
Print error message when --filter or --regex returns no matches


ericstoekl:
This change is to provide an error message if --filter or --regex returns no matches. This was mentioned by @johnmccabe in the Slack to be a desired feature.

## Expected Behaviour

When you run a --filter or --regex statement that returns no functions from the YAML file, return an error message like: `"No functions found in YAML, check --filter/--regex"`

## Current Behaviour

No error message is returned, so each command that uses --filter or --regex ends up throwing an unrelated error. `deploy` throws: `"Please provide a --image to be deployed."`

## Possible Solution

Return error when no functions remain after filter or regex loop has finished.

## Steps to Reproduce (for bugs)

1. `cd` into `faas-cli` directory
2. Run `faas-cli deploy -f samples.yml --filter alsdfhklasd`
3. See unrelated error message
4.

## Your Environment

* Docker version `docker version` (e.g. Docker 17.0.05 ):

* Are you using Docker Swarm or Kubernetes (FaaS-netes)?

* Operating System and version (e.g. Linux, Windows, MacOS):

* Link to your project or a code example to reproduce issue:

johnmccabe:
Derek add label: bug

johnmccabe:
Can you get rid of the 'proposal -' in the issue title please.

ericstoekl:
This is merged

PR #125, opened 2017-09-26
Print error message when no matches are found by --filter or --regex flag

ericstoekl:
Print error message when no matches are found by --filter or --regex flag

## Description

This change is to return an error message when all functions are filtered out by the --filter or --regex option. The current error message is `"No functions found in YAML, check --filter/--regex"`.

## Motivation and Context

Issue #124 

This will help users figure out what is wrong with their command if they expect functions to be deployed, and nothing happens.

## How Has This Been Tested?

Tests were upgraded to accommodate this change and check for this error to be properly thrown when --filter and --regex remove all functions.

## Types of changes

## Checklist:

### DIFF
--- a/stack/stack.go
+++ b/stack/stack.go
@@ -66,7 +66,6 @@ func ParseYAMLData(fileData []byte, regex string, filter string) (*Services, err
 			if regexExists {
 				match, err = regexp.MatchString(regex, function.Name)
 				if err != nil {
-					fmt.Printf("Invalid regex passed to --regex option\n")
 					return nil, err
 				}
 			} else {
@@ -76,6 +75,10 @@ func ParseYAMLData(fileData []byte, regex string, filter string) (*Services, err
 			if !match {
 				delete(services.Functions, function.Name)
 			}
+		}
+
+		if len(services.Functions) == 0 {
+			return nil, errors.New("No functions matching --filter/--regex were found in the YAML file")
 		}
 
 	}
--- a/stack/stack_test.go
+++ b/stack/stack_test.go
@@ -5,6 +5,7 @@ package stack
 import (
 	"reflect"
 	"sort"
+	"strings"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


ericstoekl:
Just added some extra test functionality, to test that when we expect an error message was returned, to verify that it is the correct error message.

stack/stack.go
```diff
@@ -66,7 +66,6 @@ func ParseYAMLData(fileData []byte, regex string, filter string) (*Services, err
 			if regexExists {
 				match, err = regexp.MatchString(regex, function.Name)
 				if err != nil {
-					fmt.Printf("Invalid regex passed to --regex option\n")
```
alexellis: This message was deleted?

alexellis:
Very good improvement to the UX here @ericstoekl

alexellis:
Can you include the expected messages for given error scenarios below?

stack/stack.go
ericstoekl: There is already an error thrown by the regex parser when the regex is invalid, which gets printed out by each command which calls `ParseYAMLFile()` by its following error checking block:

```
if err != nil {
    log.Fataln(err.Error())
    return
}
```

The line that `log.Fataln(err)` prints is something like 'invalid regexp', so I figured this printf is redundant. Also, for testing, I wanted to confirm the error message, and only have one line print out makes that easier.

ericstoekl:
Expected errors so far:

1. No matches found for `--filter` or `--regex` parameter: `No functions matching --filter/--regex were found in the YAML file`
2. Invalid regex used: `error parsing regexp: <details about how to fix regex>`

For example, if I use the command `faas-cli build -f samples.yml --regex [`, I get the message:

```
error parsing regexp: missing closing ]: `[`
```

These are the only two errors I've encountered so far related to `--filter` and `--regex`.

stack/stack.go
alexellis: As long as you bubble up the error.

alexellis:
Great. Can you now squash your commits or at least sign all of them?

```
git reset HEAD~2
git add .
git commit -s -m "COMMIT MESSAGE"
git push .... --force
```

stack/stack_test.go
```diff
@@ -48,175 +49,235 @@ const TestData_2 string = `provider:
...
-		title:       "Regex search for functions only containing 'node'",
-		searchTerm:  "node",
-		functions:   []string{"nodejs-echo"},
-		file:        TestData_1,
-		expectError: false,
+		title:            "Regex search for functions only containing 'node'",
+		searchTerm:       "node",
+		functions:        []string{"nodejs-echo"},
+		file:             TestData_1,
+		expectError:      false,
```
johnmccabe: Could get rid of the `expectedErrorMsg` and change `expectError` to a string, use `len(test.expectError) > 0` for the conditional - avoids having the redundant variable since you're always checking for the message.

stack/stack_test.go
```diff
@@ -233,28 +294,46 @@ func Test_ParseYAMLDataRegex(t *testing.T) {
...
-			sort.Strings(strkeys)
-			t.Log(strkeys)
-
-			if !reflect.DeepEqual(strkeys, test.functions) {
-				t.Errorf("Test_ParseYAMLDataFilter test [%s] failed, does not match expected result;\n  parsedYAML:   [%v]\n  expected: [%v]",
-					test.title,
-					strkeys,
-					test.functions,
-				)
+			if test.expectError {
```
johnmccabe: See earlier comment `if len(test.expectError) > 0 {`

stack/stack_test.go
ericstoekl: Thanks John, will update this evening.

ericstoekl:
Addressed the suggestions from @johnmccabe , and squashed the commits

alexellis:
LGTM

Issue #138, opened 2017-10-03
Fail "build" if Docker CLI throws an error


alexellis:
## Expected Behaviour

We need to "stop the world" if a `faas-cli build` throws an error. 

## Current Behaviour

Error is swallowed and hard to see.

## Possible Solution

At the least just `log.Fatal()` and emit non-zero exit code.

Better - print a message in red with the aec library we use for "version"/"new" commands.

## Steps to Reproduce (for bugs)

Use a Dockerfile with a bad instruction and do `faas-cli build`

shorsher:
I’ll give this one a shot 🙌🏼🙌🏼

alexellis:
Please provide output from testing in the PR (positive and negative) along with what happens if you have multiple functions or a single. GIFs are also very good for showing new behavior. Looking forward to seeing this.

shorsher:
So it seems the only way to get the error status is in ExecCommand(), would you like to handle any errors there?  The push command also utilizes ExecCommand() so that might be the best option.

alexellis:
I think you can check for Exit code on the command object

alexellis:
Fixed in 139. With some notes on a v2 implementation.

PR #139, opened 2017-10-04
Error checking for cli exec-commands

shorsher:
faas-cli will now check the exit status of exec commands
  (i.e. <docker build ...>) and will "stop the world" if
  a non-zero exit occurs.

Signed-off-by: Alex Shorsher <alex.shorsher@gmail.com>

## Description

Previously, even if `faas-cli build` failed when building a container from a Dockerfile, faas-cli would continue executing.  The correct behavior should be to stop faas-cli execution, log, and present the user with a noticeable error message.  

The issue originally just referenced throwing an error if `faas-cli build` fails, however, it seems my changes would also add error checking for other commands that use ExecuteCommand() (currently just `faas-cli push`).  I think this is acceptable?

## Motivation and Context

#138 

## How Has This Been Tested?

Here are some gif's of me running tests using the included samples.  They include a single function building  w/ a Dockerfile failing as expected, multiple functions with an expected failing Dockerfile, and one correct build using the samples.  Apologizing in advance for the crap gif quality.  

**Single function failure**
http://gph.is/2g98SlH
**Multi-function failure**
http://gph.is/2g9nmlx
**Passing example**
http://gph.is/2g7SZfs

## Types of changes

## Checklist:

### DIFF
--- a/builder/exec.go
+++ b/builder/exec.go
@@ -4,8 +4,12 @@
 package builder
 
 import (
+	"fmt"
+	"log"
 	"os"
 	"os/exec"
+
+	"github.com/morikuni/aec"
 )
 
 // ExecCommand run a system command
@@ -15,5 +19,9 @@ func ExecCommand(tempPath string, builder []string) {
 	targetCmd.Stdout = os.Stdout
 	targetCmd.Stderr = os.Stderr
 	targetCmd.Start()
-	targetCmd.Wait()
+	err := targetCmd.Wait()
+	if err != nil {
+		errString := fmt.Sprintf("ERROR - Could not execute command: %s", builder)
+		log.Fatalf(aec.RedF.Apply(errString))
+	}
 }


builder/exec.go
```diff
@@ -4,8 +4,12 @@
 package builder
 
 import (
+	"fmt"
+	"log"
 	"os"
 	"os/exec"
+
+	"github.com/morikuni/aec"
```
johnmccabe: What OS/Terminals have you validated this on? Mainly have you checked what it looks like on Windows Cmd and Powershell?

alexellis: I added `aec` for version/new commands which rendered on Windows cmd + PowerShell on Windows 10. Would be good to have it checked on one of your VMs if you have time.

shorsher: I’ve only tested in macOS in iTerm and Terminal. I don’t have any Linux vm’s on hand but can spin one up if necessary.

shorsher:
Tested aec coloring on Ubuntu 16.04, here's an attached screenshot.  
screen shot 2017-10-04 at 11 53 17 pm

alexellis:
Thanks. Colouring works well on Bash. It tends to be Windows that can have problems we already have similar code in the CLI so we need to see how it is there. If you don't have access we can see if someone else does in the channel.

alexellis:
Will take this change as it's better than what we had - but I think the error checking needs to be higher up in the code - so we are perhaps just setting the nonzero exit code and printing a message. In retrospect log.Fatal might be better after the other functions have had a chance to build.

shorsher:
I can get started on v2 and submit another PR

alexellis:
Great

alexellis:
Thank you

Issue #128, opened 2017-09-29
When creating new functions - offer to append to existing YAML


alexellis:
Here's the scenario:

I scaffold 5 functions, but end up with 5 YAML files.

```
faas-cli new --lang python --name func1
func1.yml created
faas-cli new --lang python --name func2
func2.yml created
faas-cli new --lang python --name func3
func3.yml created
```

Let's offer a new parameter `--yaml YAML_FILE` which means you can append new functions into an existing file:

```
faas-cli new --lang python --name func1
func1.yml created
faas-cli new --lang python --name func2 --append func1.yml
func1.yml appended
faas-cli new --lang python --name func3  --append func1.yml
func1.yml appended
```

rgee0:
I think this is a great idea.  Making it easier for people to build and maintain a suite of functions.

So that others are clear, is the suggestion that the new parameter should be `--yaml` or `--append`?  The narrative and codeblock seem to suggest both.  I'd lean more towards `--append` as it would be more intuitive a command.

ericstoekl:
I'm working on this ATM, appending the new function to the `Services` struct and then using  `yaml.Marshall()` to write out the modified YAML file. I gotta go to an all-day event today but I should have it done tomorrow morning (Sunday oct 1).

Also plan on adding some tests

ericstoekl:
Alright, so I've got the basic functionality working in this branch. This one doesn't leverage the `Services` interface to append the data, it just writes string data to a file. I want to get maintainers' opinions on whether we should append by adding to a `*Services` object then writing that out with `yaml.Marshall()`, or if this way is sufficient.

https://github.com/ericstoekl/faas-cli/commit/eccf237c9dcb5190c3725e7ce3d654344a1028a4

Gif of the functionality:
https://imgur.com/a/Zggm3

rgee0:
My preference would be to use `yaml.Marshall()` as I feel it keeps things cleaner.

alexellis:
What resulting template do you get for Lang Dockerfile?

ericstoekl:
@alexellis You get a folder containing an empty file named `Dockerfile`. I was thinking putting something very basic in, like `catservice` as a starting point, might be helpful for users.

alexellis:
See also #132

alexellis:
@ericstoekl did you PR for this yet?

alexellis:
[...]

PR #141, opened 2017-10-04
[WIP] Add append functionality to `faas-cli new` command

ericstoekl:
Appends function config YAML data to existing YAML file, specified by:
`faas-cli new <FuncName> --lang <Lang> -f <YAML>`

This change allows for merging the YAML config data for a new function into an existing YAML file. It also adds tests (7 test cases) for the `new` function.

## Description

Instead of appending text directly to the file, we open the file with `stack.ParseYAMLFile`, append data to the returned `stack.Services` struct, then write the data back to the file with `stack.WriteYAMLData` (a new function for the `stack` package).

## Motivation and Context

#128 

## How Has This Been Tested?

Added `new_function_test.go` which runs several test cases against the function, testing its original and new functionality.

## Types of changes

## Checklist:

### DIFF
--- a/commands/new_function.go
+++ b/commands/new_function.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/morikuni/aec"
 	"github.com/openfaas/faas-cli/builder"
+	"github.com/openfaas/faas-cli/stack"
 	"github.com/spf13/cobra"
 )
 
@@ -25,6 +26,7 @@ func init() {
 	newFunctionCmd.Flags().StringVar(&lang, "lang", "", "Language or template to use")
 	newFunctionCmd.Flags().StringVar(&gateway, "gateway", defaultGateway,
 		"Gateway URL to store in YAML stack file")
+	newFunctionCmd.Flags().StringVar(&image, "image", "", "Name for docker image")
 
 	newFunctionCmd.Flags().BoolVar(&list, "list", false, "List available languages")
 
@@ -60,22 +62,48 @@ the "Dockerfile" lang type in your YAML file.
 		return
 	}
 
-	if len(args) < 1 {
-		fmt.Println("Please provide a name for the function")
-		return
+	// Support legacy '--name' usage for function name
+	if len(functionName) == 0 {
+		if len(args) < 1 {
+			fmt.Println("Please provide a name for the function")
+			return
+		}
+		functionName = args[0]
 	}
-	functionName = args[0]
 
 	if len(lang) == 0 {
 		fmt.Println("You must supply a function language with the --lang flag")
 		return
 	}
 
-	PullTemplates("")
+	var stackFileName string
+	var services stack.Services
+	if len(yamlFile) == 0 {
+		// We will create a new YAML file for this function
+		stackFileName = functionName + ".yml"
+	} else {
+		// YAML file was passed in, so parse to see if it is valid
+		parsedServices, err := stack.ParseYAMLFile(yamlFile, "", "")
+		if err != nil {
+			fmt.Printf("Specified file (" + yamlFile + ") is not valid YAML\n")
+			return
+		}
+
+		services = *parsedServices
+		// Make sure that services.Functions is not a nil map
+		if len(services.Functions) == 0 {
+			fmt.Printf("Error in YAML file - `functions` map is empty\n")
+			return
+		}
+
+		stackFileName = yamlFile
+	}
 
 	if validTemplate(lang) == false {
 		fmt.Printf("%s is unavailable or not supported.\n", lang)
 	}
+
+	PullTemplates("")
 
 	if _, err := os.Stat(functionName); err == nil {
 		fmt.Printf("Folder: %s already exists\n", functionName)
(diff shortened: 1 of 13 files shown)


commands/new_function_test.go
```diff
@@ -0,0 +1,171 @@
...
+		os.Remove("template")
+	}()
+
+	// Change directory to testdata
+	if err := os.Chdir("testdata/new_function"); err != nil {
+		t.Fatalf("Error on cd to testdata dir: %v", err)
+	}
+
+	// Symlink template directory at root to current directory to avoid re-downloading templates
+	os.Symlink("../../../template", "template")
```
alexellis: I think this could be asking for trouble. Can we do a `cp` or just generate a minimum set of files and have that as part of the tests?

stack/schema.go
```diff
@@ -7,13 +7,13 @@ package stack
...
 	GatewayURL string `yaml:"gateway"`
-	Network    string `yaml:"network"`
+	Network    string `yaml:"network,omitempty"`
 }
 
 // Function as deployed or built on FaaS
 type Function struct {
 	// Name of deployed function
-	Name     string `yaml:"-"`
+	Name     string `yaml:"-,omitempty"`
```
alexellis: Does this re-serialize correctly? I had some issues getting this to map back in the correct way.

stack/stack.go
```diff
@@ -103,3 +103,16 @@ func fetchYAML(address *url.URL) ([]byte, error) {
 
 	return resBytes, err
 }
+
+func WriteYAMLData(services *Services, fileName string) error {
```
alexellis: Have we got any tests on the serialization on its own?

stack/schema.go
ericstoekl: That's strange, what kind of issues did you experience? The `omitempty` is there so that if the `Name` property is left out of the `Function` struct instantiation, it will not be Marshalled to YAML. Without `omitempty`, the YAML file will end up with a `name: ""` property.

stack/stack.go
ericstoekl: No but I will add some.

commands/new_function_test.go
ericstoekl: Sure, I'll copy the folder over instead of using `os.Symlink`.

alexellis:
What if we wrote a minimal example in the folder structure instead? - we are only testing that the YAML gets appended - not strictly caring about the templated function itself.

ericstoekl:
Can you clarify what you mean by "minimal example in the folder structure"? Should the test be testing only parseYAML/WriteYAML, and not doing full `faasCmd.Exec()` calls to run the whole `new` function?

alexellis:
We're copy in a whole set of templates. Could we not just write a template folder with the minimum text to make the process work?

ericstoekl:
Gotcha, thanks for clarifying. I'll use a minimal templates directory.

ericstoekl:
Added tests for the `WriteYAMLData` function.

Leaving this un-squashed for now for easier reviewing (IMO), let me know if you have any other questions/issues.

stack/stack.go
itscaro: @ericstoekl can you rename it to WriteYAMLDataForServices, in #87 I made some changes, ParseYAMLData is for general parsing, and ParseYAMLDataForServices is used for services

ericstoekl: Thanks for the heads up @itscaro. If #87 gets merged before this one, I'll make that update.

developius:
What's the status on this? @ericstoekl @alexellis

ericstoekl:
Should be good to go in my opinion, but could use some code review + testing by someone who isn't me :)

developius:
Looks good & works fine well - tested using the `samples.yml` file. My only comments would be that it will remove reformat the yaml file (including removing comments & reordering definitions). Not a big deal but would be nice for this not to happen (think `git diff` etc).

ericstoekl:
@developius Thanks for testing. @alexellis , like Finnian said the ordering of the yaml file will be potentially re-arranged when you append to it. This is because I use `yaml.Marshal(services)` in the `WriteYAMLData()` function, (https://github.com/ericstoekl/faas-cli/blob/0bfb0c8aa282978ebce326d277b8e65759b9eb11/stack/stack.go#L113)

This will take the `services` object, which has a schema here (https://github.com/ericstoekl/faas-cli/blob/0bfb0c8aa282978ebce326d277b8e65759b9eb11/stack/schema.go#L46). This enforces that the `provider:` section in the YAML comes before the `functions:` section. The ordering of the `Function` struct is also enforced in the YAML.

The result is that if the user has an ordering that deviates from the order that the schema has, that ordering will be re-arranged. Also, if the user has whitespace in their YAML file, it will all be removed.

Let me know if this is acceptable. If not, we can implement the append feature by making sure that the `function:` section is last, then write string data of the new function object to the end of the YAML file.

alexellis:
This is tricky.. I know that marshalling the file is the technically correct thing to do, but appending text may be a better user experience as @developius has a good point on the comments and whitespace.

I'd like to get an opinion from @nicholasjackson and @stefanprodan

nicholasjackson:
So I am kinda on the fence, I am pro fixing formatting and whitespace as GoFmt does and the other benefit is that using YAML parser will give early feedback that the user has borked their config.  Loosing comments however is a big one, personally I would never add comments to the file so it would be a minor problem for me however it may cause an issue to the user.  This could be avoided with a message in the help, "Appends new function to existing config, this operation will reformat your existing yaml including the removal of comments.

The other option could be as simple as read the file with the YAML parser to check correctness, use the yaml parser to create the new template, merge the two by string concatenation.  That way I think everyone wins.

developius:
I'm all for @nicholasjackson's solution - read the file and parse to check integrity, add the new function via string concatenation and then parse again to check.

alexellis:
@nicholasjackson the resulting ordering is also "borked" - you can end up with the provider section at the bottom for instance.

alexellis:
@ericstoekl we have consensus on appending rather than re-writing the whole file. Would you like to open a new PR for the work?

ericstoekl:
@alexellis shouldn't be too much of a problem to do the work in this PR, as the tests won't change too much. Is that acceptable?

ericstoekl:
My plan for the next PR to solve this issue (would appreciate feedback if this approach is correct or not @alexellis ):

1) Parse existing YAML, make sure it's OK
2) Create `Function` struct with data for the new function user wants to create
3) Read existing YAML file line-by-line, until end of `functions:` section (until another section is found or until EOF)
4) Insert struct created in step 2 into YAML file at this point

Does that sound reasonable? I'm working on it right now, will have a new PR soon.

developius:
Sounds good @ericstoekl but I'd add a 5th step to re-validate the yaml before inserting, just to be safe

commands/new_function.go
```diff
@@ -25,6 +26,7 @@ func init() {
 	newFunctionCmd.Flags().StringVar(&lang, "lang", "", "Language or template to use")
 	newFunctionCmd.Flags().StringVar(&gateway, "gateway", defaultGateway,
 		"Gateway URL to store in YAML stack file")
+	newFunctionCmd.Flags().StringVar(&image, "image", "", "Name for docker image")
```
alexellis: While an interesting change, this shouldn't be part of the PR.

alexellis: @ericstoekl please see this feedback.

ericstoekl: I've got a new PR on the way for this feature, which addresses this. I'll close that when it's ready.

ericstoekl:
Re-done with new PR - #213

Issue #22, opened 2017-07-24
CLI should support private registries


alexellis:
Private registries are supported by the FaaS REST API - https://github.com/alexellis/faas/blob/master/docs/managing-images.md

The support hasn't been added to the CLI / YAML yet.

alexellis:
See also - https://github.com/alexellis/jaas/commit/0aa57a8b8f057ded6a276d0674d5facb466464b1

PR #181, opened 2017-10-20
[WIP] Add external Docker registry authentication support via arguments/YAML

nenadilic84:
## Description

An additional argument has been added in order to pass information to the FaaS gateway in order to authenticate the docker client.
## Motivation and Context

In order to enable FaaS gateway to pull images from Docker Registries that require logins

It implements the issue #22 
More info: https://github.com/openfaas/faas/blob/master/docs/managing-images.md

## How Has This Been Tested?

I was able to pull an image from a private docker registry with 

```
docker login 
faas-cli deploy --image=test_img --name=test_fn  --gateway=https://remote-site.com:8080 --lang=python
```

## Types of changes

## Checklist:


alexellis:
Thank you. Is there a way we can access the keychain on Mac/Windows? Like the Docker CLI?

alexellis:
Hi this appears to break the CI.

alexellis:
Derek add label: design/review

nenadilic84:
@alexellis, you mean to use the `.docker/config.json`, match the registry URL and retrieve the `user:pass` from the `credsStore` if enabled, otherwise from the `auth ` field in the `.docker/config.json`?
If so, I'll try to add this.

nenadilic84:
@alexellis, updated the PR as agreed. It uses the `credsStore ` to access the `user:pass` as described. 

Let me know if you have any questions.
Note that I've tested this only on Mac...

nenadilic84:
BTW, I had to add `github.com/docker/docker-credential-helpers/client` and `github.com/mitchellh/go-homedir` but `travis` seems to have some problems with that 😄

nenadilic84:
I just realized that the Dockerfile is missing the `vndr` in order to fetch the packages. @alexellis should this a separate issue ticket?

nenadilic84:
As all other packages were checked in git I've also added the one I've added to the project. However, I suggest we don't check-in them directly but rather use `vndr` to get them with the specific `SHA`... Let me know if you want me to open a ticket.

PS. the openfaas/faas version in `vendor.conf` was an old one and didn't have `constraints` and `secrets` in `CreateFunctionRequest` struct so I had to bump the version there...

nenadilic84:
@stefanprodan, FYI

stefanprodan:
My guess is no one will opt for passing the  `--registry_auth=base64` since `faas-cli push` will fail if you don't have the registry auth in docker config. I would do this silent, on `deploy` lookup the registry auth, if it's not there move on

alexellis:
For Kubernetes we declare an image pull secret in the cluster and then pass the name of the secret we want via CLI or YAML.

The option `--registry_auth=base64_goes_here` enables people to deploy images from a private registry on Docker Swarm. It was a missing piece of the work by some guys from AT&T who hacked on it.

johnmccabe:
This looks reasonable and works for me pushing to a private repo, deploy works to a local stack - but its not clear how this is intended to work when deploying to remote OpenFaaS stacks?

@alexellis is that being handled elsewhere, or should it be covered here as well?

nenadilic84:
@johnmccabe, what exactly you meant by "remote OpenFaaS stacks" are you referring to the remote OpenFaas Gateway. If so this is done by the `gateway` parameter either in yaml file or as an argument.

johnmccabe:
@nenadilic84 yep I'm aware of the gateway parameter, I'm talking about auth for the pull from a private registry when connected to a remote gateway.

nenadilic84:
Yup, and this should be working as well. We are providing the user:pass to the gateway so no matter if it is local or remote, gateway will be capable to pull the image...

Let me know if you need a bit more of clarification on this...

johnmccabe:
@nenadilic84 gah, error between keyboard and chair... had dropped the `./` when running the deploy. Works against a remote swarm.

One concern tho, should we prevent the submission of the auth when connected to a non SSL-secured Gateway endpoint?

stefanprodan:
In #182 I'm issuing a warning if the gateway is not on HTTPS and advise on using Letsencrypt like this:

```
	if res.TLS == nil {
		fmt.Println("WARNING! Communication is not secure, please consider using HTTPS. Letsencrypt.org offers free SSL/TLS certificates.")
	}
```

If we decide to block HTTP we should do this in both cases, but I would go with a warning. For example I have prod setups where I am using a VPN from the CI to Prod registry, it's not HTTPS but it's a secure connection.

commands/deploy.go
```diff
@@ -67,7 +74,8 @@ var deployCmd = &cobra.Command{
                   [--constraint PLACEMENT_CONSTRAINT ...]
                   [--regex "REGEX"]
                   [--filter "WILDCARD"]
-				  [--secret "SECRET_NAME"]`,
+				  [--secret "SECRET_NAME"]
```
johnmccabe: Can you fix the indentation while you're changing this, `--replace` and `--update` also need fixed above this.

nenadilic84:
Tracking remaning task 😄 
- [ ] Resolve conflicts
- [ ] Add warning if no TLS
- [ ] Fix indentation

nenadilic84:
@johnmccabe, I've updated the PR. Let me know if there is anything else I can do here...

alexellis:
> Would likely only want to do so when the image being deployed is in a private repo however, and not for public images? wdyt?

What does the Docker CLI do?

alexellis:
Note from @stefanprodan 
> My guess is no one will opt for passing the --registry_auth=base64 since faas-cli push will fail if you don't have the registry auth in docker config. I would do this silent, on deploy lookup the registry auth, if it's not there move on

Is the approach we used?

nenadilic84:
So we still have the `--registry_auth=base64`, but if not passed we try to find the `auth` from docker config. I can remove the `--registry_auth=base64`. In the current setup, I'm not using it and also not sure if someone will use it as it has a convenience of relying on `docker login` anyway.

As for the TLS, we are doing as @stefanprodan suggested, just a warning in case the connection is over HTTP instead of HTTPS...

commands/deploy.go
```diff
@@ -277,3 +300,87 @@ func compileEnvironment(envvarOpts []string, yamlEnvironment map[string]string,
...
+	configFileName        = "config.json"
+	configFileDir         = ".docker"
+	defaultDockerRegistry = "https://index.docker.io/v1/"
+)
+
+var (
+	configDir = os.Getenv("DOCKER_CONFIG")
+)
+
+func readDockerConfig(config *configFile) error {
```
alexellis: If this is copied from the Docker CLI it should probably be in a new package (i.e. sub folder and package)

alexellis:
OK - let's remove the flag, less is more and we can add it in if needed later on.

commands/deploy.go
nenadilic84: It was influenced by this: https://github.com/docker/cli/blob/master/cli/config/credentials/native_store.go, but it uses a different approach to read the creds and add them to the `config.AuthConfigs[n].Auth` for later use

nenadilic84: It uses the github.com/docker/docker-credential-helpers/credentials package

nenadilic84:
@alexellis, @johnmccabe I've updated the PR ^
Let me know if there is anything else that needs to be done here

alexellis:
Where are we with this PR?

nenadilic84:
IMO, ti just needs a rebase, otherwise I've added all the things we agreed upon.
Also, would like to mention that I've been using this for a while now...

nenadilic84:
@alexellis, should I do the rebase? Are we going to merge this?

alexellis:
@nenadilic84 @stefanprodan I'm doing admin on old PRs / requests. Please can you give me a summary on where this is?

nenadilic84:
@alexellis, if you ask me, I would rebase this and merge it. I've been using it for a while now and find it quite useful. Of course, I need to rebase it atm.

@stefanprodan, if you don't have anything to add let's go forward with it...

proxy/deploy.go
```diff
@@ -36,21 +37,26 @@ func DeployFunction(fprocess string, gateway string, functionName string, image
...
-		Image:       image,
-		Network:     network,
-		Service:     functionName,
-		EnvVars:     envVars,
-		Constraints: constraints,
-		Secrets:     secrets, // TODO: allow registry auth to be specified or read from local Docker credentials store
-		Labels:      &labels,
+		EnvProcess:   fprocessTemplate,
+		Image:        image,
+		RegistryAuth: registryAuth,
```
itscaro: I don't think that we need to store credentials in the function like this, for security reason. We might implement a method on gateway so that the gateway can register credentials to docker on the gateway.

stack/schema.go
```diff
@@ -22,6 +22,9 @@ type Function struct {
 	// Image Docker image name
 	Image string `yaml:"image"`
 
+	// Docker registry Authorization
+	RegistryAuth string `yaml:"registry_auth,omitempty"`
```
itscaro: I don't think that we need to store credentials in the function like this, for security reason. We might implement a method on gateway so that the gateway can register credentials to docker on the gateway.

nenadilic84: We are not storing the credentials here, this is just an interface to gateway. The only security problem here would be if the gateway is not using TLS and the `registry_auth` is sent in the plain text. This is why we added a warning about it in order to advise the user to use https instead. The credentials are stored by `docker-cli` and we are just retrieving it... Let me know if you need more info...

itscaro: The credentials is stored in stack.yml, isn't it?

stefanprodan: @itscaro the credentials are not stored anywhere by us, these credentials are fetch from Docker local storage and send over to the gateway.

nenadilic84: I don't this is representing the stack.yaml, but the API interface to the gateway. That being said things might have changed since I last looked it in.

alexellis:
@nenadilic84 @stefanprodan @johnmccabe @itscaro 

Had a ping from a user on Twitter about this PR today: https://twitter.com/xesin97/status/975033191592611840

You guys were all involved in this work, so please can you give me your take?

vendor.conf
```diff
@@ -1,8 +1,9 @@
-github.com/openfaas/faas 9a70a1b9915d7420e71dda445349d68f3f3928d8
+github.com/openfaas/faas 1f0476448118d501b7861c3df3f660f305afe493
 gopkg.in/yaml.v2 cd8b52f8269e0feb286dfeef29f8fe4d5b397e0b
 github.com/spf13/cobra 2df9a531813370438a4d79bfc33e21f58063ed87
 github.com/spf13/pflag e57e3eeb33f795204c1ca35f56c44f83227c6e6
 github.com/ryanuber/go-glob 256dc444b735e061061cf46c809487313d5b0065
 github.com/morikuni/aec 39771216ff4c63d11f5e604076f9c45e8be1067b
 github.com/inconshreveable/mousetrap 76626ae9c91c4f2a10f34cad8ce83ea42c93bb75
+github.com/docker/docker-credential-helpers d68f9aeca33f5fd3f08eeae5e9d175edf4e731d1
```
alexellis: @nenadilic84 @itscaro does this mean having to turn on CGO?

xesina:
As @alexellis mentioned I have a problem closely related to this issue. When I deploying a func with an image on private registry I getting 'no basic auth credentials' error on docker log.
But when I manually update swarm service to have --with-registry-auth it works as expected and can pull image from private registry.

So I need to either cli tells gateway to add this param or somehow pass credentials when creating service

xesina:
I had similar issue with go-dockerclient that solved with little contribution:
https://github.com/fsouza/go-dockerclient/pull/636

alexellis:
@xesina would you be interested in picking up this work from @nenadilic84 ? 

@nenadilic84 thanks for signing off. Looks like we need a rebase again.

BurtonR:
I'll pick this up this weekend and get it tested and ready to go if @xesina doesn't beat me to it

xesina:
@BurtonR Nice 👍

BurtonR:
Derek: close 
See the updated (rebased) PR [here](https://github.com/openfaas/faas-cli/pull/372) #372

Issue #178, opened 2017-10-18
Proposal Enable basic auth for CLI


alexellis:
## Expected Behaviour

Pass username/password to remote calls - deploy/list/invoke etc.

## Current Behaviour

Need to use API / curl.

## Possible Solution / Design

We will need to discuss and review a design here before moving forward.

**Options for reading:**

- Pass a parameter `password` - should not follow this approach due to bash cache history
- Read from stdin - resolves issue with parameter but is hard to script and repetitive

A combination of both approaches may involve creating a hash of the username/password and storing that in an ~/.openfaas/config.json file. For Windows/Mac the native keychain would be a better place to store sensitive data.

**Other concerns:**

Multiple gateways are supported so we need to find a way to accept credentials or a token per gateway.

## Context

Supports blog post for WeaveWorks

Lewiscowles1986:
Should this wait for #165 and #166? I definitely feel like #166 (check issue for branch merged with #165 current works) provides a more centralised place to locate logic

Lewiscowles1986:
Btw Happy to work on this. (I take it, this would be optional and use cli arguments?)

alexellis:
I've updated the design brief and commented on #180. Still needs to have the design reviewed.

Lewiscowles1986:
what about using an env var instead of a file? It seems cleaner and isn't polluting the FS with secrets (which seems to be the concern over cli history). For now, it empowers users to come up with their own mechanism, and we get the functionality without painting into a corner?

stefanprodan:
Hi,

I would suggest adding a `login` command that would validate and store the authentication data on disk, and a middleware for the API client that would fetch the user/pass from the auth store based on the gateway URI.

The `login` command would have the following signature:

```
./faas-cli login --help
Log in to OpenFaaS gateway.
If no gateway is specified, the default local one will be used.

Usage:
  faas-cli login [--username USERNAME] [--password PASSWORD] [--gateway GATEWAY_URL] [flags]

Examples:
  faas-cli login -u user -p password --gateway https://localhost:8080
  cat ~/faas_pass.txt | faas-cli login -u user --password-stdin --gateway https://openfaas.com

Flags:
      --gateway string    Gateway URI (default "http://localhost:8080")
  -h, --help              help for login
  -p, --password string   Gateway password
      --password-stdin    Reads the gateway password from stdin
  -u, --username string   Gateway username
```

[...]

PR #180, opened 2017-10-18
Initial Basic Auth works

Lewiscowles1986:
Initial Work on Basic Auth

## Description

 * Added helper to proxy with tests for basic auth
 * Deploy
 * Delete (because required by Deploy)

## Motivation and Context

- [x] ~~I have~~ raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

See #178 

## How Has This Been Tested?

Improves coverage of proxy by 2.6%, improves coverage by 1.1%. Two tests located in proxy for the new method added. (Should work in any Go repo using HttpRequest). Canned data used is from Go internal tests for `SetBasicAuth`. Added to proxy as it's really more of an implementation detail where it's used.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (some, but needs work, unsure how to communicate best)
- [x] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] I have signed-off my commits with `git commit -s
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #182, opened 2017-10-22
[Testing] Support Basic Auth for multiple gateways

stefanprodan:
## Description

This PR adds a new command named `login` that allows users to register multiple gateways protected with basic auth. The credentials are being saved to `$HOME/.openfaas/config.yml` on Linux/macOS and `%USERPROFILE%/.openfaas/config.yml` on Windows. The `yml` extension is used to avoid overriding the openfaas-bitbar `config.yaml`. 

CLI `login` command specs:

* validates the credentials by calling `list` for a given gateway
* issues a warning if the password is passed via command line and advise on using stdin
* issues a warning if the gateway is not on HTTPS and advise on using Letsencrypt
* creates or updates credentials for a given gateway
* encodes the username and password with base64 for local storage
* sets auth to `basic` type in the YAML auth store

Upon calling a gateway API endpoint, the config store lookups a matching gateway URL entry. If an entry is found, the username and password are being decoded from base64 and used to setup the basic auth header. All API calls that return a 401 error will point the user to the `login` command.

CLI `login` command help:

```
./faas-cli login -h
Log in to OpenFaaS gateway.
If no gateway is specified, the default local one will be used.

Usage:
  faas-cli login [--username USERNAME] [--password PASSWORD] [--gateway GATEWAY_URL] [flags]

Examples:
  faas-cli login -u user -p password --gateway http://localhost:8080
  cat ~/faas_pass.txt | faas-cli login -u user --password-stdin --gateway https://openfaas.mydomain.com

Flags:
      --gateway string    Gateway URI (default "http://localhost:8080")
  -h, --help              help for login
  -p, --password string   Gateway password
      --password-stdin    Reads the gateway password from stdin
  -u, --username string   Gateway username
```

CLI `logout` command help:

```
./faas-cli logout --help
Log out from OpenFaaS gateway.
If no gateway is specified, the default local one will be used.

Usage:
  faas-cli logout [--gateway GATEWAY_URL] [flags]

Examples:
  faas-cli logout --gateway https://openfaas.mydomain.com

Flags:
      --gateway string   Gateway URI (default "http://localhost:8080")
  -h, --help             help for logout
```

## Motivation and Context
- [x] I have raised an issue to propose this change (see #178)

## How Has This Been Tested?

Tested against OpenFaaS (gateway v0.6.5) and Caddy as reverse proxy with basic auth and TLS. 

Gateway Docker Swarm test environments:
* GCP Compute Engine
* Ubuntu 16.04
* Docker Swarm CE 17.09.0-ce and Docker Swarm EE 17.06.2-ee-3
* Caddy 0.10.9 [FOSS edition](https://hub.docker.com/r/stefanprodan/caddy/)
* Letsencrypt certificate

Gateway Kubernetes test environment:
* Google Container Engine
* Kubernetes 1.8.1
* Caddy 0.10.10 [FOSS edition](https://hub.docker.com/r/stefanprodan/caddy/)

CLI test environments:
* macOS Siera 10.12.6
* Ubuntu 16.04
* Windows 10 Pro (go 1.9.2 windows/amd64)

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ The main commands supported by the CLI are:
 * `faas-cli deploy` - deploys the functions into a local or remote OpenFaaS gateway
 * `faas-cli remove` - removes the functions from a local or remote OpenFaaS gateway
 * `faas-cli invoke` - invokes the functions and reads from STDIN for the body of the request
+* `faas-cli login` - stores basic auth credentials for OpenFaaS gateway (supports multiple gateways)
+* `faas-cli logout` - removes basic auth credentials fora given gateway
 
 Help for all of the commands supported by the CLI can be found by running:
 
@@ -51,7 +53,7 @@ You can chose between using a [programming language template](https://github.com
 
 **Templates**
 
-Command: `faas-cli new FUNCTION_NAME --lang python/node/go//ruby/Dockerfile/etc`
+Command: `faas-cli new FUNCTION_NAME --lang python/node/go/ruby/Dockerfile/etc`
 
 In your YAML you can also specify `lang: node/python/go/csharp/ruby`
 
--- a/commands/login.go
+++ b/commands/login.go
@@ -1,0 +1,141 @@
+// Copyright (c) OpenFaaS Project 2017. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package commands
+
+import (
+	"crypto/tls"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/openfaas/faas-cli/config"
+	"github.com/spf13/cobra"
+)
+
+var (
+	username      string
+	password      string
+	passwordStdin bool
+)
+
+func init() {
+	loginCmd.Flags().StringVar(&gateway, "gateway", defaultGateway, "Gateway URI")
+	loginCmd.Flags().StringVarP(&username, "username", "u", "", "Gateway username")
+	loginCmd.Flags().StringVarP(&password, "password", "p", "", "Gateway password")
+	loginCmd.Flags().BoolVar(&passwordStdin, "password-stdin", false, "Reads the gateway password from stdin")
+
+	faasCmd.AddCommand(loginCmd)
+}
+
+var loginCmd = &cobra.Command{
+	Use:   `login [--username USERNAME] [--password PASSWORD] [--gateway GATEWAY_URL]`,
+	Short: "Log in to OpenFaaS gateway",
+	Long:  "Log in to OpenFaaS gateway.\nIf no gateway is specified, the default local one will be used.",
+	Example: `  faas-cli login -u user -p password --gateway http://localhost:8080
+  cat ~/faas_pass.txt | faas-cli login -u user --password-stdin --gateway https://openfaas.mydomain.com`,
+	RunE: runLogin,
+}
+
+func runLogin(cmd *cobra.Command, args []string) error {
+
+	if len(username) == 0 {
+		return errors.New("must provide --username or -u")
+	}
+
+	if len(password) > 0 {
+		fmt.Println("WARNING! Using --password is insecure, consider using: cat ~/faas_pass.txt | faas-cli login -u user --password-stdin")
+		if passwordStdin {
+			return errors.New("--password and --password-stdin are mutually exclusive")
+		}
+
+		if len(username) == 0 {
+			return errors.New("must provide --username with --password")
+		}
+	}
+
+	if passwordStdin {
+		if len(username) == 0 {
+			return errors.New("must provide --username with --password-stdin")
+		}
+
+		passwordStdin, err := ioutil.ReadAll(os.Stdin)
+		if err != nil {
+			return err
+		}
+
+		password = strings.TrimSpace(string(passwordStdin))
+	}
+
+	password = strings.TrimSpace(password)
+	if len(password) == 0 {
+		return errors.New("must provide a non-empty password via --password or --password-stdin")
+	}
+
+	fmt.Println("Calling the OpenFaaS server to validate the credentials...")
+	gateway = strings.TrimRight(strings.TrimSpace(gateway), "/")
+	if err := validateLogin(gateway, username, password); err != nil {
+		return err
+	}
+
+	if err := config.UpdateAuthConfig(gateway, username, password); err != nil {
+		return err
+	}
+
+	user, _, err := config.LookupAuthConfig(gateway)
+	if err != nil {
+		return err
+	}
+	fmt.Println("credentials saved for", user, gateway)
+
+	return nil
+}
+
+func validateLogin(url string, user string, pass string) error {
+	// TODO: provide --insecure flag for this
+	tr := &http.Transport{
+		DisableKeepAlives: true,
+		TLSClientConfig:   &tls.Config{InsecureSkipVerify: true},
+	}
+	client := &http.Client{
+		Transport: tr,
+		Timeout:   time.Duration(5 * time.Second),
+	}
+
+	// TODO: implement ping in the gateway API and call that
+	gatewayUrl := strings.TrimRight(url, "/")
+	req, _ := http.NewRequest("GET", gateway+"/system/functions", nil)
+	req.SetBasicAuth(user, pass)
+
+	res, err := client.Do(req)
+	if err != nil {
+		fmt.Println(err)
+		return fmt.Errorf("cannot connect to OpenFaaS on URL: %s", gatewayUrl)
+	}
+
+	if res.Body != nil {
+		defer res.Body.Close()
+	}
+
+	if res.TLS == nil {
+		fmt.Println("WARNING! Communication is not secure, please consider using HTTPS. Letsencrypt.org offers free SSL/TLS certificates.")
+	}
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		return nil
+	case http.StatusUnauthorized:
+		return errors.New("unable to login, either username or password is incorrect")
+	default:
+		bytesOut, err := ioutil.ReadAll(res.Body)
+		if err == nil {
+			return fmt.Errorf("server returned unexpected status code: %d - %s", res.StatusCode, string(bytesOut))
+		}
+	}
+
+	return nil
+}
--- a/config/file.go
+++ b/config/file.go
@@ -1,0 +1,282 @@
+// Copyright (c) OpenFaaS Project 2017. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package config
+
+import (
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/url"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/mitchellh/go-homedir"
+	"gopkg.in/yaml.v2"
+)
+
+var (
+	DefaultDir = "~/.openfaas"
+	// using .yml to avoid collision with openfaas-bitbar config
+	DefaultFile = "config.yml"
+)
+
+type ConfigFile struct {
+	AuthConfigs []AuthConfig `yaml:"auths"`
+	FilePath    string       `yaml:"-"`
+}
+
+type AuthConfig struct {
+	Gateway string `yaml:"gateway,omitempty"`
+	Auth    string `yaml:"auth,omitempty"`
+	Token   string `yaml:"token,omitempty"`
+}
+
+// New initializes a config file for the given file path
+func New(filePath string) (*ConfigFile, error) {
+	if filePath == "" {
+		return nil, errors.New("can't create config with empty filePath")
+	}
+	conf := &ConfigFile{
+		AuthConfigs: make([]AuthConfig, 0),
+		FilePath:    filePath,
+	}
+
+	return conf, nil
+}
+
+// EnsureFile creates the root dir and config file
+func EnsureFile() (string, error) {
+	dirPath, err := homedir.Expand(DefaultDir)
+	if err != nil {
+		return "", err
+	}
+
+	filePath := path.Clean(filepath.Join(dirPath, DefaultFile))
+	if err := os.MkdirAll(filepath.Dir(filePath), 0700); err != nil {
+		return "", err
+	}
+
+	if _, err := os.Stat(filePath); os.IsNotExist(err) {
+		file, err := os.OpenFile(filePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
+		if err != nil {
+			return "", err
+		}
+		defer file.Close()
+	}
+
+	return filePath, nil
+}
+
+// FileExists returns true if the config file is located at the default path
+func fileExists() bool {
+	dirPath, err := homedir.Expand(DefaultDir)
+	if err != nil {
+		return false
+	}
+
+	filePath := path.Clean(filepath.Join(dirPath, DefaultFile))
+	if _, err := os.Stat(filePath); os.IsNotExist(err) {
+		return false
+	}
+
+	return true
+}
+
+// Save writes the config to disk
+func (configFile *ConfigFile) save() error {
+	file, err := os.OpenFile(configFile.FilePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	data, err := yaml.Marshal(configFile)
+	if err != nil {
+		return err
+	}
+
+	_, err = file.Write(data)
+	return err
+}
+
+// Load reads the yml file from disk
+func (configFile *ConfigFile) load() error {
+	conf := &ConfigFile{}
+
+	if _, err := os.Stat(configFile.FilePath); os.IsNotExist(err) {
+		return errors.New("can't load config from non existent filePath")
+	}
+
+	data, err := ioutil.ReadFile(configFile.FilePath)
+	if err != nil {
+		return err
+	}
+
+	if err := yaml.Unmarshal(data, conf); err != nil {
+		return err
+	}
+
+	if len(conf.AuthConfigs) > 0 {
+		configFile.AuthConfigs = conf.AuthConfigs
+	}
+	return nil
+}
+
+// EncodeAuth encodes the username and password strings to base64
+func EncodeAuth(username string, password string) string {
+	input := username + ":" + password
+	msg := []byte(input)
+	encoded := make([]byte, base64.StdEncoding.EncodedLen(len(msg)))
+	base64.StdEncoding.Encode(encoded, msg)
+	return string(encoded)
+}
+
+// DecodeAuth decodes the input string from base64 to username and password
+func DecodeAuth(input string) (string, string, error) {
+	decoded, err := base64.StdEncoding.DecodeString(input)
+	if err != nil {
+		return "", "", err
+	}
+	arr := strings.SplitN(string(decoded), ":", 2)
+	if len(arr) != 2 {
+		return "", "", errors.New("invalid auth config file")
+	}
+	return arr[0], arr[1], nil
+}
+
+// UpdateAuthConfig creates or updates the username and password for a given gateway
+func UpdateAuthConfig(gateway string, username string, password string) error {
+	_, err := url.ParseRequestURI(gateway)
+	if err != nil || len(gateway) < 1 {
+		return errors.New("invalid gateway URL")
+	}
+
+	if len(username) < 1 {
+		return errors.New("username can't be an empty string")
+	}
+
+	if len(password) < 1 {
+		return errors.New("password can't be an empty string")
+	}
+
+	configPath, err := EnsureFile()
+	if err != nil {
+		return err
+	}
+
+	cfg, err := New(configPath)
+	if err != nil {
+		return err
+	}
+
+	if err := cfg.load(); err != nil {
+		return err
+	}
+
+	auth := AuthConfig{
+		Gateway: gateway,
+		Auth:    "basic",
+		Token:   EncodeAuth(username, password),
+	}
+
+	index := -1
+	for i, v := range cfg.AuthConfigs {
+		if gateway == v.Gateway {
+			index = i
+			break
+		}
+	}
+
+	if index == -1 {
+		cfg.AuthConfigs = append(cfg.AuthConfigs, auth)
+	} else {
+		cfg.AuthConfigs[index] = auth
+	}
+
+	if err := cfg.save(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// LookupAuthConfig returns the username and password for a given gateway
+func LookupAuthConfig(gateway string) (string, string, error) {
+	if !fileExists() {
+		return "", "", errors.New("config file not found")
+	}
+
+	configPath, err := EnsureFile()
+	if err != nil {
+		return "", "", err
+	}
+
+	cfg, err := New(configPath)
+	if err != nil {
+		return "", "", err
+	}
+
+	if err := cfg.load(); err != nil {
+		return "", "", err
+	}
+
+	for _, v := range cfg.AuthConfigs {
+		if gateway == v.Gateway {
+			user, pass, err := DecodeAuth(v.Token)
+			if err != nil {
+				return "", "", err
+			}
+			return user, pass, nil
+		}
+	}
+
+	return "", "", fmt.Errorf("no auth config found for %s", gateway)
+}
+
+// RemoveAuthConfig deletes the username and password for a given gateway
+func RemoveAuthConfig(gateway string) error {
+	if !fileExists() {
+		return errors.New("config file not found")
+	}
+
+	configPath, err := EnsureFile()
+	if err != nil {
+		return err
+	}
+
+	cfg, err := New(configPath)
+	if err != nil {
+		return err
+	}
+
+	if err := cfg.load(); err != nil {
+		return err
+	}
+
+	index := -1
+	for i, v := range cfg.AuthConfigs {
+		if gateway == v.Gateway {
+			index = i
+			break
+		}
+	}
+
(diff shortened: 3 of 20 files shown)


alexellis:
When we get closer to completing review please can you squash commits?

README.md
```diff
@@ -40,6 +40,8 @@ The main commands supported by the CLI are:
 * `faas-cli deploy` - deploys the functions into a local or remote OpenFaaS gateway
 * `faas-cli remove` - removes the functions from a local or remote OpenFaaS gateway
 * `faas-cli invoke` - invokes the functions and reads from STDIN for the body of the request
+* `faas-cli login` - stores basic auth credentials for OpenFaaS gateways (multi-tenancy support)
```
alexellis: > multi-tenancy support

This isn't the "multi-tenancy support" people on the project are going to expect. Maybe - supports multiple gateways.

config/file.go
```diff
@@ -0,0 +1,217 @@
...
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/mitchellh/go-homedir"
+	"gopkg.in/yaml.v2"
+)
+
+var (
```
alexellis: Out of curiosity.. is this all original work, or based upon an example?

proxy/auth_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"strings"
+
+	"github.com/openfaas/faas-cli/config"
+)
+
+func Test_AuthorizationHeader(t *testing.T) {
+	//setup store
+	config.DefaultFile = "authtest1.yml"
+	gatewayURL := strings.TrimRight("http://openfaas.test/", "/")
+	config.UpdateAuthConfig(gatewayURL, "Aladdin", "open sesame")
```
alexellis: Have we tested for empty string and edge-cases?

vendor.conf
```diff
@@ -4,4 +4,5 @@ github.com/spf13/cobra 2df9a531813370438a4d79bfc33e21f58063ed87
 github.com/spf13/pflag e57e3eeb33f795204c1ca35f56c44f83227c6e6
 github.com/ryanuber/go-glob 256dc444b735e061061cf46c809487313d5b0065
 github.com/morikuni/aec 39771216ff4c63d11f5e604076f9c45e8be1067b
-github.com/inconshreveable/mousetrap 76626ae9c91c4f2a10f34cad8ce83ea42c93bb75
\ No newline at end of file
+github.com/inconshreveable/mousetrap 76626ae9c91c4f2a10f34cad8ce83ea42c93bb75
+github.com/mitchellh/go-homedir b8bc1bf767474819792c23f32d8286a45736f1c6
```
alexellis: Is this new dep Windows compatible, too?

config/file.go
stefanprodan: The naming convention follows docker/cli/config and the decode function is using the same escape as in docker registry store, other than that is as "original" as this little pice of code can be :)

vendor.conf
stefanprodan: Yes it is see https://github.com/mitchellh/go-homedir/blob/master/homedir.go#L120

proxy/auth_test.go
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Stefan Prodan 2017. All rights reserved.
```
alexellis: Is this copied from Lewis PR or visa-versa?

https://github.com/openfaas/faas-cli/pull/180

If so is the Copyright notice appropriate?

proxy/auth_test.go
stefanprodan: The URL empty string validation is done here https://github.com/openfaas/faas-cli/pull/182/files/dc06babac728c33fb128bba45766f229c134a991#diff-4d2d07a462d7a1dda6e42b8c443d6d77R74 and the validity of the URL is tested by calling the gateway `/list` before the URL get's saved in the config.

proxy/auth_test.go
stefanprodan: The hashing comparison it is copied. 

Regarding Copyright wouldn't be better to put "OpenFaaS Team" and not the contributor name? I've seen this approach on other projects. Each commit is signed, so if someone wants to see the actual contributors it can from the commit history.

proxy/auth_test.go
alexellis: My question was - have we got a unit test?

proxy/auth_test.go
alexellis: If you like you can put OpenFaaS authors on your new commits in this PR. Unlikely to retrofit across all repos at this point.

proxy/auth_test.go
Lewiscowles1986: I'm also unsure we should be using username password. My original PR only did it because it incorrectly used cli-arguments. We are storing base64, why not add the requisite header?

stefanprodan: testes added for empty string and invalid URLs in https://github.com/openfaas/faas-cli/pull/182/commits/35df33c431d82e5137dfd0928c48940d185d2170

stefanprodan:
@alexellis you can squash all commits at merge, see https://github.com/blog/2141-squash-your-commits

vendor/github.com/mitchellh/go-homedir/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
nenadilic84: @alexellis, @stefanprodan, I don't like the fact that we checkin the code in vendor. I think I mentioned this before, but it should be that only use the vendor.conf to have the right versions, and then let the tool checkout everything before building. I would just put .gitignore in this folder.

alexellis: This is how dependencies are managed on Golang projects. It may seem odd, but you can find more information by looking at Docker or Google's Go projects.

alexellis:
@stefanprodan I won't be altering the commits at merge time because it'll lose the sign off. If you could do that when the work is ready that would work better.

stefanprodan:
Squash done and ready for merge :)

alexellis:
Thanks for squashing. I'll try to do a detailed code-review over the weekend. This is great timing.

I've done some testing on my local machine with basic auth on nginx and so far looks very good :+1:

config/file.go
alexellis: Can you link me to that code please? Thank you

alexellis:
@johnmccabe when you're up to it, please can you review the config changes?

proxy/invoke.go
```diff
@@ -26,13 +27,13 @@ func InvokeFunction(gateway string, name string, bytesIn *[]byte, contentType st
...
 	if qsErr != nil {
 		return nil, qsErr
-
 	}
 
 	gatewayURL := gateway + "/function/" + name + qs
 	// fmt.Println(gatewayURL)
 	req, _ := http.NewRequest(http.MethodPost, gatewayURL, reader)
 	req.Header.Add("Content-Type", contentType)
+	SetAuth(req, gateway)
```
johnmccabe: This is good for now.

But I think we will want to be more nuanced with locking down endpoints, with functions and system being secured using different credentials (and most likely with individual functions with their own auth (or none)). (^ @alexellis)

proxy/auth_test.go
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) OpenFaaS authors 2017. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package proxy
+
+import (
+	"net/http"
+	"testing"
+
```
johnmccabe: Delete this line

config/file.go
```diff
@@ -0,0 +1,233 @@
...
+	conf := &ConfigFile{
+		AuthConfigs: make(map[string]AuthConfig),
+		FilePath:    filePath,
+	}
+
+	return conf, nil
+}
+
+// EnsureFile creates the root dir and config file
+func EnsureFile() (string, error) {
```
johnmccabe: Would be cleaner when testing to allow the homedir to be overridden, currently you get your local `~/.openfaas/` dir polluted with test YAML.

config/file_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"os"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+func Test_LookupAuthConfig_WithNoConfigFile(t *testing.T) {
+	DefaultFile = "test1.yml"
+	_, _, err := LookupAuthConfig("http://openfaas.test1")
+
```
johnmccabe: Can you get rid of the extra line between the function calls and error handling, same for the rest of this file.

proxy/invoke.go
alexellis: Sending basic authentication when the function is open will result in the header being ignored, so we should be ok.

commands/login.go
```diff
@@ -0,0 +1,139 @@
...
+
+func validateLogin(url string, user string, pass string) error {
+	// TODO: provide --insecure flag for this
+	tr := &http.Transport{
+		DisableKeepAlives: true,
+		TLSClientConfig:   &tls.Config{InsecureSkipVerify: true},
+	}
+	client := &http.Client{
+		Transport: tr,
+		Timeout:   time.Duration(5 * time.Second),
```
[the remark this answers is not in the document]
alexellis: Agreed

stefanprodan:
@johnmccabe @alexellis I've addressed all your change requests. After this gets merged I will do the `logout` command.

README.md
```diff
@@ -49,9 +52,9 @@ You can chose between using a [programming language template](https://github.com
 
 **Templates**
 
-Command: `faas-cli new FUNCTION_NAME --lang python/node/ruby/Dockerfile/etc`
+Command: `faas-cli new FUNCTION_NAME --lang python/node/go//ruby/Dockerfile/etc`
```
ericstoekl: Are there supposed to be two `/` characters after `go`?

johnmccabe: I think you're looking at an old revision there @ericstoekl ?

johnmccabe:
Just for future reference, it would be much easier to not squash after committing review updates as I have to review the entire PR again now :(

commands/login.go
```diff
@@ -0,0 +1,139 @@
+// Copyright (c) Stefan Prodan 2017. All rights reserved.
```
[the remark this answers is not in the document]
alexellis: Yes needs to be assigned to the project for new files - OpenFaaS authors.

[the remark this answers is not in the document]
johnmccabe: @stefanprodan can you update this per - https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md#copyright-notice

config/file.go
```diff
@@ -0,0 +1,222 @@
+// Copyright (c) Stefan Prodan 2017. All rights reserved.
```
johnmccabe: Can you update this per - https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md#copyright-notice

config/file_test.go
```diff
@@ -0,0 +1,157 @@
+// Copyright (c) Stefan Prodan 2017. All rights reserved.
```
johnmccabe: Can you update this per - https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md#copyright-notice

proxy/auth.go
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) OpenFaaS authors 2017. All rights reserved.
```
johnmccabe: Can you update this per - https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md#copyright-notice

proxy/auth_test.go
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) OpenFaaS authors 2017. All rights reserved.
```
johnmccabe: Can you update this per - https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md#copyright-notice

config/file_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+func Test_New_NoFile(t *testing.T) {
+	_, err := New("")
+	if err == nil {
+		t.Error("expected to fail on empty file path")
+	}
+}
+
+func Test_EnsureFile(t *testing.T) {
+	DefaultDir = "~/.openfaas_test"
```
johnmccabe: Can you use `ioutil.TempDir("", "faas-cli-file-test")` here so we don't pollute the users directory.

config/file.go
johnmccabe: @alexellis do you still need this?

johnmccabe:
Its a 👍 from me, just need to get these two bits sorted and I think its it looks good to merge:
- typo - https://github.com/openfaas/faas-cli/pull/182#discussion_r147745082
- question from @alexellis - https://github.com/openfaas/faas-cli/pull/182#discussion_r147549275

config/file.go
stefanprodan: Sorry I've missed this question. Here it is https://github.com/docker/cli/blob/master/cli/config/configfile/file.go

stefanprodan:
@johnmccabe that typo got into my PR from a master merge. I will fix it.

commands/login.go
```diff
@@ -0,0 +1,140 @@
...
+	username      string
+	password      string
+	passwordStdin bool
+)
+
+func init() {
+	loginCmd.Flags().StringVar(&gateway, "gateway", defaultGateway, "Gateway URI")
+	loginCmd.Flags().StringVarP(&username, "username", "u", "", "Gateway username")
+	loginCmd.Flags().StringVarP(&password, "password", "p", "", "Gateway password")
+	loginCmd.Flags().BoolVar(&passwordStdin, "password-stdin", false, "Reads the gateway password from stdin")
```
alexellis: We can tweak the UX after merging but password-stdin should be as easy to use as -p so that people don't just default to having their password in the bash history.

commands/login.go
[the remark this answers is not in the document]
alexellis: @stefanprodan ..  something like?

```
Validating login... OK.
Validating login... Failed.
```

config/file.go
alexellis: Yes - talked about this with @johnmccabe on Sunday. We should be using os.Tempdir or similar for integration tests.

As a side note about the cli tests, integration tests should perhaps be in their own area so we can run them independently.

stefanprodan: I've used `os.Tempdir` take a look at the latest commit.

config/file_test.go
```diff
@@ -0,0 +1,158 @@
...
+	DefaultDir, _ = ioutil.TempDir("", "faas-cli-file-test")
+	DefaultFile = "test1.yml"
+	_, _, err := LookupAuthConfig("http://openfaas.test1")
+	if err == nil {
+		t.Errorf("Error was not returned")
+	}
+
+	r := regexp.MustCompile(`(?m:config file not found)`)
+	if !r.MatchString(err.Error()) {
+		t.Errorf("Error not matched: %s", err.Error())
```
alexellis: I don't like the pattern we have in these tests (it's not from Stefan, but from some historic tests that got copied) `Error not matched` is super vague.. can we do better? What would you suggest guys? @stefanprodan @johnmccabe 

Imagine reading this deep in an CI log.

config/file_test.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		t.Errorf("got error %s", err.Error())
+	}
+
+	if user != u || pass != p {
+		t.Errorf("got user %s and pass %s, expected %s %s", user, pass, u, p)
+	}
+}
+
+func Test_UpdateAuthConfig_Update(t *testing.T) {
```
alexellis: Thank you for taking the time to add the test coverage. This means a lot @stefanprodan

proxy/auth.go
```diff
@@ -0,0 +1,21 @@
...
+package proxy
+
+import (
+	"net/http"
+
+	"github.com/openfaas/faas-cli/config"
+)
+
+//SetAuth sets basic auth for the given gateway
+func SetAuth(req *http.Request, gateway string) {
```
alexellis: This function feels orphaned to me. I know why it's here because it gives a sense of removing duplication, but I think we could do this in a cleaner way. I saw a similar pattern for user-agent and didn't like it either.

I'm thinking of our own request factory which adds the things we know we need.

Doesn't have to be part of this PR. @stefanprodan  @rgee0 - thoughts?

commands/login.go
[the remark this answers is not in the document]
stefanprodan: I've done this, see line 78

proxy/auth_test.go
```diff
@@ -0,0 +1,45 @@
...
+		t.Errorf("got header %q, want %q", header, expected)
+	}
+}
+
+func Test_SetAuth_SkipAuthorization(t *testing.T) {
+	//setup store
+	config.DefaultDir, _ = ioutil.TempDir("", "faas-cli-auth-test")
+	config.DefaultFile = "authtest2.yml"
+	gatewayURL := strings.TrimRight("http://openfaas.test/", "/")
+	config.UpdateAuthConfig(gatewayURL, "Aladdin", "open sesame")
```
alexellis: I like this test a lot but I had to read it 3-4 times. I think it could be easier to read if we broke out the URLs into variables so it's easier to read and understand at a glance.

```
basicAuthURL
openURL
```

proxy/deploy.go
```diff
@@ -89,6 +90,8 @@ func DeployFunction(fprocess string, gateway string, functionName string, image
 
 		deployedURL := fmt.Sprintf("URL: %s/function/%s\n", gateway, functionName)
 		fmt.Println(deployedURL)
+	case 401:
```
alexellis: Please can you update to http.StatusUnauthorized etc?

proxy/delete.go
```diff
@@ -39,6 +40,8 @@ func DeleteFunction(gateway string, functionName string) {
 		fmt.Println("Removing old service.")
 	case 404:
 		fmt.Println("No existing service to remove")
+	case 401:
```
alexellis: Please can you update to http.StatusUnauthorized etc? I know these have the numbers in (that's on me..) but let's fix while we're in there.

proxy/list.go
```diff
@@ -46,7 +49,8 @@ func ListFunctions(gateway string) ([]requests.Function, error) {
 		if jsonErr != nil {
 			return nil, fmt.Errorf("cannot parse result from OpenFaaS on URL: %s\n%s", gateway, jsonErr.Error())
 		}
-
+	case 401:
```
alexellis: Please go over to http.StatusEtc...

README.md
```diff
@@ -40,6 +40,8 @@ The main commands supported by the CLI are:
 * `faas-cli deploy` - deploys the functions into a local or remote OpenFaaS gateway
 * `faas-cli remove` - removes the functions from a local or remote OpenFaaS gateway
 * `faas-cli invoke` - invokes the functions and reads from STDIN for the body of the request
+* `faas-cli login` - stores basic auth credentials for OpenFaaS gateway (supports multiple gateways)
```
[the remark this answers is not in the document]
stefanprodan: I've implemented `logout` and added tests for store remove func

alexellis:
When we spoke on Slack we settled on `logout` being out of scope for this PR. I'll have to set some more time aside to review.

stefanprodan:
The `logout` code is straight forward, 20 lines of code with 100% test coverage :)

commands/login.go
```diff
@@ -0,0 +1,140 @@
...
+		if len(username) == 0 {
+			return errors.New("must provide --username with --password-stdin")
+		}
+
+		passwordStdin, err := ioutil.ReadAll(os.Stdin)
+		if err != nil {
+			return err
+		}
+
+		password = strings.TrimSpace(string(passwordStdin))
```
alexellis: Do we also need to trim the --password flag data?

alexellis:
Started to do some testing on this. I found the output format of ~/.openfaas/config.yml looked a little odd:

```
auths:
  http://127.0.0.1:8081:
    auth: basic
    token: value
  http://other-site:8080:
    auth: basic
    token: value
  http://other-gateway:8081:
    auth: basic
    token: value
```

There was a comment Joe Beda made last Friday about using values for keys which seemed to make sense at the time. We have broken the rule for function names in the OpenFaaS stack YAML but it looks worse here with the URL and then the final colon on the line.

Do you have any other suggestions on improving the key for the URL?

stefanprodan:
@alexellis Regarding the config format, I am using the same format as Docker CLI uses for storing the registries auth. I don't see a problem with it, this is how a golang map is getting serialized in yaml.

proxy/auth.go
alexellis: Would like your thoughts.

proxy/delete.go
```diff
@@ -40,10 +41,12 @@ func DeleteFunction(gateway string, functionName string) error {
 	}
 
 	switch delRes.StatusCode {
-	case 200, 201, 202:
+	case http.StatusOK, http.StatusCreated, http.StatusAccepted:
```
alexellis: Thanks for making this better.

alexellis:
Have you tested thoroughly on Windows?

alexellis:
I think this format is closer to the "Kubernetes way of YAML" - https://github.com/estesp/bucketbench

```
name: Basic
image: alpine:latest
command: date
rootfs: /home/estesp/containers/alpine
detached: true
drivers:
  - 
   type: Docker
   threads: 5
   iterations: 15
  - 
   type: Runc
   threads: 5
   iterations: 50
```

I don't know how easy it is to serialize that way.

Another example (looking at the "containers")

```
apiVersion: apps/v1beta2 # for versions before 1.7.0 use apps/v1beta1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  replicas: 3
  template:
    metadata:
      labels:
        app: nginx
    spec:
  - name: nginx-container
    image: nginx

  - name: debian-container
    image: debian
    command: ["/bin/sh"]
```

Both appear to be applying an array of "map/object" in order to force this syntax.. I mean you can checkout the video from Joe. I think we should at least evaluate his advice. Around here - https://www.youtube.com/watch?v=iMzf_oWsRi0

commands/login.go
```diff
@@ -0,0 +1,140 @@
...
+	if res.Body != nil {
+		defer res.Body.Close()
+	}
+
+	if res.TLS == nil {
+		fmt.Println("WARNING! Communication is not secure, please consider using HTTPS. Letsencrypt.org offers free SSL/TLS certificates.")
+	}
+
+	switch res.StatusCode {
+	case 200:
```
alexellis: More numbers, but we can sort these later

alexellis:
Minor nit between `docker login` and `faas-cli login`:

```
docker login -u alex
(starts reading password from stdin)
```

```
faas-cli login -u alex
Error: password cannot be an empty string
```

Might be worth looking at how we do the `invoke` command and doing the same. Could be a follow-up PR / issue.

stefanprodan:
I've updated the description, the CLI unit tests and login/logout manual testing were performed on:

* macOS Siera 10.12.6
* Ubuntu 16.04
* Windows 10 Pro (go 1.9.2 windows/amd64)

stefanprodan:
@alexellis regarding the config, if you look at bucketbench you'll see that the drivers are an array not a map. You can't force a map to be serialized as an array, so you'll not get the `-`.

rgee0:
My sense is that the yaml format should be formed of nested maps so that each value has a name. This enhances the readability and use ability for the user.

It’s actually not too far off, probably just needs the gateway uri to identify itself as a gateway value.

stefanprodan:
A nested map is not an array, if @alexellis wants the config to look like the k8s one I need to change it to an array.

```yml
auths:
 - gateway: http://127.0.0.1:8081
    auth: basic
    token: value
 - gateway: http://other-site:8080
    auth: basic
    token: value
```

@alexellis should I rewrite the config package to have the above format?

config/file_test.go
johnmccabe: I tend to use table drive as much as possible with the name of the testing being printed and for assertion failures printing the both the expected and actual values, kinda like:

    t.Errorf("Returned error [%v] does not match expected value [%v]", err, expectedErr)

@stefanprodan as an aside, you should be using `%v` with the raw error rather than `%s` and calling the `.String()` function.

johnmccabe:
Regarding the file structure:
- Is the config file intended to be human readable/editable?
  - if it is, then yes to named keys
- if not then using a map is easier as you don’t need to iterate through the slice to find a particular gateway

stefanprodan:
I don't see users editing the auth data in the config file, but if the config will grow in time I think we should make it as readable as possible. Changing the code from map to array is an easy change with little overhead. @alexellis what's your take on this?

commands/login.go
stefanprodan: done in the latest commit

stefanprodan:
I've switched from map to array, this how it looks:

```yml
auths:
- gateway: http://35.198.149.1
  auth: basic
  token: value
- gateway: http://35.198.149.2
  auth: basic
  token: value
```

developius:
This is really cool. Clearly this PR has been merged but I had a comment about using `--password-stdin` though - all the CLIs I've ever used use the format `-p -` (or `--password -`) to mean "grab this from stdin". 
Example:

```
$ cat ~/faas_pass.txt | faas-cli login -u user -p - --gateway http://localhost:8080
```
Thoughts?

alexellis:
What does the Docker CLI do?

developius:
It uses `--password-stdin`, but that's the only occurrence I've ever seen of not using `-p -`

stefanprodan:
@developius how will you do that with cobra?

developius:
@stefanprodan I think you could just check to see if the password flag was set to `-`, and if it was do the same logic as if `passwordStdin` was `true` [login.go#L61-L72](https://github.com/stefanprodan/faas-cli/blob/ee859a3e699160a17d23ac05f33adcb41e1e8df9/commands/login.go#L61-L72)

stealthybox:
`-` is usually a program specific alias for `/dev/stdin`.
( I actually quite dislike this because it overrides other shell syntax, but it's fairly standard. )

For `-p` the param isn't a file name.
I believe the proper flag for a `-` alias would be something like `--password-file`.

This would be useful for other things like secrets in a tmpfs.

Issue #183, opened 2017-10-22
Refactoring: Return non-zero exit codes for error conditions


stefanprodan:
## Expected Behaviour

When a command fails I would expect exit code 1 so automation tools can detect an error occurred and stop/react accordingly. 

## Current Behaviour

Some examples of error handling:

```go
	response, err := proxy.InvokeFunction(gateway, functionName, &functionInput, contentType)
	if err != nil {
		fmt.Println(err)
		return
	}
```

```go
	if len(services.Functions) > 0 {
		pushStack(&services, parallel)
	} else {
		fmt.Println("You must supply a valid YAML file.")
		return
	}
```

```
	if len(services.Functions) > 0 {
		pushStack(&services, parallel)
	} else {
		fmt.Println("You must supply a valid YAML file.")
		return
	}
```

## Possible Solution

* use cobra.Command.RunE instead of Run
* change the signature of all runCmd functions to return error
* propagate errors with pkg/errors and wrap the original error with meaningful messages

ericstoekl:
Derek add label: proposal

alexellis:
Agree with this refactoring. Should ideally return an error. This is tech debt.

alexellis:
I think it would make sense to start with one of the commands in isolation, merging that then doing the others one at a time. Could be easier for bite-size chunks / contributions.

nenadilic84:
I'll work on this if anyone asks 😄

alexellis:
Which one do you want to take first?

nenadilic84:
I'll start with `build` if you're ok with it...

nenadilic84:
Added the PR ^
If it seems ok, I can continue and add other commands as well

jonocodes:
Ok if I take a stab at the list command?

alexellis:
Let's get one finished and merged before doing the next. Yes Jono please do the lists command and reach out on Slack there will be people ready to help and to get you up to speed.

alexellis:
I mean one PR per person, so we have less work in progress and more work delivered.

alexellis:
@stefanprodan @johnmccabe can we compile a quick mark-down table showing what's done/left in terms of non-zero exits? Is faas-cli login/logout covered already?

Thanks. @austinfrey @jockdarock maybe we can direct some of the newer contributors here so they can get their first Go PRs into the project?

johnmccabe:
@alexellis this should probably be closed following the merge of #207

rgee0:
Derek close

PR #207, opened 2017-11-06
Refactoring Commands to exit with non-0 status codes on failure.

Razesdark:
Signed-off-by: Tommy Stigen Olsen <tommysolsen@gmail.com>

As discussed in #183, I have gone through the commands and made all of the cobra commands 
use RunE, returning errors which triggers os.Exit(1).

## Description

Refactors all cobra commands except version (which never generates errors). This lets all of the commands exit with status-code 1 if an error occurs.
I have also cleaned up some of the different ways each function handles failure to a standard `fmt.Println` and `return err` style to be consistent across the modules.

Since RunE will print any error it receives, i've suppressed them so the software doesnt repeat itself.

## Motivation and Context
See #183 

## How Has This Been Tested?
Tested through compiling and running a few functions that failed.

## Types of changes

## Checklist:

### DIFF
--- a/commands/bash_completion.go
+++ b/commands/bash_completion.go
@@ -4,6 +4,7 @@
 package commands
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/spf13/cobra"
@@ -23,18 +24,20 @@ var bashcompletionCmd = &cobra.Command{
 This currently only works on Bash version 4, and is hidden
 pending a merge of https://github.com/spf13/cobra/pull/520.`,
 	Hidden: true,
-	Run:    runBashcompletion,
+	RunE:   runBashcompletion,
 }
 
-func runBashcompletion(cmd *cobra.Command, args []string) {
+func runBashcompletion(cmd *cobra.Command, args []string) error {
 	if len(args) < 1 {
 		fmt.Println("Please provide filename for bash completion")
-		return
+		return errors.New("Please provide filename for bash completion")
 	}
 	fileName := args[0]
 	err := faasCmd.GenBashCompletionFile(fileName)
 	if err != nil {
 		fmt.Println("Unable to create bash completion file")
-		return
+		return errors.New("unable to create bash completion file")
 	}
+
+	return nil
 }
--- a/commands/bash_completion.go
+++ b/commands/bash_completion.go
@@ -4,7 +4,6 @@
 package commands
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/spf13/cobra"
@@ -29,14 +28,12 @@ pending a merge of https://github.com/spf13/cobra/pull/520.`,
 
 func runBashcompletion(cmd *cobra.Command, args []string) error {
 	if len(args) < 1 {
-		fmt.Println("Please provide filename for bash completion")
-		return errors.New("Please provide filename for bash completion")
+		return fmt.Errorf("please provide filename for bash completion")
 	}
 	fileName := args[0]
 	err := faasCmd.GenBashCompletionFile(fileName)
 	if err != nil {
-		fmt.Println("Unable to create bash completion file")
-		return errors.New("unable to create bash completion file")
+		return fmt.Errorf("unable to create bash completion file")
 	}
 
 	return nil
--- a/commands/build.go
+++ b/commands/build.go
@@ -73,6 +73,7 @@ func runBuild(cmd *cobra.Command, args []string) error {
 	if len(yamlFile) > 0 {
 		parsedServices, err := stack.ParseYAMLFile(yamlFile, regex, filter)
 		if err != nil {
+			fmt.Println(err)
 			return err
 		}
 
@@ -89,12 +90,15 @@ func runBuild(cmd *cobra.Command, args []string) error {
 		build(&services, parallel, shrinkwrap)
 	} else {
 		if len(image) == 0 {
+			fmt.Println("Please provice a valid -image name for yout Docker image")
 			return fmt.Errorf("please provide a valid -image name for your Docker image")
 		}
 		if len(handler) == 0 {
+			fmt.Println("Please provide the full path to your function's handler")
 			return fmt.Errorf("please provide the full path to your function's handler")
 		}
 		if len(functionName) == 0 {
+			fmt.Println("Please provide the deployed -name of your function")
 			return fmt.Errorf("please provide the deployed -name of your function")
 		}
 		builder.BuildImage(image, handler, functionName, language, nocache, squash, shrinkwrap)
--- a/commands/build.go
+++ b/commands/build.go
@@ -73,7 +73,6 @@ func runBuild(cmd *cobra.Command, args []string) error {
 	if len(yamlFile) > 0 {
 		parsedServices, err := stack.ParseYAMLFile(yamlFile, regex, filter)
 		if err != nil {
-			fmt.Println(err)
 			return err
 		}
 
@@ -90,16 +89,13 @@ func runBuild(cmd *cobra.Command, args []string) error {
 		build(&services, parallel, shrinkwrap)
 	} else {
 		if len(image) == 0 {
-			fmt.Println("Please provice a valid -image name for yout Docker image")
-			return fmt.Errorf("please provide a valid -image name for your Docker image")
+			return fmt.Errorf("please provide a valid --image name for your Docker image")
 		}
 		if len(handler) == 0 {
-			fmt.Println("Please provide the full path to your function's handler")
 			return fmt.Errorf("please provide the full path to your function's handler")
 		}
 		if len(functionName) == 0 {
-			fmt.Println("Please provide the deployed -name of your function")
-			return fmt.Errorf("please provide the deployed -name of your function")
+			return fmt.Errorf("please provide the deployed --name of your function")
 		}
 		builder.BuildImage(image, handler, functionName, language, nocache, squash, shrinkwrap)
 	}
@@ -119,7 +115,7 @@ func build(services *stack.Services, queueDepth int, shrinkwrap bool) {
 			for function := range workChannel {
 				fmt.Printf("[%d] > Building: %s.\n", index, function.Name)
 				if len(function.Language) == 0 {
-					fmt.Println("Please provide a valid -lang or 'Dockerfile' for your function.")
+					fmt.Println("Please provide a valid --lang or 'Dockerfile' for your function.")
 
 				} else {
 					builder.BuildImage(function.Image, function.Handler, function.Name, function.Language, nocache, squash, shrinkwrap)
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -4,10 +4,8 @@
 package commands
 
 import (
-	"errors"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"strings"
 
 	"gopkg.in/yaml.v2"
@@ -98,14 +96,13 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 		fmt.Println(`Cannot specify --update and --replace at the same time.
   --replace    removes an existing deployment before re-creating it
   --update     provides a rolling update to a new function image or configuration`)
-		return errors.New("cannot specify --update and --replace at the same time")
+		return fmt.Errorf("cannot specify --update and --replace at the same time")
 	}
 
 	var services stack.Services
 	if len(yamlFile) > 0 {
 		parsedServices, err := stack.ParseYAMLFile(yamlFile, regex, filter)
 		if err != nil {
-			log.Println(err.Error())
 			return err
 		}
 
@@ -144,7 +141,6 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 
 			fileEnvironment, err := readFiles(function.EnvironmentFile)
 			if err != nil {
-				log.Println(err)
 				return err
 			}
 
@@ -155,15 +151,13 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 
 			labelArgumentMap, labelErr := parseMap(labelOpts, "label")
 			if labelErr != nil {
-				fmt.Printf("Error parsing labels: %v\n", labelErr)
-				return errors.New("error parsing labels")
+				return fmt.Errorf("error parsing labels: %v", labelErr)
 			}
 
 			allLabels := mergeMap(labelMap, labelArgumentMap)
 
 			allEnvironment, envErr := compileEnvironment(envvarOpts, function.Environment, fileEnvironment)
 			if envErr != nil {
-				log.Println(envErr)
 				return envErr
 			}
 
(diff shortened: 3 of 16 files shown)


Razesdark:
I've pushed some new code. I've started going through the commands one more time to capture the output and found a bug that slipped me last time. If a user would try to deploy with a malformed label not including a =, you would get a index out of range panic. Added a catch for that error.

I'll be a little more thorough and retest everything to make sure everything is as its supposed to. Will post the results when I'm done.

Razesdark:
## build
Built one function with no parameters
```
$ ./faas-cli build -f samples.yml
[0] > Building: url-ping.
Please provide a valid -lang or 'Dockerfile' for your function.
[0] < Builder done.

```
Built one function with lang parameter, but without handler or image
```
./faas-cli build -f samples.yml
[0] > Building: url-ping.
Clearing temporary build folder: ./build/url-ping/
Preparing / ./build/url-ping/function
2017/11/07 13:05:52 open : no such file or directory
```
built one with lang and handler parameters, not image
```
./faas-cli build -f samples.yml
[0] > Building: url-ping.
Clearing temporary build folder: ./build/url-ping/
Preparing ./sample/url-ping/ ./build/url-ping/function
Building:  with python template. Please wait..
invalid argument "" for t: invalid reference format
See 'docker build --help'.
```
I assume these two last ones should have failed with proper error messages,
but it seems that `stack.ParseYAMLFile` returns an instance for a service, even if both handler and image are missing.

## deploy
```
$ ./faas-cli deploy
Please provide a --image to be deployed.
```
Here is tested on a file with some giberish in it.

Other functions that parse yaml(like `list`) giberish will produce same output
```
$ ./faas-cli deploy -f test.yml
Error with YAML file
2017/11/07 12:50:19 yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `sefojrh...` into stack.Services

```

With both update and replace
```
$ ./faas-cli deploy --update --replace
Cannot specify --update and --replace at the same time.
  --replace    removes an existing deployment before re-creating it
  --update     provides a rolling update to a new function image or configuration
```

```
$ ./faas-cli deploy               
Please provide a --image to be deployed.
```

```
$ ./faas-cli deploy --image v2m    
Please provide a --name for your function as it will be deployed on FaaS
```

All label related error messages.
Including one bugfix that would panic the program
```
$ ./faas-cli deploy --image v2m --name k --label somethingiswr 
Error parsing labels: Label format is not correct, needs key=value

$ ./faas-cli deploy --image v2m --name k --label somethingiswr=
Error parsing labels: Empty label value: [somethingiswr=]

$ ./faas-cli deploy --image v2m --name k --label " =test"      
Error parsing labels: Empty label name: [ =test]

```

## invoke
No name supplied 
```
$ ./faas-cli invoke     
Please provide a name for the function
```


## list
Gateway is unreachable, triggering error from proxy.GetFunctions
```
$ ./faas-cli list --gateway http://127.0.0.1:8080

Get http://127.0.0.1:8080/system/functions: dial tcp 127.0.0.1:8080: getsockopt: connection refused
```

## login

```
$ ./faas-cli login 
Must provide --username or -u

$ ./faas-cli login --username test
Must provide a non-empty password via --password or --password-stdin
```

## logout
Trying to logout without any valid logins registered(No config file exists)
```
$ ./faas-cli logout
config file not found
```

## new
```
$ ./faas-cli new   
Please provide a name for the function

$ ./faas-cli new test  
You must supply a function language with the --lang flag

$ faas-cli new --list
Languages available as templates:
- node
- python
- python3
- ruby
- csharp
- Dockerfile
- go

Or alternatively create a folder containing a Dockerfile, then pick
the "Dockerfile" lang type in your YAML file.

$ echo $?
0
```

`validTemplate()` would return true, even for languages that did not have
templates. With that fixed

```
$ ./faas-cli new test --lang COBOL
COBOL is unavailable or not supported.
```

Trying to create a function that already exists
```
$ ./faas-cli new test --lang go   
Folder: test already exists
```

## push
Changed samples to not have valid image keys. 

Currently the go routines does not return anything
```
$ ./faas-cli push -f samples.yml
[0] > Pushing: url-ping.
Please provide a valid Image value in the YAML file.
[0] > Pushing: nodejs-echo.
Please provide a valid Image value in the YAML file.
[0] > Pushing: shrink-image.
Please provide a valid Image value in the YAML file.
[0] > Pushing: ruby-echo.
Please provide a valid Image value in the YAML file.
[0] < Pushing done.

$ echo $?
0
```

## remove
```
$ ./faas-cli remove --gateway http://localhost:8080             
Please provide the name of a function to delete
```
## version
Doesn't give any errors

## Missing tests
* invoke - Read from stdio. I don't know how to trigger this error.
* bash completion
* login - My stack doesnt run the login part yet so all messages that depend on server input responses I'll be unable to test until later this week.
* new_function Im not sure how i can trigger the gitignore error or writing to file errors. Even when creating the folder/files without read/write permissions they get created just fine.

ericstoekl:
Compiled this PR and did some cursory checks, looks good so far. The only suggestion I have is that we could use some tests for this particular functionality (test error code result when a command fails, make sure it's `0` when a command succeeds). @alexellis , do you think tests are necessary or not for this PR?

commands/bash_completion.go
```diff
@@ -23,18 +24,20 @@ var bashcompletionCmd = &cobra.Command{
...
 pending a merge of https://github.com/spf13/cobra/pull/520.`,
 	Hidden: true,
-	Run:    runBashcompletion,
+	RunE:   runBashcompletion,
 }
 
-func runBashcompletion(cmd *cobra.Command, args []string) {
+func runBashcompletion(cmd *cobra.Command, args []string) error {
 	if len(args) < 1 {
 		fmt.Println("Please provide filename for bash completion")
```
johnmccabe: This is a general comment that apply to **ALL** places where an error is returned.

The redundancy in error messaging here is not necessary with Cobra, you should *just* return the error message, and *not* print it.

By using the `RunE` field you are passing the error back to the point where the subcommand was invoked (see the comment on `faas.go`. You should be printing the error message at that point instead of in the subcommand.

commands/bash_completion.go
```diff
@@ -23,18 +24,20 @@ var bashcompletionCmd = &cobra.Command{
...
-		return
+		return errors.New("Please provide filename for bash completion")
```
johnmccabe: You can use `fmt.Errorf("error text..")` instead of pulling in the `error` import. Again a general comment that applies across the PR.

commands/build.go
```diff
@@ -89,12 +90,15 @@ func runBuild(cmd *cobra.Command, args []string) error {
 		build(&services, parallel, shrinkwrap)
 	} else {
 		if len(image) == 0 {
+			fmt.Println("Please provice a valid -image name for yout Docker image")
```
johnmccabe: `provide` is spelt incorrectly.
`your` is spelt incorrectly.

Also should be using the posix style flag here in the message `--image` not the go style.

commands/deploy.go
```diff
@@ -227,6 +231,9 @@ func parseMap(envvars []string, keyName string) (map[string]string, error) {
 	result := make(map[string]string)
 	for _, envvar := range envvars {
 		s := strings.SplitN(strings.TrimSpace(envvar), "=", 2)
+		if len(s) != 2 {
```
johnmccabe: Good catch.

commands/list.go
```diff
@@ -57,8 +57,8 @@ func runList(cmd *cobra.Command, args []string) {
 	// fmt.Println(gatewayAddress)
 	functions, err := proxy.ListFunctions(gatewayAddress)
 	if err != nil {
-		log.Println(err)
-		return
+		// ListFunctions prints out its own error messages
+		return err
```
johnmccabe: I would be tempted to just return an empty error here as updating the proxy would still leave the FaaS-netes provider requiring an update. Returning `fmt.Errorf("")` minimises the impact.

@alexellis wdyt?

commands/new_function.go
```diff
@@ -127,16 +130,16 @@ functions:
...
 
-	return
+	fmt.Printf("Stack file written: %s\n", functionName+".yml")
+	return nil
 }
 
 func validTemplate(lang string) bool {
 	var found bool
-	if strings.ToLower(lang) != "dockerfile" {
+	if strings.ToLower(lang) == "dockerfile" {
```
johnmccabe: Was there an issue open for this?

commands/faas.go
```diff
@@ -53,6 +53,7 @@ func Execute(customArgs []string) {
 	checkAndSetDefaultYaml()
 
 	faasCmd.SilenceUsage = true
+	faasCmd.SilenceErrors = true
 	faasCmd.SetArgs(customArgs[1:])
 	if err := faasCmd.Execute(); err != nil {
 		os.Exit(1)
```
johnmccabe: This is where the error message should be printed in the event of failure, can use `strings.Title()` to capitalise the error text (leave the error message without capitalisation).
```
if len(err.Error()) > 0 {
    fmt.Println(strings.Title(err.Error()))
}
```

commands/bash_completion.go
Razesdark: That is a much cleaner way of doing it than the way I've been doing it. `RunE` does actually print the error messages it receives to screen in a `Error: <Error message here>` format by default unless we suppress it. 

Should we print the messages ourselves or let cobra do it?

commands/faas.go
Razesdark: Wouldn't this capitalize the first letter in each word, including making `--username` into `--Username` and `-u` into `-U`?

commands/new_function.go
Razesdark: #211

alexellis:
@johnmccabe did you have any concerns about error handling, or is this good to go to merge?

commands/new_function.go
```diff
@@ -60,35 +59,30 @@ the "Dockerfile" lang type in your YAML file.
...
 	}
 
 	if err := os.Mkdir("./"+functionName, 0700); err == nil {
 		fmt.Printf("Folder: %s created.\n", functionName)
 	}
 
 	if err := updateGitignore(); err != nil {
-		fmt.Println("Got unexpected error while updating .gitignore file.")
-		return err
+		return fmt.Errorf("Got unexpected error while updating .gitignore file: %s", err)
```
johnmccabe: should begin with lowercase character

(see https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

commands/new_function.go
```diff
@@ -60,35 +59,30 @@ the "Dockerfile" lang type in your YAML file.
...
 	if validTemplate(lang) == false {
-		fmt.Printf("%s is unavailable or not supported.\n", lang)
-		return errors.New("language is unavailable or not supported")
+		return fmt.Errorf("%s is unavailable or not supported", lang)
 	}
 
 	if _, err := os.Stat(functionName); err == nil {
-		fmt.Printf("Folder: %s already exists\n", functionName)
-		return errors.New("folder already exists")
+		return fmt.Errorf("Folder: %s already exists", functionName)
```
johnmccabe: should begin with lowercase character

commands/remove.go
```diff
@@ -65,8 +62,7 @@ func runDelete(cmd *cobra.Command, args []string) error {
 		}
 	} else {
 		if len(args) < 1 {
-			fmt.Println("Please provide the name of a function to delete")
-			return errors.New("please provide the name of a function to delete")
+			return fmt.Errorf("Please provide the name of a function to delete")
```
johnmccabe: should begin with lowercase character

commands/build.go
```diff
@@ -119,7 +115,7 @@ func build(services *stack.Services, queueDepth int, shrinkwrap bool) {
 			for function := range workChannel {
 				fmt.Printf("[%d] > Building: %s.\n", index, function.Name)
 				if len(function.Language) == 0 {
-					fmt.Println("Please provide a valid -lang or 'Dockerfile' for your function.")
+					fmt.Println("Please provide a valid --lang or 'Dockerfile' for your function.")
```
johnmccabe: should begin with lowercase character

commands/faas.go
```diff
@@ -56,6 +57,11 @@ func Execute(customArgs []string) {
 	faasCmd.SilenceErrors = true
 	faasCmd.SetArgs(customArgs[1:])
 	if err := faasCmd.Execute(); err != nil {
+		if len(err.Error()) > 2 {
```
johnmccabe: Not sure I like this, are you setting this to 2 to protect against the SplitAfterN not returning 2 elements?

How about this:

```
func capitaliseFirst(str string) string {
    for _, c := range str {
        return string(unicode.ToUpper(c)) + str[1:]
    }  
    return ""
}
```

You can then just:
```
fmt.Println(capitaliseFirst(err.Error())
```

commands/faas.go
johnmccabe: You're correct about this, my bad.

johnmccabe:
@Razesdark @alexellis just some minor comments then I'm 👍 to merge.

This is looking nice, thanks @Razesdark for the fast and responsive turnaround 🎈

commands/faas.go
```diff
@@ -88,3 +84,10 @@ func runFaas(cmd *cobra.Command, args []string) {
 	fmt.Printf(figletStr)
 	cmd.Help()
 }
+
+func capitaliseFirst(str string) string {
+	for _, c := range str {
```
alexellis: Is there nothing in `strings` that can do this? Is there a normal way of printing errors in Golang that we can tap into?

johnmccabe: See https://github.com/golang/go/wiki/CodeReviewComments#error-strings, error strings are normally printed as-is but after some other text ie "Error: %s" type of thing, which is not how they're being used in the CLI at this point.

alexellis:
Hi, is this ready to merge? There were some comments from @johnmccabe 4 days ago. Have you reviewed these yet?

Razesdark:
As far as I have gathered, the only thing that seems left up for discussion is the `capitaliseString` function needed to show error messages the way we have been showing them prior to this PR.
The golang way seems to be something along the lines of `Error: %s`. Instead of just printing out the error message like we have been doing up until now.

Personally, I think the current style of error messages are just fine, since we are exiting the program anyways, prepending "Error" or to the last message before showing a very obvious instruction or error message the messages doesn't really do anything,.

commands/faas.go
```diff
@@ -53,8 +54,10 @@ func Execute(customArgs []string) {
 	checkAndSetDefaultYaml()
 
 	faasCmd.SilenceUsage = true
+	faasCmd.SilenceErrors = true
 	faasCmd.SetArgs(customArgs[1:])
 	if err := faasCmd.Execute(); err != nil {
+		fmt.Println(capitaliseFirst(err.Error()))
```
alexellis: Let's just inline the code for capitaliseFirst.

Is there an issue with something simpler like this?

```
error := "lowercase error"
fmt.Println(strings.ToUpper(error[:1]) + error[1:])
```

Is this because you were expecting multi-byte characters?

alexellis:
Thanks for the update.

commands/faas.go
Razesdark: There was a case where we returned a 0 length error message because another function was printing errors as well. I refactored the same code, so now all error messages should be longer that 0 chars long, making it fairly safe to use the method you mentioned here.

I hadn't really considered multibyte characters, there are a few error messages that starts with user input, so I tested those.

capitaliseString would look like this:
```
$ faas-cli new test --lang ルビ
ル��ビ is unavailable or not supported
```
while using your solution works as expected
```
$ faas-cli new test --lang ルビ
ルビ is unavailable or not supported
```
I tried some arabic as well as chinese characters as well, all seems to work, but I got to admit I don't know enough about these issues to be sure.
```
$ faas-cli new test --lang 鬱    
鬱 is unavailable or not supported
$ faas-cli new test --lang وتري  
وتري is unavailable or not supported
```

commands/new_function_test.go
```diff
@@ -79,15 +80,15 @@ func runNewFunctionTest(t *testing.T, nft NewFunctionTest) {
 		"--gateway=" + defaultGateway,
 	}
 
-	stdOut := test.CaptureStdout(func() {
-		faasCmd.SetArgs(cmdParameters)
-		faasCmd.Execute()
-	})
+	/*
+		stdOut := test.CaptureStdout(func() {
```
alexellis: Should this code block be deleted?

Razesdark: done

alexellis:
Merging. Thank you for this contribution @Razesdark I know it'll make a big difference for a lot of people.

PR #213, opened 2017-11-08
Add append functionality to `fc new` command

ericstoekl:
Appends function config YAML data to existing YAML file, specified by:
`faas-cli new <FuncName> --lang <Lang> -f <YAML>`

## Description

This change allows for merging the YAML config data for a new function into an existing YAML file. It also adds tests for the new function.

This is version 2.0 of the original pull request: #141 . The difference is that this version will not alter the existing state of the YAML except for the new function you intend to append. It should leave the rest of your YAML file as it was before the `faas-cli new` command was issued.

This functionality is implemented using a regex which matches top-level blocks in the YAML file.

The regex is: `(?m:^([^\#\s]+:)\s*$)`

- Multiline
- Match a capture group starting at the beginning of the line
  - Match one or more characters that are NOT `#` and NOT whitepace, and ends with a colon
- Must be only whitespace following the capture group until the end of the string.

So for example, you have a stack YAML that looks like:

```
provider:
  name: faas
  gateway: http://localhost:8080
functions:
  url-ping:
    lang: python
    handler: ./sample/url-ping
    image: alexellis/faas-url-ping
```

This regex will match `provider:` and `functions:` only. `url-ping:` will not be matched because that line begins with whitespace.

## Motivation and Context

#128 

## How Has This Been Tested?

Three types of test cases were included:
1. Test function creation without an existing YAML file
2. Test function creation with existing YAML file, make sure it appends the new function block correctly
3. Test with invalid inputs, make sure they fail with the correct error message.

Section 1 validates by making sure the function folder and stack YAML were created, the stdout from the `new` command is what we expect, and it validates the YAML object by parsing it.
Section 2 validates by the same method as above, and in addition validates that the same functions and provider that existed before the `new` command was called are still present.
Section 3 validates by making sure that the stdout is what we expect (failure message)

## Types of changes

## Checklist:

### DIFF
--- a/commands/deploy_test.go
+++ b/commands/deploy_test.go
@@ -83,7 +83,12 @@ func Test_deploy(t *testing.T) {
 			ResponseStatusCode: http.StatusOK,
 		},
 	})
-	defer s.Close()
+
+	defer func() {
+		gateway = defaultGateway
+		image = ""
+		s.Close()
+	}()
 
 	stdOut := test.CaptureStdout(func() {
 		faasCmd.SetArgs([]string{
--- a/commands/new_function.go
+++ b/commands/new_function.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"sort"
+	"regexp"
 	"strings"
 
 	"github.com/morikuni/aec"
@@ -39,13 +40,16 @@ func init() {
 
 // newFunctionCmd displays newFunction information
 var newFunctionCmd = &cobra.Command{
-	Use:   "new FUNCTION_NAME --lang=FUNCTION_LANGUAGE [--gateway=http://domain:port] | --list)",
+	Use: `new FUNCTION_NAME --lang=FUNCTION_LANGUAGE
+				[--gateway=http://domain:port]
+				[-f=stack.yml] | --list)`,
 	Short: "Create a new template in the current folder with the name given as name",
 	Long: `The new command creates a new function based upon hello-world in the given
 language or type in --list for a list of languages available.`,
 	Example: `faas-cli new chatbot --lang node
   faas-cli new textparser --lang python --gateway http://mydomain:8080
-  faas-cli new --list`,
+  faas-cli new --list
+  faas-cli new myfunction --lang node -f stack.yml`,
 	RunE: runNewFunction,
 }
 
@@ -81,11 +85,27 @@ the "Dockerfile" lang type in your YAML file.
 		return fmt.Errorf("you must supply a function language with the --lang flag")
 	}
 
-	PullTemplates("")
+	existingYaml := len(yamlFile) > 0
+	var stackFileName string
+	if !existingYaml {
+		// We will create a new YAML file for this function
+		stackFileName = functionName + ".yml"
+	} else {
+		stackFileName = yamlFile
+		// YAML file was passed in, so parse to see if it is valid
+		parsedServices, err := stack.ParseYAMLFile(stackFileName, "", "")
+		if err != nil || parsedServices == nil {
+			fmt.Printf("Specified file (" + stackFileName + ") is not valid YAML\n")
+			return
+		}
+		//TODO: Verify that services is good
+	}
 
 	if stack.IsValidTemplate(lang) == false {
 		return fmt.Errorf("%s is unavailable or not supported", lang)
 	}
+
+	PullTemplates("")
 
 	if _, err := os.Stat(functionName); err == nil {
 		return fmt.Errorf("folder: %s already exists", functionName)
@@ -103,39 +123,36 @@ the "Dockerfile" lang type in your YAML file.
 	if strings.ToLower(lang) != "dockerfile" {
 		builder.CopyFiles("./template/"+lang+"/function/", "./"+functionName+"/", true)
 	} else {
-		ioutil.WriteFile("./"+functionName+"/Dockerfile", []byte(`FROM alpine:3.6
-# Use any image as your base image, or "scratch"
-# Add fwatchdog binary via https://github.com/openfaas/faas/releases/
-# Then set fprocess to the process you want to invoke per request - i.e. "cat" or "my_binary"
-
-ADD https://github.com/openfaas/faas/releases/download/0.6.9/fwatchdog /usr/bin
-# COPY ./fwatchdog /usr/bin/
-RUN chmod +x /usr/bin/fwatchdog
-
-# Populate example here - i.e. "cat", "sha512sum" or "node index.js"
-ENV fprocess="wc -l"
-
-HEALTHCHECK --interval=5s CMD [ -e /tmp/.lock ] || exit 1
-CMD ["fwatchdog"]
-`), 0600)
+		if err := writeDockerfile(functionName); err != nil {
+			fmt.Printf("Error writing Dockerfile to %s: %v\n", functionName, err)
+		}
 	}
 
-	stack := `provider:
-  name: faas
-  gateway: ` + gateway + `
-
-functions:
+	var stackFileText string
+	functionYAMLData := `
   ` + functionName + `:
     lang: ` + lang + `
     handler: ./` + functionName + `
     image: ` + functionName + `
 `
 
+	if !existingYaml {
+		stackFileText = newStackFileText(functionYAMLData, gateway)
+	} else {
+		existingYAMLText, err := readFileToString(stackFileName)
+		if err != nil {
+			fmt.Printf("Could not read stack file: %s\n", stackFileName)
+			return
+		}
+
+		stackFileText = insertFunctionIntoYAML(existingYAMLText, functionYAMLData)
+	}
+
 	fmt.Printf(aec.BlueF.Apply(figletStr))
 	fmt.Println()
 	fmt.Printf("Function created in folder: %s\n", functionName)
 
-	stackWriteErr := ioutil.WriteFile("./"+functionName+".yml", []byte(stack), 0600)
+	stackWriteErr := ioutil.WriteFile("./"+stackFileName, []byte(stackFileText), 0600)
 	if stackWriteErr != nil {
 		return fmt.Errorf("error writing stack file %s", stackWriteErr)
 	}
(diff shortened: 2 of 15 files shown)


alexellis:
Going to add this into the review queue

commands/new_function.go
```diff
@@ -145,3 +164,73 @@ func validTemplate(lang string) bool {
 
 	return found
 }
+
+// insertAtRegexpMatch takes a full YAML file data (fileData), and
+// data to insert at the match location (insertData). Returns a string
+// that is fileData with functionYAMLData inserted in the 'functions:'
+// block of the YAML file.
+func insertFunctionIntoYAML(fileData string, functionYAMLData string) string {
```
alexellis: What is this doing @ericstoekl ?

commands/new_function_test.go
```diff
@@ -5,37 +5,334 @@ package commands
...
 	"testing"
 
+	"github.com/openfaas/faas-cli/stack"
 	"github.com/openfaas/faas-cli/test"
 )
 
-func Test_new(t *testing.T) {
-	//TODO activate the teardown when PR#87 is merged defer teardown()
-	funcName := "test-function"
+const SuccessMsg = `(?m:Function created in folder)`
```
alexellis: When you capitalize variables in Go it has unexpected behavior of making them available globally and external to the package. Best to put these in the tests and repeat them for readability (we don't want to be jumping around) and when we absolutely have to use consts etc they should start with lowercase.

commands/new_function_test.go
```diff
@@ -5,37 +5,334 @@ package commands
...
+	},
+}
+
+var AppendFunctionTests = []NewFunctionTest{
+	{
+		title:       "new_append_1",
+		funcName:    "new-test-append-1",
+		funcLang:    "python",
+		file:        "append1.yml",
+		expectedMsg: SuccessMsg,
```
alexellis: In this example it may be OK to have the const / remove the repetition.

commands/new_function_test.go
```diff
@@ -5,37 +5,334 @@ package commands
...
+
+		testFunction := stack.Function{Language: funcLang, Image: funcName, Handler: "./" + funcName}
+		if !reflect.DeepEqual(services.Functions[funcName], testFunction) {
+			t.Fatalf("YAML `functions` section was not created correctly for file %s, got %v", funcYAML, services.Functions[funcName])
+		}
+	}
+}
+
+func runAppendFunctionTest(t *testing.T, nft NewFunctionTest) {
+	defer func() {
```
alexellis: I think we should explicitly call this before every function rather than using defer.

commands/new_function.go
ericstoekl: We need to append the function data to the existing YAML, but keep the YAML in its original order. I do this by grabbing the top-level objects in the YAML (for now that means just `provider:` and `functions:`, i.e. the lines of the YAML file that do not begin with whitespace and are not comments) and then inserting the new data into the `functions:` section. If the `functions:` section comes before the `provider:` section, then we need to know where the `provider:` section is.

The other way of doing this that does not use a regex is to loop through every line of the file, scanning for top-level objects (`functions:` and `provider:`). I figured the regex would be less code and therefore better, but I can see why we might not want to use regex. Do you think this is a reasonable situation for regex, or would looping through each line of the file be better? Or is there a better solution that I haven't thought of?

johnmccabe:
ping @ericstoekl, some comments need your eyeballs

commands/new_function_test.go
ericstoekl: You mean only have one test case for `AppendFunctionTests`? Or something like, one positive one negative test case?

ericstoekl:
Hey just an update -- I am working on this, it's kind of a gnarly merge conflict that I'm working through, but the review comments will be addressed soon. Appreciate y'all's patience!! :+1:

ericstoekl:
Attempted to address all the comments in the latest commit.

alexellis:
Still want to get this change in @ericstoekl @johnmccabe - PTAL.

alexellis:
I heard from @johnmccabe that he will be taking a quick look at this.

johnmccabe:
yep, will take a look over lunch today

ericstoekl:
Hey, I missed these recent comments, my apologies. I'm currently resolving the merge conflicts. I'll have an update out today-tomorrow.

ericstoekl:
On hold until the test issue gets sorted: https://github.com/openfaas/faas-cli/issues/299

commands/new_function.go
```diff
@@ -81,12 +85,27 @@ the "Dockerfile" lang type in your YAML file.
...
+			return fmt.Errorf("Specified file (" + stackFileName + ") is not valid YAML\n")
+		}
+		//TODO: Verify that services is good
+	}
 
 	if stack.IsValidTemplate(lang) == false {
 		return fmt.Errorf("%s is unavailable or not supported", lang)
 	}
 
+	PullTemplates("")
```
johnmccabe: The `new` will always fail with "X is unavailable or not supported" as you're validating before pulling in the templates. I assume you were testing with templates already present?

commands/new_function.go
```diff
@@ -81,12 +85,27 @@ the "Dockerfile" lang type in your YAML file.
...
+		// We will create a new YAML file for this function
+		stackFileName = functionName + ".yml"
+	} else {
+		stackFileName = yamlFile
+		// YAML file was passed in, so parse to see if it is valid
+		parsedServices, err := stack.ParseYAMLFile(stackFileName, "", "")
+		if err != nil || parsedServices == nil {
+			return fmt.Errorf("Specified file (" + stackFileName + ") is not valid YAML\n")
+		}
+		//TODO: Verify that services is good
```
johnmccabe: What were you intending to do here? If the PR is complete as-is then remove this comment, if its a non-blocking enhancement then raise as a separate issue if not implementing now.

commands/new_function.go
```diff
@@ -103,39 +122,35 @@ the "Dockerfile" lang type in your YAML file.
...
+	} else {
+		existingYAMLText, err := readFileToString(stackFileName)
+		if err != nil {
+			return fmt.Errorf("Could not read stack file: %s\n", stackFileName)
+		}
+
+		stackFileText = insertFunctionIntoYAML(existingYAMLText, functionYAMLData)
+	}
+
 	fmt.Printf(aec.BlueF.Apply(figletStr))
```
johnmccabe: Kinda of out of scope here, but I don't think we should be printing the logo for commands other than help, since it makes scripting with `faas-cli` messy - @alexellis wdyt?

commands/new_function.go
```diff
@@ -81,12 +85,27 @@ the "Dockerfile" lang type in your YAML file.
 		return fmt.Errorf("you must supply a function language with the --lang flag")
 	}
 
-	PullTemplates("")
+	existingYaml := len(yamlFile) > 0
+	var stackFileName string
+	if !existingYaml {
+		// We will create a new YAML file for this function
+		stackFileName = functionName + ".yml"
```
johnmccabe: This should be `stack.yaml` not `functionname.yml`

commands/new_function.go
```diff
@@ -81,12 +85,27 @@ the "Dockerfile" lang type in your YAML file.
 		return fmt.Errorf("you must supply a function language with the --lang flag")
 	}
 
-	PullTemplates("")
+	existingYaml := len(yamlFile) > 0
```
johnmccabe: Should be looking for the default `stack.yaml` if no explicit path is provided.

commands/new_function.go
```diff
@@ -81,12 +85,27 @@ the "Dockerfile" lang type in your YAML file.
...
+	var stackFileName string
+	if !existingYaml {
+		// We will create a new YAML file for this function
+		stackFileName = functionName + ".yml"
+	} else {
+		stackFileName = yamlFile
```
johnmccabe: Looks like this change isn't pushed through to the point at which the write is echo'd to stdout. For example creating a function `myfn` and passing the yaml explicitly `--yaml stack.yaml` results in the following being output to the user.
```
Function created in folder: myfn
Stack file written: myfn.yml
```
It should reflect the stack file written to:
```
Stack file written: stack.yml
```

commands/new_function.go
```diff
@@ -81,12 +85,27 @@ the "Dockerfile" lang type in your YAML file.
...
+	}
 
 	if stack.IsValidTemplate(lang) == false {
 		return fmt.Errorf("%s is unavailable or not supported", lang)
 	}
 
+	PullTemplates("")
+
 	if _, err := os.Stat(functionName); err == nil {
 		return fmt.Errorf("folder: %s already exists", functionName)
```
johnmccabe: No checks are in place to ensure you don't end up with multiple identical function keys in the yaml, for example from repeated `new` commands where the function dir is deleted in between runs.
```
provider:
  name: faas
  gateway: http://localhost:8080
functions:
  newpython:
    lang: python
    handler: ./newpython
    image: newpython

  newpython2:
    lang: python
    handler: ./newpython2
    image: newpython2

  newpython2:
    lang: python
    handler: ./newpython2
    image: newpython2

  newpython2:
    lang: python
    handler: ./newpython2
    image: newpython2
```

commands/new_function.go
johnmccabe: @ericstoekl can you remind me what the need to keep the `yaml` in the same order was for please? I've forgotten the context here..

test/utils.go
```diff
@@ -130,3 +130,25 @@ func CaptureStdout(f func()) string {
...
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return err
+	}
+	return out.Close()
```
johnmccabe: Why are you calling Close twice?

commands/testdata/new_function/invalid3.yml
```diff
@@ -0,0 +1,10 @@
+proovider:
```
johnmccabe: Can you use an obviously deliberate invalid keys in testcases like this - `invalidprovider:` for example, makes it much clearer for readers.

commands/new_function.go
```diff
@@ -153,3 +168,73 @@ func printAvailableTemplates(availableTemplates []string) string {
...
+				retData = fileData + functionYAMLData
+			}
+			break
+		}
+	}
+	return retData
+}
+
+func writeDockerfile(functionName string) error {
+	return ioutil.WriteFile("./"+functionName+"/Dockerfile", []byte(`FROM alpine:3.6
```
johnmccabe: Shouldn't be hardcoding this inline, at the very least pull it out into a variable - @alexellis does this belong outside of the cli?

commands/deploy_test.go
```diff
@@ -83,7 +83,12 @@ func Test_deploy(t *testing.T) {
 			ResponseStatusCode: http.StatusOK,
 		},
 	})
-	defer s.Close()
+
+	defer func() {
+		gateway = defaultGateway
```
johnmccabe: This feels odd, cleared to initialise vars within the tests for clarity rather than having this outside? Could you point to why this is needed by the way?

alexellis:
Merged in part via #301 - it takes a simplistic approach to get past the problems we were having trying to make a complete solution. @johnmccabe please add suggestions to that PR so we can pick it up later.

alexellis:
@ericstoekl thanks for your work on this and for testing #301

Issue #253, opened 2017-12-10
issue with function file permissions 


ganesshkumar:
I am creating a template and provide some default scripts(with execute permission for the user) that belong to the function
```
|--template
  |-- my-template-name
     |--function
        |-- default-script
```

I pulled the template from github, examined the template folders. It has the execute permission as it had been set. 

Later, I create a function out of my template and in the new generated function folder, the script has only read and write permission for the user. It does not have the execute permission. 

## Expected Behaviour
The newly created function should have the permission of the files as in the template.

## Current Behaviour
The generated function files have only read and write permission.

## Possible Solution
I have narrowed down the code to [new_function.go](https://github.com/openfaas/faas-cli/blob/master/commands/new_function.go#L104) and [build.go](https://github.com/openfaas/faas-cli/blob/master/builder/build.go#L89). I am not sure how go handles the permission during file creation.

## Steps to Reproduce (for bugs)
```
$ faas-cli template pull https://github.com/ganesshkumar/openfaas-template-java
$ faas-cli new --lang java-gradle gradle-function
```
Since, I am creating a gradle based function, there is a `gradlew` executable in the function folder, that will build the jar.

**Difference between the permissions in function file and template file**
```
$  ll java-function/gradlew
-rw-r-----  1 ganessh  staff   5.2K Dec 10 14:59 java-function/gradlew

$  ll template/java-gradle/function/gradlew
-rwxr-xr-x  1 ganessh  staff   5.2K Dec 10 14:30 template/java-gradle/function/gradlew
```

It leads to failure of the project build using docker
```
$  functions faas-cli build -f java-function.yml

[0] > Building: java-function.
Clearing temporary build folder: ./build/java-function/
Preparing ./java-function/ ./build/java-function/function
Building: java-function with java-gradle template. Please wait..
Sending build context to Docker daemon  33.28kB
Step 1/21 : FROM anapsix/alpine-java:8_jdk as builder
 ---> 46796749789c
Step 2/21 : USER root
 ---> Using cache
 ---> 6b354a844d9d
Step 3/21 : RUN mkdir -p /root/src/
 ---> Using cache
 ---> 151625409012
Step 4/21 : ADD function /root/src/
 ---> Using cache
 ---> fee9b46f6986
Step 5/21 : WORKDIR /root/src/
 ---> Using cache
 ---> 51b126cb2a6a
Step 6/21 : RUN ./gradlew
 ---> Running in 988f3123c86c
[...]

PR #254, opened 2017-12-11
while copying a file use file mode from the source file

ganesshkumar:
Signed-off-by: Ganessh Kumar R P <rpganesshkumar@gmail.com>

## Description
This is a fix to #253, where cp removes any execute permission on the file by using default `0660` on the destination file.

This commit reads the mode of the source file using `os.Stat(src)` and use the source file's permission on the destination file.  The destination file will have the same permission as the source file.

## Motivation and Context

- [x] I have raised an issue to propose this change ([#253](https://github.com/openfaas/faas-cli/issues/253))

## How Has This Been Tested?
- [x] `go build` and `go test`
- [ ]  use the build fass-cli to check the file permission (yet to do)

## Types of changes

## Checklist:
- ~~My change requires a change to the documentation.~~
- ~~I have updated the documentation accordingly.~~
- ~~I have added tests to cover my changes.~~
- [x] All ~~new and~~ existing tests passed.

### DIFF
--- a/builder/build_test.go
+++ b/builder/build_test.go
@@ -1,0 +1,77 @@
+// Copyright (c) OpenFaaS Project 2017. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package builder
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"testing"
+)
+
+func Test_CopyFiles(t *testing.T) {
+	fileModes := []int{0600, 0640, 0644, 0700, 0755}
+
+	dir := os.TempDir()
+	for _, mode := range fileModes {
+		// set up a source folder with 2 file
+		srcDir, srcDirErr := setupSourceFolder(2, mode)
+		if srcDirErr != nil {
+			log.Fatal("Error creating source folder")
+		}
+		defer os.RemoveAll(srcDir)
+
+		// create a destination folder to copy the files to
+		destDir, destDirErr := ioutil.TempDir(dir, "openfaas-test-destination-")
+		if destDirErr != nil {
+			t.Fatalf("Error creating destination folder\n%v", destDirErr)
+		}
+		defer os.RemoveAll(destDir)
+
+		CopyFiles(srcDir, destDir+"/", true)
+		err := checkDestinationFiles(destDir, 2, mode)
+		if err != nil {
+			t.Fatalf("Destination file mode differs from source file mode\n%v", err)
+		}
+	}
+}
+
+func setupSourceFolder(numberOfFiles, mode int) (string, error) {
+	dir := os.TempDir()
+	data := []byte("open faas")
+
+	// create a folder for source files
+	srcDir, dirError := ioutil.TempDir(dir, "openfaas-test-source-")
+	if dirError != nil {
+		return "", dirError
+	}
+
+	// create n files inside the created folder
+	for i := 1; i <= numberOfFiles; i++ {
+		srcFile := fmt.Sprintf("%s/test-file-%d", srcDir, i)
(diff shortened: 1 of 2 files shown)


alexellis:
Hi, thanks for this change. I'm surprised we haven't seen this error yet so it's good you've caught it now.

@rgee0 can you review?

I would ideally like a unit / integration test for this too. Do we have any tests that touch this yet? cc/ @nicholasjackson 

Alex

itscaro:
PR seems good to me

ganesshkumar:
@itscaro  I am thinking of create a builder_test.go in the same folder and unit test the `cp` function. Do you have any other tests in mind?

itscaro:
@ganesshkumar I think unit test for cp function is sufficient for this PR, if you can cover all cases of the `cp` function that would be great

ganesshkumar:
@itscaro I have added tests, which tests CopyFiles function.
The test does the following,
* Creates a source folder and populates some files in it with given permission
* Creates a destination folder to copy the files to
* Uses `CopyFiles` function to copy the contents of source to destination
* Check all the destination files for the mode in which they are created

builder/build_test.go
```diff
@@ -0,0 +1,77 @@
...
+	fileModes := []int{0600, 0640, 0644, 0700, 0755}
+
+	dir := os.TempDir()
+	for _, mode := range fileModes {
+		// set up a source folder with 2 file
+		srcDir, srcDirErr := setupSourceFolder(2, mode)
+		if srcDirErr != nil {
+			log.Fatal("Error creating source folder")
+		}
+		defer os.RemoveAll(srcDir)
```
alexellis: You may also want to do this up front for the case where remove didn't work or is running in parallel on the same system. Can you randomise the path any more than this?

builder/build_test.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"testing"
+)
+
+func Test_CopyFiles(t *testing.T) {
```
alexellis: @itscaro thoughts on these tests?

builder/build_test.go
ganesshkumar: @alexellis, the temporary folders had some random integers attached at the end of the directory name example `openfaas-test-destination-290650684`. So I guess that will handle the case where several tests are run in parallel on the same system.

However, I notice that if a Fatal is logged then the directed till that point are not deleted. So before throwing the Fatals we should clean up all the temporary directories created. I will update the PR tomorrow with the relevant changes.

itscaro: @ganesshkumar I prefer using t.Error() and call a function doing the teardown ops, as in fetch_template_test.go

alexellis: @itscaro I need to merge these changes and release to unblock @ganesshkumar and a couple of other people for the Java template. Please could you raise a PR to fix the tests how you like them after this merge?

alexellis:
@itscaro please can you raise a PR to fix up anything you think needs to be changed?

builder/build_test.go
itscaro: @ganesshkumar IMO, it's more relevant to assert file modes for ops done by `cp` function, which is in charge of doing the stuff with file modes. The test for `CopyFiles` should test only the number of files copied and the name of the destination dir.

The current test will fail on Windows, you can rename `build_test` to `build_unix_test` and add this build flag `// +build darwin dragonfly freebsd linux netbsd openbsd solaris` at the beginning of the test file

Example:
https://github.com/openfaas/faas-cli/compare/master...itscaro:pr/254?expand=1

itscaro:
@alexellis I have raised #258 as an update for this PR

Issue #256, opened 2017-12-16
Backlog: prefer rolling update over delete/create


alexellis:
## Description

For `faas deploy` we should prefer rolling updates over our current delete/create pattern.

You can get similar behaviour by doing `--update=true --replace=false`.

## Expected Behaviour

CLI should try to update service rather than delete/create. We did this initially to get to "update-like" behaviour faster but it causes issues with K8s which is slow to delete definitions.

When deploying functions let's first try an update, if we get a 404 we'll then send in a create instead.

Help wanted for this one, looking for the community to lead the changes including testing.

ericstoekl:
derek assign: ericstoekl

ericstoekl:
Derek assign: ericstoekl

PR #257, opened 2017-12-18
[WIP] Use rolling update instead of delete/create when function exists

ericstoekl:
## Description

This change is to perform in-place update of functions rather than delete/create when the function already exists.

If the function does not already exist and we get 404, create the function.

## Motivation and Context

#256 

## How Has This Been Tested?

Testing is improved in `proxy/deploy_test.go` by adding scaffolding with array-of-struct test cases.

We need to improve testing in this area by adding more test cases (hence why this is WIP).

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/proxy/deploy.go
+++ b/proxy/deploy.go
@@ -31,12 +31,10 @@ func DeployFunction(fprocess string, gateway string, functionName string, image 
 
 	if update == true && statusCode == http.StatusNotFound {
 		// Re-run the function with update=false
-		update = false
-		Deploy(fprocess, gateway, functionName, image, language, replace, envVars, network, constraints, update, secrets, labels, functionResourceRequest1)
+		Deploy(fprocess, gateway, functionName, image, language, replace, envVars, network, constraints, false, secrets, labels, functionResourceRequest1)
 	}
 }
 
-// Call FaaS server to deploy a new function
 func Deploy(fprocess string, gateway string, functionName string, image string,
 	language string, replace bool, envVars map[string]string, network string,
 	constraints []string, update bool, secrets []string, labels map[string]string,
--- a/proxy/deploy.go
+++ b/proxy/deploy.go
@@ -27,18 +27,17 @@ func DeployFunction(fprocess string, gateway string, functionName string, image 
 	constraints []string, update bool, secrets []string, labels map[string]string,
 	functionResourceRequest1 FunctionResourceRequest) {
 
-	statusCode := DeployFunctionImpl(fprocess, gateway, functionName, image, language, replace, envVars, network, constraints, update, secrets, labels, functionResourceRequest1)
+	statusCode := Deploy(fprocess, gateway, functionName, image, language, replace, envVars, network, constraints, update, secrets, labels, functionResourceRequest1)
 
 	if update == true && statusCode == http.StatusNotFound {
 		// Re-run the function with update=false
 		update = false
-		DeployFunctionImpl(fprocess, gateway, functionName, image, language, replace, envVars, network, constraints, update, secrets, labels, functionResourceRequest1)
+		Deploy(fprocess, gateway, functionName, image, language, replace, envVars, network, constraints, update, secrets, labels, functionResourceRequest1)
 	}
-
 }
 
-// DeployFunction call FaaS server to deploy a new function
-func DeployFunctionImpl(fprocess string, gateway string, functionName string, image string,
+// Call FaaS server to deploy a new function
+func Deploy(fprocess string, gateway string, functionName string, image string,
 	language string, replace bool, envVars map[string]string, network string,
 	constraints []string, update bool, secrets []string, labels map[string]string,
 	functionResourceRequest1 FunctionResourceRequest) int {
@@ -113,14 +112,14 @@ func DeployFunctionImpl(fprocess string, gateway string, functionName string, im
 	SetAuth(request, gateway)
 	if err != nil {
 		fmt.Println(err)
-		return -1
+		return http.StatusInternalServerError
 	}
 
 	res, err := client.Do(request)
 	if err != nil {
 		fmt.Println("Is FaaS deployed? Do you need to specify the --gateway flag?")
 		fmt.Println(err)
-		return -1
+		return http.StatusInternalServerError
 	}
 
 	if res.Body != nil {
--- a/proxy/deploy.go
+++ b/proxy/deploy.go
@@ -22,10 +22,26 @@ type FunctionResourceRequest struct {
 	Requests *stack.FunctionResources
 }
 
-// DeployFunction call FaaS server to deploy a new function
 func DeployFunction(fprocess string, gateway string, functionName string, image string,
 	language string, replace bool, envVars map[string]string, network string,
-	constraints []string, update bool, secrets []string, labels map[string]string, functionResourceRequest1 FunctionResourceRequest) {
+	constraints []string, update bool, secrets []string, labels map[string]string,
+	functionResourceRequest1 FunctionResourceRequest) {
+
+	statusCode := DeployFunctionImpl(fprocess, gateway, functionName, image, language, replace, envVars, network, constraints, update, secrets, labels, functionResourceRequest1)
+
+	if update == true && statusCode == http.StatusNotFound {
+		// Re-run the function with update=false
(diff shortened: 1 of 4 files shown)


proxy/deploy.go
```diff
@@ -79,14 +95,14 @@ func DeployFunction(fprocess string, gateway string, functionName string, image
 	SetAuth(request, gateway)
 	if err != nil {
 		fmt.Println(err)
-		return
+		return -1
```
BurtonR: Not sure if this matters, but the usage shows that you're returning a status code (in the function call, not the signature), so should this return 500 (`http.StatusInternalServerError`) instead?

Same comment for line 105 with the other `return -1`

Purely a nit-pick thing, just thinking about consistency for down the road

proxy/deploy.go
```diff
@@ -21,10 +21,26 @@ type FunctionResourceRequest struct {
...
 }
 
-// DeployFunction call FaaS server to deploy a new function
 func DeployFunction(fprocess string, gateway string, functionName string, image string,
 	language string, replace bool, envVars map[string]string, network string,
-	constraints []string, update bool, secrets []string, labels map[string]string, functionResourceRequest1 FunctionResourceRequest) {
+	constraints []string, update bool, secrets []string, labels map[string]string,
+	functionResourceRequest1 FunctionResourceRequest) {
+
+	statusCode := DeployFunctionImpl(fprocess, gateway, functionName, image, language, replace, envVars, network, constraints, update, secrets, labels, functionResourceRequest1)
```
alexellis: Not keen on names like "Impl".. we can just call this `Deploy`? I saw we broke out a function for this - could we make the change one level higher? Does that even make sense?

proxy/deploy.go
alexellis: Yes what does -1 mean?

proxy/deploy.go
ericstoekl: By make the change one level higher -- do you mean having the retry logic (retry if 404 error occurs) in `commands/deploy.go`, in the `runDeploy()` function?

alexellis:
```
How Has This Been Tested?
```

@ericstoekl please tell us what end-to-end testing has been done on this.

@BurtonR have you played with it yet?

proxy/deploy.go
```diff
@@ -21,10 +21,25 @@ type FunctionResourceRequest struct {
...
-// DeployFunction call FaaS server to deploy a new function
 func DeployFunction(fprocess string, gateway string, functionName string, image string,
 	language string, replace bool, envVars map[string]string, network string,
-	constraints []string, update bool, secrets []string, labels map[string]string, functionResourceRequest1 FunctionResourceRequest) {
+	constraints []string, update bool, secrets []string, labels map[string]string,
+	functionResourceRequest1 FunctionResourceRequest) {
+
+	statusCode := Deploy(fprocess, gateway, functionName, image, language, replace, envVars, network, constraints, update, secrets, labels, functionResourceRequest1)
+
+	if update == true && statusCode == http.StatusNotFound {
```
alexellis: Just a *nit* here - can we avoid setting `update = false` and just pass `false` through the message signature? It will make debugging easier. Or you could create a new local variable as an alternative: `updateFunction := false`

Just to avoid destructively updating the value

proxy/deploy.go
```diff
@@ -21,10 +21,25 @@ type FunctionResourceRequest struct {
...
+		// Re-run the function with update=false
+		update = false
+		Deploy(fprocess, gateway, functionName, image, language, replace, envVars, network, constraints, update, secrets, labels, functionResourceRequest1)
+	}
+}
+
+// Call FaaS server to deploy a new function
```
alexellis: While here let's remove the comment here.

proxy/deploy_test.go
```diff
@@ -61,23 +60,31 @@ func Test_DeployFunction_Not2xx(t *testing.T) {
...
-			false,
+			dpt.update,
 			[]string{},
 			map[string]string{},
 			FunctionResourceRequest{},
 		)
 	})
 
-	r := regexp.MustCompile(`(?m:Unexpected status: 404)`)
+	r := regexp.MustCompile(dpt.expectedOutput)
```
alexellis: Can we change the name of `dpt` to something else?

BurtonR:
I've tested this by running the tests (all pass) as well as running through the commands as a normal user. 

`faas-cli deploy -f tester.yml` deploys a new function
`faas-cli deploy -f tester.yml` again, updates the function
`faas-cli deploy -f tester.yml --update=false --replace=true` replaces the function
All working as expected 👍 

_However_, the `faas-cli deploy --help` shows incorrect usage. This is not new for this PR, but something that's been lurking for a bit, I just forgot about it as I rarely use that any more.

In `commands/deploy.go`, lines 85-86, the help text for the deploy function is missing the second flag
Both flags are required as the CLI uses default values and must be explicitly overridden when passing the `--replace=true` flag
```
  faas-cli deploy -f ./samples.yml --replace=false
  faas-cli deploy -f ./samples.yml --update=true
```
should be:
```
  faas-cli deploy -f ./samples.yml --replace=false --update=true
  faas-cli deploy -f ./samples.yml --replace=true --update=false
```
or perhaps, remove the flags from the `update=true` as that's default behavior now.

alexellis:
@ericstoekl can you address the comment from @BurtonR please?

ericstoekl:
@alexellis just noticed there's been activity here, I'll have an update for this later today.

ericstoekl:
Just updated this, thanks to @BurtonR for helping with testing and suggestions. The latest commit addresses the comments.

@alexellis, I have tested this by running the tests as well as manually going through the scenario with a docker swarm deployment.

One thought I have is combining the `--update` and `--replace` flags into a single flag. What are your thoughts on this (Burton and/or Alex?)

BurtonR:
I agree, I think one flag would be more intuitive since we default one of them to true always and there's no way to check (that I'm aware of) if the flag was actually passed in.

So, just having the `--replace` flag would make a lot of sense. It updates automatically, but if the `--replace` flag is present, remove and deploy. Bonus points for also including a `-r` option :)

Also, I think that should go into a different PR since this one is focused on setting the default behavior.

proxy/deploy_test.go
```diff
@@ -14,43 +14,42 @@ import (
...
+	replace             bool
+	update              bool
+	expectedOutput      string
+}
 
-	r := regexp.MustCompile(`(?m:Deployed.)`)
-	if !r.MatchString(stdout) {
-		t.Fatalf("Output not matched: %s", stdout)
-	}
+var DeployProxyTests = []DeployProxyTest{
```
alexellis: The capital letter means this is exported outside the package. Was that the intention?

proxy/deploy_test.go
```diff
@@ -61,23 +60,31 @@ func Test_DeployFunction_Not2xx(t *testing.T) {
...
-	r := regexp.MustCompile(`(?m:Unexpected status: 404)`)
+	r := regexp.MustCompile(deployTest.expectedOutput)
 	if !r.MatchString(stdout) {
 		t.Fatalf("Output not matched: %s", stdout)
 	}
 }
 
+func Test_RunDeployProxyTests(t *testing.T) {
+	for _, tst := range DeployProxyTests {
+		t.Run(tst.title, func(t *testing.T) {
```
alexellis: Just wanted to ask why you did it this way instead of our usual coding style of test-cases and an in-line struct?

ericstoekl: I could change it so the test cases are defined in the `Test_RunDeployProxyTests()` method, but I like having `runDeployProxyTest()` as a separate function because it makes it easier to read

proxy/deploy_test.go
ericstoekl: Was not the intention, I'll fix this.

alexellis:
I think users will find the 404 message confusing. Please could you open a new PR so that we get this output?

When it exists:


```
Function already exists, attempting rolling-update.

Deployed.
URL: http://192.168.0.66:8080/function/cat

200 OK
```

When it doesn't:

```
Deployed.
URL: http://192.168.0.66:8080/function/cat

200 OK
```

PR #260, opened 2017-12-18
[WIP] Re-factoring for errors

### LINKED ISSUE #259: [re-factoring] Gather errors into one place & use errors.New() to define errors
## Context
Currently errors are generated with `fmt.Errorf` which is not re-usable to compare errors in unit-test, there are many copy-paste of the error strings.

Instead of creating errors with `fmt.Errorf`, we can pre-defined errors with `errors.New()` and we can re-use the errors everywhere in the package.

This applies to errors whose string is not formatted with `printf` since the string is constant.

itscaro:
## Motivation and Context
Currently errors are generated with `fmt.Errorf` which is not re-usable to compare errors in unit-test. Instead of creating errors with `fmt.Errorf`, we can pre-defined errors with `errors.New()` and we can re-use the errors everywhere in the package.

- [x] I have raised an issue to propose this change (fixes #259)

## How Has This Been Tested?
With current test suite

## Types of changes

- [x] Improvement of quality of code

## Checklist:

### DIFF
--- a/commands/login.go
+++ b/commands/login.go
@@ -35,7 +35,7 @@ var loginCmd = &cobra.Command{
 	Use:   `login [--username USERNAME] [--password PASSWORD] [--gateway GATEWAY_URL]`,
 	Short: "Log in to OpenFaaS gateway",
 	Long:  "Log in to OpenFaaS gateway.\nIf no gateway is specified, the default local one will be used.",
-	Example: `  faas-cli login -u user -p password --gateway http://localhost:8080
+	Example: `  faas-cli login -u user -p password --gateway ` + defaultGateway + `
   cat ~/faas_pass.txt | faas-cli login -u user --password-stdin --gateway https://openfaas.mydomain.com`,
 	RunE: runLogin,
 }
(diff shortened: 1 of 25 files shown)


stefanprodan:
@itscaro this looks great!

alexellis:
@itscaro there was no design/approval or agreement given for this work.. Please can you update the issue you raised and explain the pros/cons as it was not obvious from the description. 

cc @rgee0

commands/login.go
```diff
@@ -20,13 +20,15 @@ var (
...
 	passwordStdin bool
+	insecure      bool
 )
 
 func init() {
 	loginCmd.Flags().StringVarP(&gateway, "gateway", "g", defaultGateway, "Gateway URL starting with http(s)://")
 	loginCmd.Flags().StringVarP(&username, "username", "u", "", "Gateway username")
 	loginCmd.Flags().StringVarP(&password, "password", "p", "", "Gateway password")
 	loginCmd.Flags().BoolVar(&passwordStdin, "password-stdin", false, "Reads the gateway password from stdin")
+	loginCmd.Flags().BoolVar(&insecure, "insecure", false, "Do not validate SSL certificate")
```
stefanprodan: The default should be set to `true` so it doesn't break any CI pipelines users could have.

stefanprodan: I think is better to add this flag in a separate PR. Refactoring should not add new functionality.

alexellis:
Thanks for explaining the reasoning behind this PR over Zoom. The idea is to de-duplicate the errors from the tests and Go code. The question I have is when we have a string in the error and we have to interpolate it i.e. "you cannot deploy language type: %s, due to error: %s". cc @stefanprodan

johnmccabe:
@alexellis what about creating custom error types rather than just the default, you can carry out simple comparisons on the type, but also detailed checks on the error types fields... this kind of thing?

```go
package main

import (
	"fmt"
	"reflect"
)

type LanguageDeployError struct {
	lang     string
	errormsg string
}

func (e LanguageDeployError) Error() string {
	return fmt.Sprintf("you cannot deploy language type: %s, due to error: %s", e.lang, e.errormsg)
}

func main() {
	lang := "python"
	errormsg := "stuff blowed up"
	err := LanguageDeployError{lang, errormsg}

	fmt.Println("Error output:")
	fmt.Println(err)

	fmt.Println("\nError type:")
	fmt.Printf("%s\n", reflect.TypeOf(err))
}
```

alexellis:
@itscaro please see feedback from @johnmccabe 

Can you list some of the pros/cons of this refactoring?

I'm reading the diff for build.go for instance and finding it very hard to read. I can imagine it would be hard to review what the messages are in the future.

Example:

https://github.com/openfaas/faas-cli/pull/260/files#diff-e3adafa635d215187fe5408840f529b9

I do like the idea from John that if we need to generate a complex error we could template it with a type, but this CLI is relatively small / simple. I'm not sure where it could be applied where it has a low overhead and good benefit.

alexellis:
Needs a rebase and I'm unsure of the benefit

Issue #237, opened 2017-11-30
Feature: OpenFaaS store integration


alexellis:
The `faas` CLI should allow users to browse and deploy functions from the [OpenFaaS Function Store](https://github.com/openfaas/store)

Three new commands:

* `faas store list`
Fetches/parses JSON and lists all items

* `faas store inspect`
Gives specific info on one item

* `faas store deploy`
Same as faas-cli deploy except pre-loaded with arguments from JSON file. Must take all valid overrides such as `--gateway` and `--env`

## Example usage:

```
$ faas store list

Function                 Description
figlet                       Create ASCII logos

$ faas store inspect figlet

Function                 Description                                                     Image                         Function Process
figlet                       Create ASCII logos                                         alexellis/asciilogos     node handler.js


$ faas store deploy figlet
```

## Out of scope

* `faas store search`
* `faas store list --filter`

amirkarimi:
If it doesn't have a very high priority, I would like to work on this. I suppose it can get the data directly from the store repo?

alexellis:
Yes we download and parse the JSON file over the GitRaw CDN.

alexellis:
Thank you @amirkarimi - just to update Amir has started working on this change. Look out for a PR - maybe @BurtonR + @ericstoekl can collaborate / help out with stuff.

PR #265, opened 2017-12-23
Feature: OpenFaaS store integration

amirkarimi:
## Description

The faas CLI should allow users to browse and deploy functions from the OpenFaaS Function Store.
As described in https://github.com/openfaas/faas-cli/issues/237.

## How Has This Been Tested?

Three new commands tested manually like other existing commands. As described in https://github.com/openfaas/faas-cli/issues/237.

We can also add some automated tests in another PR.

## Types of changes

## Checklist:

- [ ] I have signed-off my commits with `git commit -s` (I've signed just last one)


alexellis:
Hey @amirkarimi - excited to see this coming together. Probably best to squash all these merge commits down into one and then you can sign them off at the same time too. Let me know if I can.

amirkarimi:
@alexellis Cool, I think it's easier to be done on Github during the merge, so can you please squash them?

NOTES.md
```diff
@@ -0,0 +1,43 @@
+| Command | Description |
```
johnmccabe: This file shouldn't be in here?  It's my notes that were committed in error during the Cobra PR and was subsequently removed.

template/node-arm64/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
johnmccabe: Was this intended? Not clear what this has to do with the `store` subcommand.

template/node/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
johnmccabe: Was this intended? Not clear what this has to do with the `store` subcommand.

commands/store.go
```diff
@@ -0,0 +1,262 @@
...
+
+func makeHTTPClient(timeout *time.Duration) http.Client {
+	if timeout != nil {
+		return http.Client{
+			Timeout: *timeout,
+			Transport: &http.Transport{
+				Proxy: http.ProxyFromEnvironment,
+				DialContext: (&net.Dialer{
+					Timeout: *timeout,
+					// KeepAlive: 0,
```
johnmccabe: Should get rid of commented out options.

commands/store.go
```diff
@@ -0,0 +1,262 @@
...
+				}).DialContext,
+				// MaxIdleConns:          1,
+				// DisableKeepAlives:     true,
+				IdleConnTimeout:       120 * time.Millisecond,
```
johnmccabe: I'd be inclined to set these defaults in constant at the top of the file rather than magic values here.

commands/store.go
```diff
@@ -0,0 +1,262 @@
...
+				ExpectContinueTimeout: 1500 * time.Millisecond,
+			},
+		}
+	}
+
+	// This should be used for faas-cli invoke etc.
```
johnmccabe: Is this copied from elsewhere (reviewing on phone), if so are you sure it's not already available in the command scope from the invoke.command?

alexellis:
We have code to make HTTP clients already. Should reuse that really

template/node-arm64/build.sh
amirkarimi: I think these files including NOTES.md have been there when I forked, before sending the PR I merge the upstream but seems that for some reason it didn't remove these files! Anyway, removed all of them.

amirkarimi:
@alexellis @johnmccabe Thanks for your review. Your comments are addressed. Let me know if there is anything else.

commands/store.go
```diff
@@ -0,0 +1,240 @@
...
+	if len(args) < 1 {
+		return fmt.Errorf("please provide the function name")
+	}
+
+	item, err := findFunction(args[0])
+	if err != nil {
+		return err
+	}
+
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', 0)
```
alexellis: Are you able to write this all into a string that is printed rather than directly to `os.Stdout`?

The `\t` mechanism may not do what you expect in all circumstances. I have used spacing instructions for `fmt.Sprintf()` before I think they would work well here.

Example: https://gobyexample.com/string-formatting

commands/store.go
```diff
@@ -0,0 +1,240 @@
...
+
+	"github.com/openfaas/faas-cli/proxy"
+	"github.com/spf13/cobra"
+)
+
+var storeAddress string
+
+const defaultStore = "https://cdn.rawgit.com/openfaas/store/master/store.json"
+
+type storeItem struct {
```
alexellis: I wonder if we should move this into the `schema` or a similar package? @johnmccabe @rgee0

commands/store.go
```diff
@@ -0,0 +1,240 @@
...
+		return item, err
+	}
+
+	for _, item = range items {
+		if item.Name == functionName || item.Title == functionName {
+			return item, nil
+		}
+	}
+
+	return item, fmt.Errorf("function '%s' not found", functionName)
```
alexellis: Is this an error or should this return empty and be handled in the calling code?

commands/store.go
```diff
@@ -0,0 +1,240 @@
...
+			return nil, fmt.Errorf("server returned unexpected status code: %d - %s", res.StatusCode, string(bytesOut))
+		}
+	}
+	return results, nil
+}
+
+func findFunction(functionName string) (storeItem, error) {
+	var item storeItem
+
+	items, err := storeList(storeAddress)
```
alexellis: Might be better to have findFunction hold a single responsibility and pass the result of storeList via a function parameter i.e. `findFunction(functionName string, storeItems)`

alexellis:
Hi @amirkarimi you'll need to sign off all the commits before we can take it through. From what I can see you've only signed-off on one of them.

Might be worth waiting until we're finished here then doing this.

Thanks for working on this feature and for responding to the feedback from John and myself. 👍

commands/store.go
amirkarimi: That's how I started working on this but finally I found [`tabwriter`](https://golang.org/pkg/text/tabwriter/) which helps generating text-based tables. `\t` here is just a separator character it could be anything, it's not rendering tab character at all. Moreover, it won't render the text directly to output as it needs to receive all the rows data and calculate the required space for each column.

commands/store.go
amirkarimi: It's just an empty result. If we `return nil, nil` then the caller should check the item value in addition to error. I'm new to go so not sure if it's a good practice to return `nil` without any error? Seems unsafe to me.

amirkarimi: Oh got your point. I should've used `fmt.Sprintf`. Right?

amirkarimi: Anyway, I removed the error from return type and return `nil` when the function is not found

amirkarimi:
@alexellis Thanks for your feedback. I've updated the code and answered some comments. Just not sure about [the comment about moving storeItem to schema package](https://github.com/openfaas/faas-cli/pull/265#discussion_r158791806). Just let me know if we want to move it to `schema` package (and it means to create a new folder called `schema` at root)?

alexellis:
Yes how about we move `storeItem` to `StoreItem` in the `schema` package?

commands/store.go
```diff
@@ -0,0 +1,245 @@
...
+
+	item := findFunction(args[0], storeItems)
+	if item == nil {
+		return fmt.Errorf("function '%s' not found", functionName)
+	}
+
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', 0)
+	fmt.Fprintln(w)
+	fmt.Fprintln(w, "FUNCTION\tDESCRIPTION\tIMAGE\tFUNCTION PROCESS\tREPO")
+	fmt.Fprintf(w, "%s\t%s\t%s\t%s\t%s\n",
```
alexellis: My comment seems to have been removed here by the rebasing. Please use fmt.Sprintf() and build up a string. It should use the spacing directives not tabs. Thank you 👍

amirkarimi: @alexellis I explained it in another comment that these `\t` are just a separator as I used tabwriter. They're not rendering tab characters. It automatically aligns the columns using spaces.

commands/store.go
```diff
@@ -0,0 +1,240 @@
...
+	if err != nil {
+		return err
+	}
+
+	item := findFunction(args[0], storeItems)
+	if item == nil {
+		return fmt.Errorf("function '%s' not found", functionName)
+	}
+
+	var b bytes.Buffer
```
alexellis: Thanks for iterating on the feedback. Here's where I'm going with this - the method should be single responsibility and also unit testable. So printing straight to stdout makes that really hard.

One method should create a string or buffer and that should be printed elsewhere. It means we can test the formatting function and resulting output.

Can you provide an example of wrapping or truncating output with long descriptions in the PR description please?

amirkarimi:
@alexellis You're totally right about unit tests. I tried to apply the single responsibility principle to the store command functions. Also added `verbose` flag to switch between truncated and long descriptions.

Here is an example of truncated (default) output:

```
$ faas-cli store inspect Inception   

FUNCTION  DESCRIPTION                              IMAGE               PROCESS          REPO
Inception This is a forked version of the work ... alexellis/inception python3 index.py https://github.com/faas-and-furious/inception-function
```

And this the verbose one:
```
$ faas-cli store inspect -v Inception

FUNCTION  DESCRIPTION                                                                                                                      IMAGE               PROCESS          REPO
Inception This is a forked version of the work by Magnus Erik Hvass Pedersen - it has been re-packaged as an OpenFaaS serverless function. alexellis/inception python3 index.py https://github.com/faas-and-furious/inception-function
```

The same switch is working for `store list` command.

The verbose output would be wrapped by the terminal app and might not seem very beautiful but this the same behavior I've seen from other terminal apps.

alexellis:
It feels like we're working against the row/column format because even in the short-mode it's still very long and squashed up. How about version 0.2 in a table? What do you think?

```
Function
-------------
Inception

Description
--------------
Inception This is a forked version of the work by Magnus Erik Hvass Pedersen - it has been re-packaged as an OpenFaaS serverless function.

Image
-------------
alexellis/inception

Repo
-------
https://github.com/faas-and-furious/inception-function
```

alexellis:
I can't merge this until we fix the build either.

This is the part that's failing - which involves several steps.

I'd run each one individually to see which fails.

```
/bin/sh -c license-check -path ./ --verbose=false  && go test $(go list ./... | grep -v /vendor/ | grep -v /template/|grep -v /build/) -cover  && VERSION=$(git describe --all --exact-match `git rev-parse HEAD` | grep tags | sed 's/tags\///')  && GIT_COMMIT=$(git rev-list -1 HEAD)  && CGO_ENABLED=0 GOOS=linux go build --ldflags "-s -w -X github.com/openfaas/faas-cli/version.GitCommit=${GIT_COMMIT} -X github.com/openfaas/faas-cli/version.Version=${VERSION}" -a -installsuffix cgo -o faas-cli
```

alexellis:
Hey @amirkarimi is it possible for you to reset the history of the merge and then add in and re-push?

I want to avoid these commits:

> Merge remote-tracking branch 'upstream/master' into store-commands

```
git reset HEAD~1
git add .
git commit -s -m "Whatever changes I made"
git push store-commands upstream --force
```

rgee0:
Not sure if you've updated since I pulled the branch to look at (git suggests not), but the failure in the branch I pulled seemed to be in `Test_deploy` in `commands/deploy_tests.go`.  It looks as though the test is seeing a DELETE and a POST while the test condition specifies to expect a PUT.  There is also some weirdness around slice sizes in `utils.go`.

I also pulled a clean upstream master and ran the test.  It passed.  I then put your changes in to this branch and it started to fail.  I then deleted `store.go` leaving the remaining changes and the test passed.

I haven't entirely got my head around it yet but posting an update in case it helps narrow this down.

amirkarimi:
@alexellis @rgee0 Thanks for your comments. Found the problem; I was reusing the existing vars of the main `deploy` command for `store deploy` command inside `store.go`. Seems that the default value of `replace` flag was updated to `false` in master branch but as I copied the old set of flags into `store.go` the default value for this flag was `true`, causing the `deploy` command to delete the function (DELETE request) before deploy. I need to tidy up these flags by defining them once or having separate flags for main deploy and store deploy commands.

alexellis:
@amirkarimi where are we with your PR? Are we good to merge or is there any pending feedback?

Issue #271, opened 2018-01-01
Make "Dockerfile" a language template


alexellis:
Make "Dockerfile" a language template

## Expected/Current Behaviour

The "Dockerfile" language template is currently kept partially in code. We should find a way to make it into an OpenFaaS template so that it can be versioned in the same way as the other templates.

I started the work in #221 but I need to focus on other areas right now.

I'd like help with this. 

Alex

alexellis:
@itscaro @johnmccabe @nicholasjackson do any of you guys want to start off the New Year with a little OpenFaaS PR? :smile_cat:

itscaro:
@alexellis I take this PR if no one took it already

itscaro:
Derek assign: itscaro

itscaro:
Derek assign: me

dungeonmaster18:
Ping @alexellis close this

PR #221, opened 2017-11-11
Make Dockerfile language a proper template

alexellis:
Signed-off-by: Alex Ellis <alexellis2@gmail.com>

## Description

Make Dockerfile language a proper template

## Motivation and Context

Remove temporary hard-coded entry

## How Has This Been Tested?

Checkout PR, then:
```
$ go build && sudo cp ./faas-cli /usr/local/bin/faas-cli ; rm -rf doc1 doc1.yml; faas-cli new --lang dockerfile doc1 ; echo ; cat ./doc1/Dockerfile  && faas-cli build -f doc1.yml
```
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] I have signed-off my commits with `git commit -s`
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #274, opened 2018-01-07
[WIP] Use exec to clone templates

LucasRoesler:
## Description

This address proposal #266.  This WIP pull request allows an initial review of the changes.

Breifly this will
- Add generic vcs support
- Add specific git support to pull templates from arbitrary git repos
- Add a global debug flag

Instead of using Github archive zips, this uses the users installed git environment to clone projects to a temp folder and then copy the functions found within the clone. After the copy is done, it removes the temp folder. 

This breaks the current template pull because it requires the user to supply a proper git repository url as would be used in `git clone`.

## Motivation and Context

- [X] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

## How Has This Been Tested?

I have only tested this manually using the following commands

``` sh
faas-cli template pull
faas-cli template pull --overwrite
faas-cli template pull git@github.com:LucasRoesler/openfaas-templates.git
faas-cli template pull git@github.com:LucasRoesler/openfaas-templates.git --overwrite
faas-cli template pull git@github.com:LucasRoesler/openfaas-templates.git
faas-cli template pull git@github.com:LucasRoesler/openfaas-templates.git --overwrite
```

I am current working through and updating the unit tests.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [X] I have signed-off my commits with `git commit -s`
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #266, opened 2017-12-24
[Proposal] Pull templates from private Git repos using CLI


LucasRoesler:
The `faas-cli template pull` command should support pulling template definitions from both public and private git repositories.  It would also be good to support hosts beyond Github.  The cli should support pulling via ssh or via https with username and password.

Supporting private repos and other git servers improve the support for users with high security constraints, e.g. large enterprises that host on-prem.  

## Expected Behaviour
If I run 
```sh
faas-cli template pull https://github.com/LucasRoesler/privatefunctions0.git
faas-cli template pull git@github.com:LucasRoesler/privatefunctions1.git
faas-cli template pull git@gitlab.com:LucasRoesler/privatefunctions2.git
faas-cli template pull https://lucasroesler.com/bitbucket/privatefunctions3.git
```
I expect that a `template` folder in the project is populated with the functions from each of the repositories.  If there is a name conflict, I expect the command to report the issue, it should ask me to skip or overwrite. If I pass the `--overwrite` flag, it should assume overwrite for any name conflicts.

## Current Behaviour
```sh
$ faas-cli template pull git@github.com:LucasRoesler/privatefunctions1.git
The repository URL must be in the format https://github.com/<owner>/<repository>
```

## Possible Solution
Mimic the solution in `go get` by using `exec.Command` to run the git command and utilize the users git configuration including credentials.  See https://github.com/golang/go/blob/master/src/cmd/go/internal/get/vcs.go

alexellis:
@LucasRoesler - this is a sensible request and I know it will help your business use-case. I'd be concerned about reinventing the wheel here or having to vendor too many packages - so I like the idea of using a fork/exec out to `git` - that way we get access to any credential cache/helper for `git` too.

We'd need to bring the folder down into a temporary location and then find the templates folder and bring it back to the local context.

Would you be able to put together a sequence diagram to attach to the issue?

## Other notes:

This will also mean we'd remove the code responsible for handling .zip files as this is the only purpose for the code.

Suggested reviewers: @ericstoekl @itscaro

LucasRoesler:
For a sequence diagram, were you thinking something like this
![faas-cli-template-pull-flow](https://user-images.githubusercontent.com/891889/34352828-173a266c-ea25-11e7-9142-c9e01655e522.png)

alexellis:
I was thinking about a UML sequence diagram, but this works well to convey the flow.

alexellis:
[...]

PR #275, opened 2018-01-13
Dockerfile as a language

itscaro:
PR over openfaas/faas-cli#221 for Issue openfaas/faas-cli#271

docker-language


alexellis:
Hey @itscaro do you know what caused the CI to break here?

itscaro:
@alexellis I need to port back this PR openfaas/templates#4 into dataset for tests.

itscaro:
@alexellis should be ok now CI, tests for new_function will get templates from master_test.zip instead of using committed files in testdata/new_function, we will have that zip as dataset for testing from now on.

There are some tests using github.com to get the zip file, I fixed one in fetch_templates_test.go

alexellis:
FYI @itscaro I will be merging PR #266 which is likely to impact on the zip code / tests.

Makefile
```diff
@@ -7,4 +10,9 @@ build_redist:
 	./build_redist.sh
 
 test-unit:
-	go test $(shell go list ./... | grep -v /vendor/ | grep -v /template/ | grep -v build) -cover
+	go test $(shell go list ./... | grep -v /vendor/ | grep -v /template/ | grep -v /build/ | grep -v /sample/) $(TEST_OPTS) -cover
+test-unit-docker:
+	docker run --rm \
```
alexellis: I'm not sure we need this?

builder/build.go
```diff
@@ -97,11 +98,11 @@ func CopyFiles(src string, destination string, recursive bool) {
 
 		if file.IsDir() == false {
 
-			cp(src+"/"+file.Name(), destination+file.Name())
+			cp(filepath.Join(src, file.Name()), filepath.Join(destination, file.Name()))
```
alexellis: I don't think this change is relevant to the Dockerfile language template. Can you comment please?

commands/new_function.go
```diff
@@ -63,12 +62,9 @@ func runNewFunction(cmd *cobra.Command, args []string) error {
 			}
 		}
 
-		fmt.Printf(`Languages available as templates:
-` + printAvailableTemplates(availableTemplates) + `
+		fmt.Println(`Languages available as templates:
```
alexellis: This may as well be a Printf with a %s directive.

commands/new_function.go
```diff
@@ -91,36 +87,19 @@ the "Dockerfile" lang type in your YAML file.
 		return fmt.Errorf("folder: %s already exists", functionName)
 	}
 
-	if err := os.Mkdir("./"+functionName, 0700); err == nil {
+	if err := os.Mkdir(functionName, 0700); err == nil {
```
alexellis: Are you sure that this is exactly the same behaviour as previously? Why did it change?

alexellis:
@itscaro we have some merge conflicts now due to #274 - please could you take a look and also at the other feedback? Ideally this should only contain the minimum change-set for the Dockerfile language and not additional refactoring around filename.join etc.

Makefile
itscaro: it's a shortcut to run test on unix-based image, since I'm on windows mostly

builder/build.go
itscaro: It's not relevant, I can make a PR regrouping changes related to coding style, tell me if you prefer that way :-)

commands/new_function.go
itscaro: I changed it to avoid using concatenation in file path operations as suggested in another PR. The behaviour is the same.

"./" is no need, the path is relative to current working dir

itscaro:
@alexellis OK, I'm separating refactoring changes to another PR

alexellis:
That GIF is 💯 🏅

commands/fetch_templates_test.go
```diff
@@ -35,21 +36,28 @@ func Test_PullTemplates(t *testing.T) {
 
 // setupLocalTemplateRepo will create a local copy of the core OpenFaaS templates, this
 // can be refered to as a local git repository.
-func setupLocalTemplateRepo(t *testing.T) string {
+func setupLocalTemplateRepo(t *testing.T, useMock bool) string {
```
alexellis: What code calls this with "false" ?

commands/template_pull_test.go
```diff
@@ -124,3 +127,72 @@ func Test_repositoryUrlRemoteRegExp(t *testing.T) {
 		})
 	}
 }
+
+// templatePullLocalTemplateRepo executes `template pull` on a local repository to get templates
+func templatePullLocalTemplateRepo(t *testing.T) {
```
alexellis: What's this additional code doing with zips? I thought the changes @LucasRoesler made removed the need for zips now?

itscaro: I have a zip containing the templates repo, that zip is to used to have a local repo for tests.

`commands/testdata/templates.git.zip`

commands/fetch_templates_test.go
itscaro: The current repo tests against `github.com/openfaas/templates`, that's why I let the possibility to get this current behaviour. I think we do not this, tests should always use mock.

Should I remove the parameter?

alexellis: Could we avoid the zip and generate the git folder instead?
```
git init
echo etc > ./template.yml
git commit -m "Initial"
```

etc?

Or add the .git folder as a sub-module?

commands/template_pull_test.go
itscaro: replace with git submodule & git init; git add; git commit to create a local repo during the tests

commands/new_function.go
```diff
@@ -91,36 +86,19 @@ the "Dockerfile" lang type in your YAML file.
...
-ENV fprocess="wc -l"
+	builder.CopyFiles(filepath.Join("template", lang, "function"), functionName)
 
-HEALTHCHECK --interval=5s CMD [ -e /tmp/.lock ] || exit 1
-CMD ["fwatchdog"]
-`), 0600)
-	}
-
-	stack := `provider:
+	stackYaml := `provider:
```
itscaro: stack => stackYaml to avoid conflict with imported package

stack/language_template.go
```diff
@@ -50,13 +50,17 @@ func ParseYAMLDataForLanguageTemplate(fileData []byte) (*LanguageTemplate, error
 
 func IsValidTemplate(lang string) bool {
 	var found bool
+
+	// TODO harmonise to lowercase when fetching template & parsing yaml
+	// Ensure that `lang` is lowercase in case of Dockerfile
 	if strings.ToLower(lang) == "dockerfile" {
```
itscaro: lowercase only dockerfile in order to avoid 3rd template with mixed cases

PR #278, opened 2018-01-15
error string starts with lowercase and do not contain trailing '.'

itscaro:
Coding style


alexellis:
@itscaro LGTM but we have to fix that CI error - please can you find out what's going wrong?

itscaro:
@alexellis Some tests base on the error messages which were modified, I'll finish this tonight :)

PR #290, opened 2018-01-22
use docker/pkg/term to correct colored texts on Windows & create a function to print figlet

### LINKED ISSUE #289: Colored texts in PowerShell native console
## Current Behaviour
In PowerShell native console, ANSI escape are used and the PowerShell does not understand. The printed logs are not much readable.

With WSL (Windows Subsystem for Linux), we can still use the CLI correctly, but it's preferable that we can use it with the same UX in the PowerShell native console.

PowerShell native
PowerShell native

PowerShell in WSL - ubuntu
PowerShell in WSL - ubuntu

## Possible Solution
Use docker/pkg/term

itscaro:
## Motivation and Context

- [x] I have raised an issue to propose this change (resolve #289)

## How Has This Been Tested?
PowerShell & Bash in WSL
image

Debian 9 via Bash in WSL
image

## Types of changes

## Checklist:

### DIFF
--- a/commands/faas.go
+++ b/commands/faas.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/docker/docker/pkg/term"
 	"github.com/spf13/cobra"
 )
 
@@ -47,6 +48,9 @@ func resetForTest() {
 }
 
 func init() {
+	// Setup terminal std
+	term.StdStreams()
+
 	faasCmd.PersistentFlags().StringVarP(&yamlFile, "yaml", "f", "", "Path to YAML file describing function(s)")
 	faasCmd.PersistentFlags().StringVarP(&regex, "regex", "", "", "Regex to match with function names in YAML file")
 	faasCmd.PersistentFlags().StringVarP(&filter, "filter", "", "", "Wildcard to match with function names in YAML file")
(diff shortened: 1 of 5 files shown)


commands/version.go
```diff
@@ -40,16 +41,20 @@ func runVersion(cmd *cobra.Command, args []string) {
...
 		fmt.Printf("Version: %s\n", version.BuildVersion())
 	}
 }
 
+func printFiglet() {
+	figletColoured := aec.BlueF.Apply(figletStr)
+	if runtime.GOOS == "windows" {
+		figletColoured = aec.GreenF.Apply(figletStr)
+	}
+	ansi.Printf(figletColoured)
```
alexellis: Thanks for the change. I just don't understand why this stopped working.. We also have to test this on Linux and Mac before merging.

itscaro: @alexellis CI breaks because I did not commit vendored files

commands/faas.go
```diff
@@ -47,6 +48,9 @@ func resetForTest() {
 }
 
 func init() {
+	// Setup terminal std
+	term.StdStreams()
```
itscaro: This will do the job :-)

commands/version.go
dungeonmaster18: It works propely with linux @itscaro . Just if you can add the dependency it would be great.

itscaro:
Test results

johnmccabe:
@itscaro LGTM, colour rendering as expected in Powershell 5.1.16299.98 and cmd.exe on Win10.0.16299.98

alexellis:
Merged - thank you for this fix 👍

Issue #186, opened 2017-10-24
Proposal: Build and run tests on Windows with AppVeyor


stefanprodan:
## Expected Behaviour

We should build the CLI and run the unit tests on Windows to ensure everything works ok.

## Current Behaviour
No automated CI for Windows.

## Possible Solution
Reuse the AppVeyor setup from the watchdog project and add go test to it. 

Build script:
https://github.com/openfaas/faas/blob/master/watchdog/build.ps1

Docker file:
https://github.com/openfaas/faas/blob/master/sample-functions/BaseFunctions/golang/Dockerfile.win

ericstoekl:
Derek add label: proposal

ericstoekl:
Derek add label: skill/beginner

ivanayov:
@stefanprodan is this still a valid proposal?
I can see the PR implementing it closed. Should we close the issue?

stefanprodan:
Yes I'll close it. thanks for the ping

PR #298, opened 2018-01-25
Add appveyor CI for tests on Windows

itscaro:
In case it's worth using Appveyor

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,24 @@
+version: "{build}"
+
+# Source Config
+clone_folder: c:\gopath\src\github.com\openfaas\faas-cli
+
+# Build host
+environment:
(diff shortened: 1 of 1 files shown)


alexellis:
What is this giving us over and above building with Linux and cross compilation?

itscaro:
@alexellis it's just for running windows-specific test cases

stefanprodan:
This would cover #186

itscaro:
@stefanprodan thanks for pointing out the related issue

alexellis:
What is available on the "free tier"? How long will it take us to cut releases and validate PRs if we have to do builds on AppVeyor and Travis?

johnmccabe:
I would be inclined to just try this - can always revert/disable.

PR #302, opened 2018-01-28
ensure that the handler path is valid before building

itscaro:
## Description
Currently, when the handler path is not valid, build passes with the `function` directory coming from template, we need to validate the handler path so that the build will stop if the function path is not correct.

## Motivation and Context

Should I raise an issue?

## Types of changes

## Checklist:

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -27,16 +27,16 @@ func BuildImage(image string, handler string, functionName string, language stri
 			}
 
 			tempPath = handler
-			if _, err := os.Stat(handler); err != nil {
-				fmt.Printf("Unable to build %s, %s is an invalid path\n", image, handler)
-				fmt.Printf("Image: %s not built.\n", image)
-
+			if err := ensureHandlerPath(image, handler); err != nil {
 				return
 			}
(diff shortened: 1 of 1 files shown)


builder/build.go
```diff
@@ -105,3 +105,14 @@ func buildFlagString(nocache bool, squash bool, httpProxy string, httpsProxy str
 
 	return buildFlags
 }
+
+func ensureHandlerPath(image string, handler string) error {
+	if _, err := os.Stat(handler); err != nil {
+		fmt.Printf("Unable to build %s, %s is an invalid path\n", image, handler)
```
alexellis: Good changes. Can we have this ensure function actually just return an error instead and then return that in the command so it bubbles up to the main handler?

itscaro: @alexellis I chose to put 2 fmt here to avoid repeating them in `BuildImage()` function, the `ensure` func returns already an error
When we re-work on build command, we can remove these fmt then.

alexellis: In this instance I'd rather have the repetition in the calling code and a simpler function that doesn't print messages and also return an error. It should do one or the other.

alexellis:
LGTM 👍

Issue #317, opened 2018-02-06
faas-cli new with lang 'Dockerfile' results in empty function


ericstoekl:
## Expected Behaviour

When running `faas-cli new <func_name> --lang Dockerfile` on Linux (tested on Ubuntu 16.04) you get an empty function folder.

## Current Behaviour

```
~/github/functions/cli2func $ faas new nmap --lang Dockerfile
2018/02/05 20:34:41 No templates found in current directory.
2018/02/05 20:34:41 Attempting to expand templates from https://github.com/openfaas/templates.git
2018/02/05 20:34:41 Fetched 11 template(s) : [csharp dockerfile go go-armhf node node-arm64 node-armhf python python-armhf python3 ruby] from https://github.com/openfaas/templates.git
Folder: nmap created.
...

Function created in folder: nmap
Stack file written: nmap.yml
~/github/functions/cli2func $ ls
nmap  nmap.yml  template
~/github/functions/cli2func $ ls nmap/
~/github/functions/cli2func $ 
```

## Possible Solution

Either add `Dockerfile/` folder to the `openfaas/templates` repository (the `Dockerfile/` folder would be identical to the `dockerfile/` folder), or apply my patch:
https://github.com/ericstoekl/faas-cli/commit/54a293da30221f8fd0351381e355f0d1b865c17e

## Steps to Reproduce (for bugs)

1. Get on ubuntu linux
2. `faas new func --lang Dockerfile`
3. Check for existence of `func/Dockerfile`
4.

## Your Environment

* Docker version `docker version` (e.g. Docker 17.0.05 ):
17.12.0
* Are you using Docker Swarm or Kubernetes (FaaS-netes)?

* Operating System and version (e.g. Linux, Windows, MacOS):
Ubuntu 16.04
* Link to your project or a code example to reproduce issue:

kenfdev:
Confirmed this bug occurs. But only on Ubuntu (not Mac)

Below is Mac version (Dockerfile exists)
```
[➜  ~ ] faas new nmap --lang Dockerfile
2018/02/06 13:20:09 No templates found in current directory.
2018/02/06 13:20:09 Attempting to expand templates from https://github.com/openfaas/templates.git
2018/02/06 13:20:11 Fetched 11 template(s) : [csharp dockerfile go go-armhf node node-arm64 node-armhf python python-armhf python3 ruby] from https://github.com/openfaas/templates.git
Folder: nmap created.
  ___                   _____           ____
[...]

PR #318, opened 2018-02-06
Convert case internally : Dockerfile to dockerfile

itscaro:
## Motivation and Context

- [x] I have raised an issue to propose this change (#317)

## Types of changes

## Checklist:

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -65,7 +65,17 @@ via flags.`,
   faas-cli build -f ./stack.yml --regex "fn[0-9]_.*"
   faas-cli build --image=my_image --lang=python --handler=/path/to/fn/ 
                  --name=my_fn --squash`,
-	RunE: runBuild,
+	PreRunE: preRunBuild,
+	RunE:    runBuild,
+}
+
+// preRunBuild validates args & flags
+func preRunBuild(cmd *cobra.Command, args []string) error {
+	if language == "Dockerfile" {
+		language = "dockerfile"
+	}
+
+	return nil
 }
 
 func runBuild(cmd *cobra.Command, args []string) error {
--- a/commands/build.go
+++ b/commands/build.go
@@ -71,9 +71,7 @@ via flags.`,
 
 // preRunBuild validates args & flags
 func preRunBuild(cmd *cobra.Command, args []string) error {
-	if language == "Dockerfile" {
-		language = "dockerfile"
-	}
+	language, _ = validateLanguageFlag(language)
 
 	return nil
 }
--- a/commands/build.go
+++ b/commands/build.go
@@ -50,7 +50,7 @@ var buildCmd = &cobra.Command{
   faas-cli build --image IMAGE_NAME
                  --handler HANDLER_DIR
                  --name FUNCTION_NAME
-                 [--lang <ruby|python|python3|node|csharp|Dockerfile>]
+                 [--lang <ruby|python|python3|node|csharp|dockerfile>]
                  [--no-cache] [--squash]
                  [--regex "REGEX"]
 				 [--filter "WILDCARD"]
@@ -115,7 +115,7 @@ func build(services *stack.Services, queueDepth int, shrinkwrap bool) {
 			for function := range workChannel {
 				fmt.Printf(aec.YellowF.Apply("[%d] > Building %s.\n"), index, function.Name)
 				if len(function.Language) == 0 {
-					fmt.Println("Please provide a valid --lang or 'Dockerfile' for your function.")
+					fmt.Println("Please provide a valid language for your function.")
 				} else {
 					builder.BuildImage(function.Image, function.Handler, function.Name, function.Language, nocache, squash, shrinkwrap)
 				}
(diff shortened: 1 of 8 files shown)


rgee0:
I'm really not sure about the merits of deprecating `Dockerfile`.  The whole world is already far too familiar with a capital D.  

It does feel like we should be able to be agnostic to case without too much of a fuss.

alexellis:
@itscaro we need to update this - https://github.com/openfaas/faas-cli/blob/master/samples.yml#L36 and this - https://github.com/openfaas/faas/blob/master/sample-functions/samples.yml

commands/build.go
```diff
@@ -65,7 +65,19 @@ via flags.`,
...
-	RunE: runBuild,
+	PreRunE: preRunBuild,
+	RunE:    runBuild,
+}
+
+// preRunBuild validates args & flags
+func preRunBuild(cmd *cobra.Command, args []string) error {
+	if language == "Dockerfile" {
+		// TODO DEPRECATION: special case to handle --lang Dockerfile
+		fmt.Println(aec.YellowF.Apply("/!\\ DEPRECATION NOTICE: Please use 'dockerfile' as language instead of 'Dockerfile', language will accept only lowercases in future releases"))
```
alexellis: Can we avoid duplicating the message 3x times?

alexellis: What does `/!\\` do? Could Deprecation notice be less shouty/loud?

commands/deploy.go
```diff
@@ -89,7 +90,19 @@ via flags. Note: --replace and --update are mutually exclusive.`,
...
                   --gateway=http://remote-site.com:8080 --lang=python
                   --env=MYVAR=myval`,
-	RunE: runDeploy,
+	PreRunE: preRunDeploy,
+	RunE:    runDeploy,
+}
+
+// preRunDeploy validates args & flags
+func preRunDeploy(cmd *cobra.Command, args []string) error {
+	if language == "Dockerfile" {
```
alexellis: What happens if someone is using a YAML file rather than CLI args?

itscaro:
@alexellis I forgot to re-integrate the check in build.go, just did it and removed also the special case in language_template.go `IsValidTemplate()`

commands/common.go
```diff
@@ -0,0 +1,18 @@
...
+
+package commands
+
+import (
+	"fmt"
+
+	"github.com/morikuni/aec"
+)
+
+func validateLanguageFlag(language *string) {
```
itscaro: move to common.go to remove repetition

commands/new_function.go
```diff
@@ -18,12 +18,11 @@ import (
 
 var (
 	appendFile string
-	lang       string
 	list       bool
 )
 
 func init() {
-	newFunctionCmd.Flags().StringVar(&lang, "lang", "", "Language or template to use")
+	newFunctionCmd.Flags().StringVar(&language, "lang", "", "Language or template to use")
```
itscaro: use the global var `language` instead

itscaro:
@rgee0 docker image is lower case, for language like PHP which is mostly referred in upper case, in the docker world, it's `php` instead
We can apply this rule to language template too.

```
docker pull PHP
invalid reference format: repository name must be lowercase
```

rgee0:
@itscaro I am thinking more about convention.  Looking at all our repos, we follow the convention, uppercase D for `Dockerfile`.  From my perspective if I'm used to typing `Dockerfile` and within OF I need to remember its `dockerfile` then I would find that hugely frustrating.

itscaro:
@alexellis @rgee0 So I will make change to YAML files: dockerfile to Dockerfile, is that correct before I do the change?

rgee0:
See also #320

alexellis:
@itscaro I updated those YAML files.. do you know if there are any others which need doing?

alexellis:
I also merged #320 as it seemed like a smaller change to put into a minor patch release.

itscaro:
@alexellis it covers only new_function case instead of the globality of this bug.

commands/common.go
```diff
@@ -0,0 +1,20 @@
...
+	"errors"
+)
+
+func validateLanguageFlag(language string) (string, error) {
+	var err error = nil
+
+	if language == "Dockerfile" {
+		language = "dockerfile"
+
+		err = errors.New("Dockerfile was converted to dockerfile")
```
alexellis: What is this error for? ( I thought we were re-moving the message? )

commands/build.go
```diff
@@ -115,7 +123,7 @@ func build(services *stack.Services, queueDepth int, shrinkwrap bool) {
 			for function := range workChannel {
 				fmt.Printf(aec.YellowF.Apply("[%d] > Building %s.\n"), index, function.Name)
 				if len(function.Language) == 0 {
-					fmt.Println("Please provide a valid --lang or 'Dockerfile' for your function.")
+					fmt.Println("Please provide a valid language for your function.")
```
alexellis: looks fine

alexellis:
Makes sense @itscaro .. @ericstoekl Quan is saying that we may have missed one or two cases in the previously merged PR. PTAL

commands/common.go
itscaro: Just in case we need to display the event, it's not a real error, just a notice to know that we modified the language variable

alexellis:
Hi @itscaro I'm looking to merge this now

Issue #297, opened 2018-01-24
Provide a command to self-update to latest version


itscaro:
It would be useful to have a sub-command like `version update --beta` to perform a self-update

## Expected Behaviour
`version update` updates to latest stable version - version without suffix `-beta`

`version update --beta` updates to latest version including version suffixed with `-beta`

Progress bar would be nice but not required

![image](https://user-images.githubusercontent.com/266018/35350715-134aa960-013f-11e8-8159-4cf2f7ed17fc.png)

![image](https://user-images.githubusercontent.com/266018/35350944-ba9558dc-013f-11e8-98b9-d723e5a737eb.png)

![image](https://user-images.githubusercontent.com/266018/35350850-737b87aa-013f-11e8-8241-c36039c71cd0.png)

Attempt to address this issue: https://github.com/openfaas/faas-cli/compare/master...itscaro:version-update-cmd

alexellis:
Thanks for the idea Minh. Since we have a mixture of package manages including `brew` - this may sound like a good idea on paper but could cause problems. Let's keep it in the issue tracker for reference but I don't think we should work on this right now.

dungeonmaster18:
Derek add label: proposal

alexellis:
Derek add label: revisit

alexellis:
I think having people update their CLIs would be useful but can we do it in another way like nagging them? What do you think @johnmccabe ? You are a packaging/config guy

itscaro:
this feature would be great on linux, since users can run directly ./faas-cli version update (like other tools, helm, minishift, minikube), instead of getting to a documentation to file the `curl command` with the link :)

alexellis:
How about we just add a help message instead?

```
$ faas-cli upgrade

For Linux type in :

 curl .....

For Mac and Homebrew:

  brew upgrade faas-cli

```

dungeonmaster18:
I like @alexellis idea because any way we have brew formula.

ericstoekl:
I agree that I like @alexellis suggestion of just outputting a help message. We all want to use a package manager eventually anyway.

rgee0:
The API rate limit could be a factor here unless we shift the version lookup 'somewhere else'

alexellis:
I'll close and mark for #revisit

PR #328, opened 2018-02-17
Update "faas-cli version" to check for latest version

itscaro:
## Description
When user run `faas-cli version`, CLI does a check for newer release, if there is one, display the info to user.

## Motivation and Context

- [x] I have raised an issue to propose this change (#297)

## Types of changes

## Checklist:

### DIFF
--- a/commands/version.go
+++ b/commands/version.go
@@ -4,12 +4,21 @@
 package commands
 
 import (
+	"encoding/json"
+	"errors"
 	"fmt"
+	"io/ioutil"
 	"runtime"
+	"time"
 
 	"github.com/morikuni/aec"
+	"github.com/openfaas/faas-cli/proxy"
 	"github.com/openfaas/faas-cli/version"
 	"github.com/spf13/cobra"
+)
+
+const (
+	githubApiReleases = "https://api.github.com/repos/openfaas/faas-cli/releases"
 )
 
 // GitCommit injected at build-time
@@ -17,8 +26,18 @@ var (
 	shortVersion bool
 )
 
+type githubApiRelease struct {
+	PreRelease  bool      `json:"prerelease"`
+	Tag         string    `json:"tag_name"`
+	HtmlUrl     string    `json:"html_url"`
+	CreatedAt   time.Time `json:"created_at"`
+	PublishedAt time.Time `json:"published_at"`
+	Name        string    `json:"name"`
+	Body        string    `json:"body"`
+}
+
 func init() {
-	versionCmd.Flags().BoolVar(&shortVersion, "short-version", false, "Just print Git SHA")
+	versionCmd.Flags().BoolVar(&shortVersion, "short-version", false, "Just print the version")
 
 	faasCmd.AddCommand(versionCmd)
 }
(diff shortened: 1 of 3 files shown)


commands/version.go
```diff
@@ -54,6 +80,33 @@ func printFiglet() {
 	fmt.Printf(figletColoured)
 }
 
+func getLatestVersion() (*githubApiRelease, error) {
+	timeout := 1 * time.Second
```
alexellis: 1 second may not be enough.

commands/version.go
```diff
@@ -54,6 +80,33 @@ func printFiglet() {
...
+	client := proxy.MakeHTTPClient(&timeout)
+
+	if resp, err := client.Get(githubApiReleases); err != nil {
+		return nil, err
+	} else {
```
alexellis: Is the else needed here?

alexellis:
Hi @itscaro this looks useful. It may be incorrect for `brew` and the other distribution methods, but it should always be right for the `curl` method.

@johnmccabe what do you think?

BurtonR:
ping @johnmccabe and @itscaro  :smile: 
This would be really helpful and possibly help some people who have issues that are fixed in a newer release of the cli by letting them know that there is a new version available without having to go look for it

johnmccabe:
@alexellis / @itscaro there is a time cost there and its not predictable, so default timeouts could be tricky - I'm not convinced hiding this behind the version command makes sense though. Doing so implies that the user suspects there may be a new version and runs that command - I think the version command is better left just returning the current binaries version.

Few thoughts..

1. pick a timeout and have `version` wait for that on every execution 
   1. short timeout : you either swallow the failed check or let the user know the check failed (what it they just want the current version) - unreliable check
   2. long timeout : what about the user that just wanted to check the version and doesn't care about the check - reliable check but potentially inconvenient
2. add a `--check` flag to `version` and use that to check for a new version
3. the version check runs for all commands, but only periodically. You can disable/configure how often it checks, daily/weekly/monthly - the cli checks the date on a check file under `.openfaas` when deciding whether or not to perform a check, when checking the user should be informed clearly via stderr - ie let them know the check is being performed and its result - you could potentially block the subsequent commands execution
4. whichever option is decided there should be some feedback to the user to indicate that the command has not hung, cycling elipsis, rotating slash, a timer (I prefer a timer) etc (this feedback should be suppressible ofc for anyone wanting to use the check in a script)

BurtonR:
I like option 2, to explicitly say "see if there's a new version". What about making it an opt-out rather than opt-in? ie: `--no-check`

My thinking is that someone may be using the CLI and (in theory) while submitting an issue, they run `faas version` to see that they are a few versions behind. Maybe they would upgrade and find their issue has already been solved? 

For the timeout, we could use the short timeout so it's not as inconvenient, but show a message indicating that the check took longer than expected. Don't fail, just give some indication that an attempt was made, but we can't tell if there is a new version.

alexellis:
Nobody has requested this feature as of yet, but updating the CLI is something we'd like people to do often. I'm not sure what to do with this PR right now.

PR #330, opened 2018-02-19
Check for username, ip, or registry on faas push

### LINKED ISSUE #329: faas-cli push should not work without a user or registry prefix [help wanted]
Some new users are trying to run `faas-cli push` without reading or understanding the instructions on adding their own user account as a prefix to the image name.

## Expected Behaviour

```
$ faas push --image function1
or
$ faas push -f ./file.yml

Unable to push image "function1" without a username prefix such as user1/function.
If you need a Docker Hub account you can sign up here: https://hub.docker.com
```

## Current Behaviour

```
Pushing to docker.io/library/function1:latest

Access denied...
```

## Possible Solution

Pre-validate any CLI or YAML file and prevent going any further unless all of the images defined have a valid prefix. Also take note of the `skip_build` flag which means an image will not be built.

Test scenarios:

* no prefix i.e. `function1:latest` (fail)
* prefix given i.e. `alexellis2/function1` (pass)
* private server given but no user prefix i.e. `127.0.0.1:5000/function1` (pass)
* private server given and a prefix i.e. `127.0.0.1:5000/myfunctions/function1` (pass)

Test all of above with CLI arguments and a YAML file.

1. `faas new --lang go test1`
2. `faas push -f test1.yml`

BurtonR:
Signed-off-by: Burton Rheutan <rheutan7@gmail.com>

Added a check before calling `docker push` to ensure that the `image` field in the `.yml` file contains a repository by looking for `/`.

## Description

Added new function to `push.go` to validate that the image contains a repository by looking for `/` in the image string

Added output message with list of failed function names with message to indicate the failure (and an example of the correct usage)

I also moved the `"Pushing function1 done"` message into the else block that calls the `docker push` command. I thought it confusing to have that message show when the `push` never actually happened...

## Motivation and Context

Closes #329 

## How Has This Been Tested?

Built the CLI locally and ran `./faas-cli push -f test.yml` with 3 functions defined. One with a repository `burtonr/movies:0.2`, one with `skip_build: true`, and one without a repository `ecs-cli:latest`. Verified that the push was blocked (in both synchronous and `--parallel`). Updated the `.yml` file to ensure both functions without `skip_build` had the repository in the image. Ran `./faas-cli push -f test.yml` again, and ensured that the push happened. See example `.yml` below. 

## Types of changes

## Checklist:

I did not add any tests as I'm not very familiar with go tests and there were no existing tests for the `push.go` file. I don't feel as I have the skills needed to add a new test file and avoid running the `docker push` command for the test.

Here is a sample .yml file I used while testing:
```yaml
provider:
  name: faas
  gateway: http://localhost:8080

functions:
  movies:
    lang: node
    handler: ./movies
    image: burtonr/movies:0.1

  stronghash:
    skip_build: true
    image: alpine:latest
    fprocess: "sha512sum"

  ecs-cli:
    lang: Dockerfile
    handler: ./ecs-cli
    image: ecs-cli:latest
```

and the output (notice the "stronghash" function has `skip_build: true` and `ecs-cli` has no repository)
```bash
[0] % ./faas-cli push -f image-test.yml --parallel 3

Unable to push one or more of your functions to Docker Hub
 - ecs-cli

You must provide a username or registry prefix to the Function's image such as user1/function1
```

### DIFF
--- a/commands/push.go
+++ b/commands/push.go
@@ -111,11 +111,8 @@ func pushStack(services *stack.Services, queueDepth int) {
 func validateImages(services *stack.Services) string {
 	errMsg := make([]string, 0)
 	for name, function := range services.Functions {
-		if function.SkipBuild {
-			continue
-		}
 
-		if !strings.Contains(function.Image, `/`) {
+		if !function.SkipBuild && !strings.Contains(function.Image, `/`) {
 			errMsg = append(errMsg, fmt.Sprintf(" - %s", name))
 		}
 	}
--- a/commands/push.go
+++ b/commands/push.go
@@ -5,7 +5,6 @@ package commands
 
 import (
 	"fmt"
-	"regexp"
 	"strings"
 	"sync"
 
@@ -55,7 +54,11 @@ func runPush(cmd *cobra.Command, args []string) error {
 	if len(services.Functions) > 0 {
 		invalidImages := validateImages(&services)
 		if len(invalidImages) > 0 {
-			return fmt.Errorf("Unable to push one or more of your functions to Docker Hub\n" + invalidImages + "\n\nYou must provide a username or registry prefix to the Function's image such as user1/function1.\nIf you need a Docker Hub account, you can sign up here: https://hub.docker.com")
+			return fmt.Errorf(`
+Unable to push one or more of your functions to Docker Hub
+` + invalidImages + `
+
+You must provide a username or registry prefix to the Function's image such as user1/function1`)
 		}
 
 		pushStack(&services, parallel)
@@ -111,11 +114,8 @@ func validateImages(services *stack.Services) string {
 		if function.SkipBuild {
 			continue
 		}
-		ip := "(\\b\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\b)(:\\d+)?"
-		re := regexp.MustCompile("([a-z].+)\\/|" + ip + "\\/")
-		ma := re.FindAllString(function.Image, 1)
 
-		if len(ma) < 1 {
+		if !strings.Contains(function.Image, `/`) {
 			errMsg = append(errMsg, fmt.Sprintf(" - %s", name))
 		}
 	}
--- a/commands/push.go
+++ b/commands/push.go
@@ -6,6 +6,7 @@ package commands
 import (
 	"fmt"
 	"regexp"
+	"strings"
 	"sync"
 
 	"github.com/morikuni/aec"
@@ -52,6 +53,11 @@ func runPush(cmd *cobra.Command, args []string) error {
 	}
 
 	if len(services.Functions) > 0 {
+		invalidImages := validateImages(&services)
+		if len(invalidImages) > 0 {
+			return fmt.Errorf("Unable to push one or more of your functions to Docker Hub\n" + invalidImages + "\n\nYou must provide a username or registry prefix to the Function's image such as user1/function1.\nIf you need a Docker Hub account, you can sign up here: https://hub.docker.com")
+		}
+
 		pushStack(&services, parallel)
 	} else {
 		return fmt.Errorf("you must supply a valid YAML file")
@@ -75,8 +81,8 @@ func pushStack(services *stack.Services, queueDepth int) {
 				fmt.Printf(aec.YellowF.Apply("[%d] > Pushing %s.\n"), index, function.Name)
 				if len(function.Image) == 0 {
 					fmt.Println("Please provide a valid Image value in the YAML file.")
-				} else if !validImageString(function.Image) {
-					fmt.Printf("Unable to push %s. You must provide a username or registry prefix such as user1/function.\nIf you need a Docker Hub account, you can sign up here: https://hub.docker.com\n", function.Name)
+				} else if function.SkipBuild {
+					fmt.Printf("Skipping %s\n", function.Name)
 				} else {
 					pushImage(function.Image)
 					fmt.Printf(aec.YellowF.Apply("[%d] < Pushing %s done.\n"), index, function.Name)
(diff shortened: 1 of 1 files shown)


alexellis:
Hi @BurtonR - for this to be effective I think we need to block the entire flow if one or more of the functions fails the validation. In this example if we had half a dozen images the user may miss the error message which appears mid-flow - especially if in a parallel build.

Thoughts?

Alex

BurtonR:
I had considered that, but what happens when you have a function like the one I have in the example where it's just a base image? Would you have to comment that out in order for the push to happen?

I could update it to have the text in red... That would at least stand out and make it more obvious that something not-so-good happened. 

Or, another option could be to keep the error messages and display them all at the end, either just before, or just after the `worker done` message.

alexellis:
@BurtonR do you mean `stronghash`? The work item says that for `skip_build` you should bypass the validation.

This should be a completely blocking pre-validation.

```
$ faas-cli push
We were unable to push one or more of your functions to the Docker Hub.
- movie-lister
- ec2-autoscaler
- faas-cli

You must provide a username or registry prefix to the Function's image such as user1/function1.
If you need a Docker Hub account, you can sign up here: https://hub.docker.com
```

BurtonR:
Something like this?

Blocks the push completely. Also, note the "Skipping stronghash" on checking the `skip_build` property

screenshot from 2018-02-19 11-45-28

rgee0:
@BurtonR lets also take the opportunity in the `unable` message to explain that the user will need to be logged in to docker hub.  

It may help head off the subsequent question, once the user has added their repo/prefix to the yaml

BurtonR:
@rgee0 

How does this look?

```
Unable to push one or more of your functions to Docker Hub
 - movies
 - ecs-cli

You must provide a username or registry prefix to the Function's image such as user1/function1.
Are you logged in to Docker Hub via "docker login"?
If you need a Docker Hub account, you can sign up here: https://hub.docker.com
```

rgee0:
Keep in mind that it won’t always be a public push.  Are we picking up where the user is targeting with the push and tailoring?

I’d be tempted to merge the two Docker hub messages.  Something like:

`Also, ensure you are logged in to Docker hub - need an account? Create one here...`

alexellis:
I think our messages are quite succinct now so I think it makes sense to keep them consistent. How about we leave the login info until later if we really need it. Less is more in that way.

BurtonR:
So, what _should_ the message say? 

To be succinct, perhaps we should just leave it at the first line:
`You must provide a username or registry prefix to the function's image such as user1/function1`

The "sign up here" bit does feel pointed at beginners. Perhaps that would be better off in the docs only?

alexellis:
OpenFaaS has a core value of *developer-first* - and many of our developers may be using Docker for the first time by using OpenFaaS. 

If you feel leaving off `If you need a Docker Hub account, you can sign up here: https://hub.docker.com` is the right approach then let's do it. If we need to add it back at a later date then we can 👍

commands/push.go
```diff
@@ -95,3 +104,20 @@ func pushStack(services *stack.Services, queueDepth int) {
...
 
 }
+
+func validateImages(services *stack.Services) string {
+	errMsg := make([]string, 0)
+	for name, function := range services.Functions {
+		if function.SkipBuild {
+			continue
+		}
+		ip := "(\\b\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\b)(:\\d+)?"
```
alexellis: My gut feeling is that this is too complicated. Why are we using a `regex` when a string/contains would probably suffice?

```
if there's a `/` in the string.. then go for it?
```

BurtonR: That sounds like a logical check... I was obviously over thinking this!

BurtonR:
I've updated the PR description as I've updated the check to look for `/` and updated the error message to a more succinct version (without the login or "sign up" text)

alexellis:
Thank you for updating the code @BurtonR - here's my next question.. what if someone applies the `--filter` flag in your example?

```
./faas-cli push -f image-test.yml --filter=movies
```

This should pass / push. Please could you confirm that it does?

commands/push.go
```diff
@@ -95,3 +107,17 @@ func pushStack(services *stack.Services, queueDepth int) {
...
 }
+
+func validateImages(services *stack.Services) string {
+	errMsg := make([]string, 0)
+	for name, function := range services.Functions {
+		if function.SkipBuild {
+			continue
+		}
+
+		if !strings.Contains(function.Image, `/`) {
```
alexellis: how about `&& !function.SkipBuild` ?

alexellis:
We're looking good now - the code is concise too. Would you be able to take a look at the testing done by @ericstoekl and @itscaro to see if you can cover the positive and negative cases with tests? If you would like some help feel free to ping in #contributors on Slack.

BurtonR:
@alexellis I've verified that the `--filter` option does function as expected.

On a `.yml` file in the PR description, where `movies` function _does_ have a repository, but the others do not, filtering by `--filter=movies` pushes the `movies` function successfully (does not check the other functions images).

alexellis:
💯

commands/push.go
```diff
@@ -95,3 +107,14 @@ func pushStack(services *stack.Services, queueDepth int) {
 	wg.Wait()
 
 }
+
+func validateImages(services *stack.Services) string {
+	errMsg := make([]string, 0)
+	for name, function := range services.Functions {
+
+		if !function.SkipBuild && !strings.Contains(function.Image, `/`) {
```
johnmccabe: Do you need to handle the case where the image only contains the `/`, for example `/imgfn`?

alexellis: I don't think so? In that instance Docker will still error.

alexellis:
Please see the branch test_push_validation for an initial approach on unit testing the behaviour.

commands/push.go
alexellis: Thoughts?

BurtonR: This is the error message that docker provides

```
invalid reference format
2018/02/22 09:21:09 ERROR - Could not execute command: [docker push /movies:0.1]
```

I think that's pretty clear, but do you think the `faas-cli` should provide something?

johnmccabe: I just don't like the `invalid reference format` that results from the docker stderr when we are in a position to absorb it.

Its by no means a blocker however, just a personal preference for actionable output.

Collinshbk:
I am having the same issue. But all these back and forth but no clear solution or examples

alexellis:
The clear solution is that you need to log in.

@Collinshbk every eBook, training course and piece of official documentation explains that.

What solution are you wanting?

PR #357, opened 2018-03-27
Prevent duplicate function when using append flag

### LINKED ISSUE #354: Suggestion: block appending of the same function more than once
Summary: Using "append" with the "new" command allows duplicate function names

Using the `--append` option appends function to yaml file regardless of if the function name is already taken, creating ambiguous yaml file.

## Expected Behaviour
It should return an error if the function name would result in a duplicate.

## Current Behaviour
it will continue to append.

Example when running `faas-cli new yee --lang ruby -f functions.yml` more than once.

```yml
functions:
  yee:
    lang: ruby
    handler: ./yee
    image: yee
  yee:
    lang: ruby
    handler: ./yee
    image: yee
```

## Your Environment
- Docker `17.12.0-ce`
- Using Docker Swarm
- MacOS

BurtonR:
According to YAML spec, YAML keys MUST be unique
http://www.yaml.org/spec/1.2/spec.html section 1.3
This change blocks the new command from creating duplicates when using the append flag
The block happens before the function is created to prevent having a conflict between
the function generated and the function defined in the yaml file

Signed-off-by: Burton Rheutan <rheutan7@gmail.com>

## Description

Added check based on function name to the `new --append` command + flag
Moved the append file read function higher in the order to be able to block before the function gets created. 

## Motivation and Context

[Specs](http://www.yaml.org/spec/1.2/spec.html) say YAML keys MUST be unique (section 1.3)

Resolves #354 

## How Has This Been Tested?

Added unit test to the `new_function_test.go` that creates a yml file with a sample function definition and runs the `new` command with the `--append` flag. 
Test fails if the expected error message is not returned.

## Types of changes

## Checklist:

### DIFF
--- a/commands/new_function.go
+++ b/commands/new_function.go
@@ -99,6 +99,12 @@ func runNewFunction(cmd *cobra.Command, args []string) error {
 		if _, statErr := os.Stat(appendFile); statErr != nil {
 			return fmt.Errorf("unable to find file: %s - %s", appendFile, statErr.Error())
 		}
+
+		duplicateError := duplicateFunctionName(functionName, appendFile)
+
+		if duplicateError != nil {
+			return duplicateError
+		}
 	}
 
 	if _, err := os.Stat(functionName); err == nil {
@@ -155,6 +161,7 @@ functions:
 		if readErr != nil {
 			fmt.Printf("unable to read %s to append, %s", appendFile, readErr)
 		}
+
 		buffer := string(originalBytes) + stackYaml
 
 		stackWriteErr = ioutil.WriteFile(appendFile, []byte(buffer), 0600)
@@ -183,6 +190,27 @@ func printAvailableTemplates(availableTemplates []string) string {
 		result += fmt.Sprintf("- %s\n", template)
 	}
 	return result
+}
+
+func duplicateFunctionName(functionName string, appendFile string) error {
+	fileBytes, readErr := ioutil.ReadFile(appendFile)
+	if readErr != nil {
+		return fmt.Errorf("unable to read %s to append, %s", appendFile, readErr)
+	}
+
+	services, parseErr := stack.ParseYAMLData(fileBytes, "", "")
+
+	if parseErr != nil {
+		return fmt.Errorf("Error parsing %s yml file", appendFile)
+	}
+
+	if _, ok := services.Functions[functionName]; ok {
+		return fmt.Errorf(`
+Function %s already exists in %s file. 
+Cannot have duplicate function names in same yml file`, functionName, appendFile)
+	}
+
+	return nil
 }
 
 // StrSort sort strings
--- a/commands/new_function_test.go
+++ b/commands/new_function_test.go
@@ -5,6 +5,7 @@ package commands
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"reflect"
 	"regexp"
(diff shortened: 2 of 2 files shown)


commands/new_function_test.go
```diff
@@ -208,6 +210,60 @@ func Test_languageNotExists(t *testing.T) {
...
+func Test_duplicateFunctionName(t *testing.T) {
+	templatePullLocalTemplateRepo(t)
+	defer tearDownFetchTemplates(t)
+	defer tearDownNewFunction(t)
+
+	const functionName = "sampleFunc"
+	const functionLang = "ruby"
+
+	defer func() {
+		os.Remove(functionName + ".yml")
```
ericstoekl: May want to check if file exists, and if so remove it.

commands/new_function.go
```diff
@@ -99,6 +102,24 @@ func runNewFunction(cmd *cobra.Command, args []string) error {
...
+			fmt.Printf("unable to read %s to append, %s", appendFile, readErr)
+		}
+
+		services, parseErr := stack.ParseYAMLData(originalBytes, "", "")
+
+		if parseErr != nil {
+			fmt.Printf("Error parsing %s yml file", appendFile)
+		}
+
+		for function := range services.Functions {
```
ericstoekl: `services.Functions` is a map object. You don't need to loop through it to search for the function:

```		if _, ok := services.Functions[functionName]; ok {
			return fmt.Errorf(`Function %s already exists in %s file. 
 Cannot have duplicate function names in same yml file`, functionName, appendFile)
		}```

**Comment by BurtonR** [comment]
@ericstoekl Thanks for the review! Good catch on looping through the map! I've updated this with your suggestions
@alexellis Please have a look. This should be ready to go if you're good with it

File: commands/new_function.go [hunk 3]
```diff
@@ -99,6 +102,22 @@ func runNewFunction(cmd *cobra.Command, args []string) error {
 		if _, statErr := os.Stat(appendFile); statErr != nil {
 			return fmt.Errorf("unable to find file: %s - %s", appendFile, statErr.Error())
 		}
+
+		originalBytes, readErr = ioutil.ReadFile(appendFile)
+		if readErr != nil {
+			fmt.Printf("unable to read %s to append, %s", appendFile, readErr)
+		}
+
+		services, parseErr := stack.ParseYAMLData(originalBytes, "", "")
```
alexellis: I'd like to see these 3-4 statements which pertain to duplicate-checking extracted into a new local/private function.

commands/new_function_test.go
```diff
@@ -208,6 +210,65 @@ func Test_languageNotExists(t *testing.T) {
 	}
 }
 
+func Test_duplicateFunctionName(t *testing.T) {
```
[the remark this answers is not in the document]
BurtonR: I agree, it's not a best practice, but unsure of how to go about that in a "pure" unit test style.

The trick is that the `runNewFunction` uses the `ioutil.ReadFile()` function to get the data from the file it is attempting to append to. I'm not sure if there's a stub, or some other utility, that would allow us to fake that process without actually writing to the file system.

BurtonR:
I've updated the process to use a separate private function to check for duplicates. This new function returns an error that, if present, will be returned from the `runNewFunction`

Also, see my comment about changing the test to not write to the file system

alexellis:
The new style looks easier to maintain and clearer too. I'll merge it.

Thanks for taking this on.

PR #373, opened 2018-04-09
[WIP] use `--yaml` flag to both specify and append yaml

### LINKED ISSUE #371: "faas-cli new" should accept option to use default stack.yml name
When typing in `faas-cli new --lang node html-project` for instance you get a YAML file named html-project.

We introduced "stack.yml" as a default file-name, but have no way of generating a new function with the stack name of stack.yml. You have to rename it afterwards.

What ideas do you have for allowing this without getting to verbose in the CLI?

```
faas new --lang python tester --stack=stack.yml
```

We do have an append flag - perhaps we could look at collapsing the `--stack` and `--append` flag into one option?

s8sg:
Signed-off-by: Swarvanu Sengupta <swarvanusg@gmail.com>

resolves #371

## Description
`--append` flag is discarded
`--yaml` | `-f` flag is being used to specify the new function yaml
                 or to append new function to existing yaml file

## Motivation and Context
it fixes an open issue: [issue-371](https://github.com/openfaas/faas-cli/issues/371)

## How Has This Been Tested?
it is tested by running existing [test cases](https://github.com/openfaas/faas-cli/compare/master...s8sg:issue-371?expand=1#diff-7eed3f568a73108a10df6a5d68a3b2ed)
Manually tested as: 
```bash
$ faas-cli new my_func --lang go
...
$ ls
my_func  my_func.yml  template
$ faas-cli new my_func_2 --lang dockerfile --yaml my_func.yml
...
$ ls
my_func  my_func_2  my_func.yml  template
$ faas-cli new my_func_3 --lang dockerfile --yaml my_func_3.yml
...
$ ls
my_func  my_func_2  my_func_3  my_func_3.yml  my_func.yml  template
$ cat my_func.yml
provider:
  name: faas

functions:
  my_func:
    lang: go
    handler: ./my_func
    image: my_func
  my_func_2:
    lang: dockerfile
    handler: ./my_func_2
    image: my_func_2
$ cat my_func_3.yml
provider:
  name: faas

functions:
  my_func_3:
    lang: dockerfile
    handler: ./my_func_3
    image: my_func_3
```

## Types of changes

## Checklist:


rgee0:
@s8sg thanks for this.  We’ll get it reviewed shortly.

In the meantime, can you add a reference to the description that follows [this guidance](https://help.github.com/articles/closing-issues-using-keywords/), please?  It just helps us keep the repos in check.  You’re almost there but im not sure how its referenced will cause GH to trigger.

Thank you.

alexellis:
Sorry to do this. Just had a thought. We already use a flag for this elsewhere called --yaml or -f. We should be consistent.

alexellis:
This will go on my list for review next week unless someone can get to it sooner?

commands/new_function.go
```diff
@@ -27,20 +26,20 @@ func init() {
...
 // newFunctionCmd displays newFunction information
 var newFunctionCmd = &cobra.Command{
-	Use:   "new FUNCTION_NAME --lang=FUNCTION_LANGUAGE [--gateway=http://domain:port] | --list | --append=STACK_FILE)",
+	Use:   "new FUNCTION_NAME --lang=FUNCTION_LANGUAGE [--gateway=http://domain:port] | --list | --yaml=YAML_FILE)",
 	Short: "Create a new template in the current folder with the name given as name",
-	Long: `The new command creates a new function based upon hello-world in the given
-language or type in --list for a list of languages available.`,
+	Long: `The new command creates a new function based upon hello-world in the given language
+or type in --list for a list of languages available.  YAML file can be specified by
+--yaml flag to generate new file otherwise it will append function to existing file`,
```
rgee0: I think this needs to more clearly express what will happen.  As I understand it if `stack.yml` exists then it will be appended to there unless `--yaml` is added and in that case if the file exists it will be appended to, otherwise a new file will result.

commands/new_function.go
```diff
@@ -90,20 +89,24 @@ func runNewFunction(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("%s is unavailable or not supported", language)
 	}
 
-	appendMode := len(appendFile) > 0
-	if appendMode {
-		if (strings.HasSuffix(appendFile, ".yml") || strings.HasSuffix(appendFile, ".yaml")) == false {
-			return fmt.Errorf("when appending to a stack the suffix should be .yml or .yaml")
+	appendMode := false
+	fileProvided := len(yamlFile) > 0
```
rgee0: This can evaluate to true if `--yaml` is absent but `stack.yml` exists.  Is that the intended behaviour?

commands/new_function.go
```diff
@@ -112,9 +115,9 @@ func runNewFunction(cmd *cobra.Command, args []string) error {
 	}
 
 	if err := os.Mkdir(functionName, 0700); err == nil {
-		fmt.Printf("Folder: %s created.\n", functionName)
+		fmt.Printf("Folder: %s is created.\n", functionName)
 	} else {
-		return fmt.Errorf("folder: could not create %s : %s", functionName, err)
+		return fmt.Errorf("Folder: %s failed to create: %s", functionName, err)
```
rgee0: Were we following a lowercase error messages convention?

commands/new_function.go
```diff
@@ -192,22 +199,22 @@ func printAvailableTemplates(availableTemplates []string) string {
 	return result
 }
 
-func duplicateFunctionName(functionName string, appendFile string) error {
-	fileBytes, readErr := ioutil.ReadFile(appendFile)
+func duplicateFunctionName(functionName string, yamlFile string) error {
```
rgee0: I know this was pre-existing but it could be better named.  It suggests that its going to produce a duplicate of something, when it is actually testing uniqueness.  It also returns an error when it finds a duplicate, which is counter-intuitive.

commands/new_function.go
alexellis: They should start lowercase yes

alexellis:
@rgee0 would you be able to put together some input/output scenarios / text snippets that show how this should work if a `stack.yml` file is present in the folder.

alexellis:
Thanks for the additional changes. @rgee0 can you provide the test-cases input/output for @s8sg so we can get this PR merged and closed out?

commands/new_function.go
```diff
@@ -90,20 +90,28 @@ func runNewFunction(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("%s is unavailable or not supported", language)
 	}
 
-	appendMode := len(appendFile) > 0
-	if appendMode {
-		if (strings.HasSuffix(appendFile, ".yml") || strings.HasSuffix(appendFile, ".yaml")) == false {
-			return fmt.Errorf("when appending to a stack the suffix should be .yml or .yaml")
+	appendMode := false
+	if len(yamlFile) > 0 {
```
rgee0: Doesn't this need to be compared with `defaultYAML` to ensure that when a `--yaml` hasn't been passed we aren't writing to `stack.yml` if it exists?

rgee0:
I haven't tested the latest commit yet, but I'm still not clear from reading the code how the value of `defaultYAML` is being over-ridden.  I think maybe the value of `defaultYAML` being `stack.yml` has possibly confused the issue.

I'll be interested to see the results from the testing scenarios.

It might be more fruitful to pick this up on Slack if we're about at the same time.

s8sg:
The Default Yaml file if not provide  it is same as it was in original code. That is <func_name>.yaml.  Currently there is no default value of yamlFile value like stack.yaml (I’m not sure if its a intended behavior). If yaml file is specified it will use the provided file name.  In either case if provided or not, if the file exists it will append to existing file

rgee0:
Having caught up on Slack I think we're heading in the right direction.  To summarise our discussion for the benefit of the wider audience:

This snippet is affecting the change because its causing `yamlFile` to be set to `defaultYAML` when the user doesn't pass in `--yaml`, which means that rather then getting `fn1.yml` the config will be appended to `stack.yml` should that file exist.

https://github.com/openfaas/faas-cli/blob/4e72ea6ef77a3375904a8b84e257efc1663607bc/commands/faas.go#L78-L83

There needs to be means in `new_function.go` to check whether the `defaultYAML` is being used and over-ride with `functionName`.  One way might be to check whether `—yaml` has been passed and if it hasn’t && `yamlFile==defaultYAML` then use the function name.  There was a concern that this may not be viable as `yamlFile` has already been overwritten, and so the idea of setting a flag in `checkAndSetDefaultYaml()` was suggested.

Whatever the means, we need to ensure we also accommodate cases where `stack.yml` has been passed via `—yaml`.

I look forward to seeing how @s8sg solves this.

s8sg:
On the updated PR as per previous discussion we are using the flag `usingDefaultYaml ` to identify if the `yamlFile` is default value or passed value

Previously we only checked if `stack.yml` is present, which would make the `usingDefaultYaml` `true` even if `-f stack.yml` is provided
 
To avoid that scenario `checkAndSetDefaultYaml ` update `yamlFile` to `defaultYAML` only if `yamlFile` is not specified using `-f`
```go
func checkAndSetDefaultYaml() {
	_, err := stat(defaultYAML)
	// Check if there is a default yaml file and set it
	if len(yamlFile) == 0 && err == nil {
		yamlFile = defaultYAML
		usingDefaultYaml = true
	}
}
``` 
Although previously the `checkAndSetDefaultYaml()` was being executed before the execution of cobra `fass command`. Although we didn't have any `yamlFile` value parsed by then. 

To have parsed `yamlFile` value and to execute before each command we are using the cobra `PersistentPreRun` functionality
```go
var faasCmd = &cobra.Command{
	Use:   "faas-cli",
	Short: "Manage your OpenFaaS functions from the command line",
	Long: `
Manage your OpenFaaS functions from the command line`,
	PersistentPreRun: func(cmd *cobra.Command, args []string) {
		checkAndSetDefaultYaml()
	},
	Run: runFaas,
}
```

rgee0:
Are the test scenarios provided above resulting in the intended behaviour?

rgee0:
Derek edit title: [WIP] use `--yaml` flag to both specify and append yaml

s8sg:
@rgee0 Yes it is working as it the intended behaviour is mentioned, can I rebase the commit for marge? There are some conflicts I'll resolve them same at one go with the rebase commit

commands/new_function.go
```diff
@@ -92,6 +92,10 @@ func runNewFunction(cmd *cobra.Command, args []string) error {
 
 	appendMode := false
 	if len(yamlFile) > 0 {
+		if (yamlFile == defaultYAML) && usingDefaultYaml {
```
rgee0: I think this can be simplified.   I wonder if `if (len(yamlFile) > 0)  && (usingDefaultYaml==false) {` would remove the need for the inner `if`?

rgee0:
I've left a comment in the code

We'll need to rebase with upstream master to resolve the conflicts and I'd like to test wider than this feature due to the changes to `faas.go`.

commands/new_function.go
s8sg: your right. if `usingDefaultYaml` is true it means always `yamlFile == defaultYAML` true. I will remove the nested check and add it as you suggested

alexellis:
@s8sg there's two issues with this PR:

- CI is failing
- at least one commit is not signed off

Please can you address this ASAP?

Hint:

I think this is the cause of the issue:

> Merge branch 'master' into issue-371

Please do not type in `git pull` or `git merge` - use `git merge --rebase` or `git rebase origin/master` instead.

Alex

ericstoekl:
Continuation of the work in this PR is here: https://github.com/openfaas/faas-cli/pull/434

PR #390, opened 2018-04-24
[Needs testing] Refactor deploy proxy to only call deployment once

### LINKED ISSUE #387: 0.6.6 spams two messages about HTTPS when deploying
## Expected Behaviour

A single warning about HTTPS not being used on a remote gateway - or actually nothing for loopback addresses of `127.0.0.1` and `localhost` cc @stefanprodan 

## Current Behaviour

Since the change via @BurtonR @nenadilic84 to make registry auth strings get passed on every deployment.

```
go build && ./faas-cli deploy -f ~/go/src/github.com/open
faas/faas/sample-functions/dockerhub-stats.yml
Deploying: dockerhub-stats.
WARNING! Communication is not secure, please consider using HTTPS. Letsencryp
t.org offers free SSL/TLS certificates.
WARNING! Communication is not secure, please consider using HTTPS. Letsencryp
t.org offers free SSL/TLS certificates.

Deployed. 200 OK.URL: http://127.0.0.1:8080/function/dockerhub-stats
```

## Possible Solution

We need to only print the message once - not twice. It could be since it was a rolling deployment?

Registry Auth behavior is giving a bit more of an obtrusive experience. If the majority of users will be on Kubernetes then should we even send the registry auth string on every deployment?

See #386 

## Your Environment

Provided in #386

BurtonR:
This change is to prevent the deployment call from happening twice
if a function does not yet exist. 

## Description

When calling the `deploy` function in the proxy, we first call the `ListFunctions` command to see what's deployed to the gateway to know whether or not it should be an update. 

If the update flag is set to true, but the function does not yet exist, we change the update flag to false to prevent errors. 

Also, if the update flag is set to false and the function _does_ exist, we display an error and exit. This prevents the current behavior that the request returns an error:
```
Unexpected status: 400, message: Deployment error: Error response from daemon: rpc error: code = Unknown desc = name conflicts with an existing object
```

## Motivation and Context

This was causing warning messages to be displayed twice and resulted in a poor developer experience.
Resolves #387 

## How Has This Been Tested?

Quick first round testing so. Will add more soon...
* Deploy new function (from blank template)
* Deploy same function again
* Remove that function
* Deploy same function again (essentially a new function)

~Here's where the `[WIP]` flag comes in. I'll need to run the unit tests. Update them, and potentially add a new one to cover this new functionality. Just putting this here to get some feedback while I work through the tests.~

I've added unit tests to cover my changes, and also had to update one of the existing ones to account for the slight change in behavior.
 
## Types of changes

## Checklist:

### DIFF
--- a/proxy/deploy.go
+++ b/proxy/deploy.go
@@ -22,25 +22,35 @@ type FunctionResourceRequest struct {
 	Requests *stack.FunctionResources
 }
 
+// DeployFunction will first check if the function exists, and if it does, or the update bool is set to true, will perform a rolling update
+// otherwise, the function will be deployed as new
 func DeployFunction(fprocess string, gateway string, functionName string, image string,
 	registryAuth string, language string, replace bool, envVars map[string]string,
 	network string, constraints []string, update bool, secrets []string,
 	labels map[string]string, functionResourceRequest1 FunctionResourceRequest) {
 
 	rollingUpdateInfo := fmt.Sprintf("Function %s already exists, attempting rolling-update.", functionName)
-	statusCode, deployOutput := Deploy(fprocess, gateway, functionName, image, registryAuth, language, replace, envVars, network, constraints, update, secrets, labels, functionResourceRequest1)
 
-	if update == true && statusCode == http.StatusNotFound {
-		// Re-run the function with update=false
-		_, deployOutput = Deploy(fprocess, gateway, functionName, image, registryAuth, language, replace, envVars, network, constraints, false, secrets, labels, functionResourceRequest1)
-	} else if statusCode == http.StatusOK {
+	existing := functionExists(gateway, functionName)
+
+	if update {
+		update = existing
+	} else if !update && existing {
+		fmt.Printf("Function %s already exists, you must either remove it first, or update it", functionName)
+		return
+	}
+
+	_, deployOutput := deploy(fprocess, gateway, functionName, image, registryAuth, language, replace, envVars, network, constraints, update, secrets, labels, functionResourceRequest1)
+
+	if update || existing {
 		fmt.Println(rollingUpdateInfo)
 	}
+
 	fmt.Println()
 	fmt.Println(deployOutput)
 }
 
-func Deploy(fprocess string, gateway string, functionName string, image string,
+func deploy(fprocess string, gateway string, functionName string, image string,
 	registryAuth string, language string, replace bool, envVars map[string]string,
 	network string, constraints []string, update bool, secrets []string,
 	labels map[string]string, functionResourceRequest1 FunctionResourceRequest) (int, string) {
@@ -157,4 +167,22 @@ func Deploy(fprocess string, gateway string, functionName string, image string,
 	}
 
 	return res.StatusCode, deployOutput
+}
+
+func functionExists(gateway string, funcName string) bool {
+
+	funcs, err := ListFunctions(gateway)
+
+	if err != nil {
+		fmt.Println(err.Error())
+		return false
+	}
+
+	for i := range funcs {
+		if funcs[i].Name == funcName {
+			return true
+		}
+	}
+
+	return false
 }
--- a/proxy/deploy.go
+++ b/proxy/deploy.go
@@ -36,7 +36,7 @@ func DeployFunction(fprocess string, gateway string, functionName string, image 
 	if update {
 		update = existing
 	} else if !update && existing {
-		fmt.Printf("Function %s already exists, you must either remove it first, or update it", functionName)
+		fmt.Printf("Function %s already exists, you must either remove it first, or update it\n", functionName)
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


alexellis:
CI appears to be failing at the moment.

BurtonR:
Pretty sure I broke all the tests around the `deploy` command by adding the `ListFunctions` call to the deploy. 
Here's the relevant bit from the CI logs:
```
--- FAIL: Test_RunDeployProxyTests (0.00s)
    --- FAIL: Test_RunDeployProxyTests/200_Deploy (0.00s)
    	deploy_test.go:53: Output not matched: cannot parse result from OpenFaaS on URL: http://127.0.0.1:33515
    		unexpected end of JSON input
```
I expected this, just haven't got into the tests to update them just yet. Rest assured, the tests will be updated and passing before I remove the "[WIP]" tag from the title

BurtonR:
@alexellis  and @rgee0 I've added unit tests for the change I made. All the current, and new tests are passing, but I'd appreciate a review to make sure everything looks like it should.

Please let me know if you see anything that should be changed. Thanks!

alexellis:
Please can you ask @ericstoekl to help verify as Eric wrote the original code for the two-stage deployment?

proxy/deploy.go
```diff
@@ -22,25 +22,35 @@ type FunctionResourceRequest struct {
...
-	if update == true && statusCode == http.StatusNotFound {
-		// Re-run the function with update=false
-		_, deployOutput = Deploy(fprocess, gateway, functionName, image, registryAuth, language, replace, envVars, network, constraints, false, secrets, labels, functionResourceRequest1)
-	} else if statusCode == http.StatusOK {
+	existing := functionExists(gateway, functionName)
+
+	if update {
+		update = existing
+	} else if !update && existing {
+		fmt.Printf("Function %s already exists, you must either remove it first, or update it", functionName)
```
ericstoekl: Could use a `\n` at the end of this printf message

BurtonR:
Thanks for checking this out @ericstoekl! 
I updated that error message to include a new line at the end. Good call! :smile:

proxy/deploy.go
```diff
@@ -22,25 +22,35 @@ type FunctionResourceRequest struct {
...
 	rollingUpdateInfo := fmt.Sprintf("Function %s already exists, attempting rolling-update.", functionName)
-	statusCode, deployOutput := Deploy(fprocess, gateway, functionName, image, registryAuth, language, replace, envVars, network, constraints, update, secrets, labels, functionResourceRequest1)
 
-	if update == true && statusCode == http.StatusNotFound {
-		// Re-run the function with update=false
-		_, deployOutput = Deploy(fprocess, gateway, functionName, image, registryAuth, language, replace, envVars, network, constraints, false, secrets, labels, functionResourceRequest1)
-	} else if statusCode == http.StatusOK {
+	existing := functionExists(gateway, functionName)
+
+	if update {
```
ericstoekl: When `update=false` and `replace=true` we would want to do a force update, rather than rejecting the command?

BurtonR: Good point Eric! Here, I'm only focusing on the `update` flag since update and replace are effectively the same thing. So, if you don't want to update the function (by setting `update=false`), but the function exists, I want to make sure that you really do want to update it... The default value is `true` for the `update` flag, so if you, the user, is explicitly setting the flag to false, we don't want to update/replace that function unexpectedly.

What do you guys think @alexellis, @johnmccabe, and @rgee0?

BurtonR:
The recently added unit tests: `store_deploy_test.go` don't match up with the changes made here to the `deploy` function. So, I'll need to rebase and update the tests to make sure they pass with this new functionality.

proxy/deploy.go
rgee0: This is about refactoring to address what is essentially a bug where a message is displayed twice in particular scenarios.  The refactor should aim to address the issue without changing the CLI behaviour beyond not outputting the message twice.

alexellis:
I don't understand why CI is broken. @BurtonR can you look into this?

alexellis:
Everyone should be doing a local build (including running tests) before pushing changes into a PR.

alexellis:
Fixed via https://github.com/openfaas/faas-cli/commit/6532f3c66967b3a872208a29b1b60a873b7d2490

If you would like to keep working on this behavior change then we need to agree between us about changing the behavior so there is a 1x GET & 1x POST instead of 1x POST (new) or 2x POST (existing).

alexellis:
@stefanprodan may have an opinion on this?

BurtonR:
@alexellis  please see my comment from last night (just before your comment an hour ago) where I explained that the CI is now failing due to the newly added tests. The tests only started to fail yesterday on the merges from #405 . The new tests don't account for this new functionality with the additional call to `ListFunctions()`

If you've made a fix for the message displaying twice already, then there is no need for this PR and it can be closed.

alexellis:
Thank you @BurtonR

alexellis:
Derek lock

Issue #394, opened 2018-04-26
Make sending registry auth optional for "store deploy"


alexellis:
## Expected Behaviour

No errors when deploying figlet from the store.

## Current Behaviour

I get two warnings then an error (I can't get past this issue)

```
$ faas store deploy figlet
WARNING! Communication is not secure, please consider using HTTPS. Letsencrypt.org offers free SSL/TLS certificates.
WARNING! Communication is not secure, please consider using HTTPS. Letsencrypt.org offers free SSL/TLS certificates.
Unexpected status: 400, message: Invalid registry
```

https://github.com/openfaas/faas-cli/pull/388

## Possible Solution

Fix is in https://github.com/openfaas/faas-cli/pull/388 - replicate the work for the store deploy command.

## Steps to Reproduce (for bugs)

1. Get CLI 0.6.6, .7 or .8 and type in `faas store deploy figlet`

## Your Environment

Docker for Mac
Local faas on Swarm

PR #396, opened 2018-04-26
[Work needed] Make registry auth optional for store deploy

viveksyngh:
This changes make registry auth optional for store deploy commands for
faas-cli. It also adds boolean flag to send registry auth.

Fixes registry auth issue for functions which are deloyed by passing image name

Fixes: #394

## Motivation and Context

## How Has This Been Tested?

Tested on my Local

```
./faas-cli store deploy figlet -g http://localhost:31112
Function figlet already exists, attempting rolling-update.

Deployed. 200 OK.
URL: http://localhost:31112/function/figlet
```

## Types of changes

## Checklist:

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -254,7 +254,7 @@ func deployImage(
 
 	functionResourceRequest1 := proxy.FunctionResourceRequest{}
 	proxy.DeployFunction(fprocess, gateway, functionName,
-		registryAuth, image, language,
+		image, registryAuth, language,
 		deployFlags.replace, envvars, network,
 		deployFlags.constraints, deployFlags.update, deployFlags.secrets,
 		labelMap, functionResourceRequest1)
--- a/commands/store_deploy.go
+++ b/commands/store_deploy.go
@@ -5,6 +5,7 @@ package commands
 
 import (
 	"fmt"
+	"log"
 	"os"
 
 	"github.com/spf13/cobra"
@@ -21,6 +22,7 @@ func init() {
 	storeDeployCmd.Flags().BoolVar(&storeDeployFlags.update, "update", true, "Update existing functions")
 	storeDeployCmd.Flags().StringArrayVar(&storeDeployFlags.constraints, "constraint", []string{}, "Apply a constraint to the function")
 	storeDeployCmd.Flags().StringArrayVar(&storeDeployFlags.secrets, "secret", []string{}, "Give the function access to a secure secret")
+	storeDeployCmd.Flags().BoolVarP(&storeDeployFlags.sendRegistryAuth, "send-registry-auth", "a", false, "send registryAuth from Docker credentials manager with the request")
 
 	// Set bash-completion.
 	_ = storeDeployCmd.Flags().SetAnnotation("handler", cobra.BashCompSubdirsInDir, []string{})
@@ -54,6 +56,12 @@ func runStoreDeploy(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("please provide the function name")
 	}
 
+	dockerConfig := configFile{}
+	err := readDockerConfig(&dockerConfig)
+	if err != nil {
+		log.Printf("Unable to read the docker config - %v\n", err.Error())
+	}
+
 	storeItems, err := storeList(storeAddress)
 	if err != nil {
 		return err
@@ -85,7 +93,11 @@ func runStoreDeploy(cmd *cobra.Command, args []string) error {
 		network = item.Network
 	}
 
+	var registryAuth string
+	if storeDeployFlags.sendRegistryAuth {
+		registryAuth = getRegistryAuth(&dockerConfig, item.Image)
+	}
 	gateway = getGatewayURL(gateway, defaultGateway, "", os.Getenv(openFaaSURLEnvironment))
 
-	return deployImage(item.Image, item.Fprocess, item.Name, "", storeDeployFlags)
+	return deployImage(item.Image, item.Fprocess, item.Name, registryAuth, storeDeployFlags)
 }


alexellis:
Thank you for turning this around. Please can you provide some examples of how this has been tested?

commands/store_deploy.go
```diff
@@ -21,6 +22,7 @@ func init() {
 	storeDeployCmd.Flags().BoolVar(&storeDeployFlags.update, "update", true, "Update existing functions")
 	storeDeployCmd.Flags().StringArrayVar(&storeDeployFlags.constraints, "constraint", []string{}, "Apply a constraint to the function")
 	storeDeployCmd.Flags().StringArrayVar(&storeDeployFlags.secrets, "secret", []string{}, "Give the function access to a secure secret")
+	storeDeployCmd.Flags().BoolVarP(&storeDeployFlags.sendRegistryAuth, "send-registry-auth", "a", false, "send registryAuth from Docker credentials manager with the request")
```
BurtonR: This shouldn't be needed at all. The function store are all public images. Adding it just adds unnecessary code to the store that has a potential for a security issue sending the credentials to the gateway when it's not needed.

commands/store_deploy.go
```diff
@@ -54,6 +56,12 @@ func runStoreDeploy(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("please provide the function name")
 	}
 
+	dockerConfig := configFile{}
+	err := readDockerConfig(&dockerConfig)
```
BurtonR: I don't believe there's a need for the users' credentials for any of the images in the store. Those images shouldn't require a login to pull.

BurtonR:
Also, not to nit-pick, but the PR template needs a little update...
>  [ ] I have raised an issue to propose this change (required)

You should check this as there is an issue that was raised.

>   [ ] Bug fix (non-breaking change which fixes an issue)
>   [ ] New feature (non-breaking change which adds functionality)
>   [X] Breaking change (fix or feature that would cause existing functionality to change)

This should be a "Bug Fix". The change you're making isn't going to break any previous versions

alexellis:
There is an issue for this which is #396 - the error shows that the code was looking up auth for a store function and blowing up. In this change we need to disable the registryAuth lookup for the store and then test the change.

BurtonR:
Disregard my comment about removing the auth from the store.

If a user changes the store url using the `--url` flag, there is a possibility that the functions there could need authentication passed in. I've updated the code comment to reflect that revelation.

ericstoekl:
I checked out this change, and I'm having some trouble getting the fix to work. Here are my repro steps:

Deploy the private docker registry:

```
docker run   --entrypoint htpasswd   registry:2 -Bbn testuser testpassword > auth/htpasswd

docker run -d   -p 5000:5000   --restart=always   --name registry   -v `pwd`/auth:/auth   -e "REGISTRY_AUTH=htpasswd"   -e "REGISTRY_AUTH_HTPASSWD_REALM=Registry Realm"   -e REGISTRY_AUTH_HTPASSWD_PATH=/auth/htpasswd   registry:2
```

Log in to the private registry

```
docker login localhost:5000 -u testuser -p testpassword
```

Tag the figlet image and push to the private registry

```
docker tag jmkhael/faas-figlet localhost:5000/faas-figlet
docker push localhost:5000/faas-figlet
```

Deploy the image with a store URL that uses this private registry

```
faas store deploy figlet --url https://gist.githubusercontent.com/ericstoekl/903d16fb9076b5f389733428997c928b/raw/36621109a21e15ee3a06c7f2e18835ba0bb91fbe/store.json
```

After doing this, the function stays at `0/1` status:

```
ID                  NAME                MODE                REPLICAS            IMAGE
     PORTS
t88erhpbnd69        figlet              replicated          0/1                 localhost:5000/faas-figlet:latest
```

@viveksyngh, can you confirm that this use case work?

viveksyngh:
This use case looks fine. I think to make the command use registry auth .. you need to set `send-registry-auth` flag. Default it is false.

BurtonR:
Just tested this on a PWD Swarm using a custom store url that included a function that is in a private Docker Hub repository.

* Deploying the `figlet` function from the default store works
* Deploying a private function and passing the `-a` flag (to include credentials) **did** work. 
* Deploying a private function **without** the `-a` flag deployed the function (it shows up on the UI), but never becomes invokable. This is the same functionality that happens with a normal `faas-cli deploy`

Looks good to me! Thanks for fixing that so quickly @viveksyngh!

BurtonR:
@ericstoekl  I see you were missing the `-a` flag to include the authentication credentials on the `store deploy` call. 
That flag makes sure that the user intends to send the credentials with the request so we don't include it when it's not needed

viveksyngh:
@ericstoekl yes it needs a HTTPS gateway.

alexellis:
The gateway doesn't need to be HTTPS for this test. The remote repository will need it though or you'll have to trust the CA if it's a self-signed cert. 

Alternatively add the registry as an insecure registry to the local machine.

alexellis:
The last comment was 11 days ago. Where are we with this? @viveksyngh @rgee0 

The last update shows Eric not being able to test it (possibly a mix-up)

alexellis:
The testing should show negative and positive scenarios.

viveksyngh:
I will test. I need to create a private docker registry and with authentication. If someone already has private docker registry with https can also give it a try.  

@ericstoekl @alexellis

alexellis:
@viveksyngh I thought the Docker hub gives one free account?

The logic shouldn't have changed for private repos.. all you're doing is making sure the store doesn't use them by default.

commands/deploy.go
```diff
@@ -254,7 +254,7 @@ func deployImage(
 
 	functionResourceRequest1 := proxy.FunctionResourceRequest{}
 	proxy.DeployFunction(fprocess, gateway, functionName,
-		registryAuth, image, language,
+		image, registryAuth, language,
```
alexellis: You're right - I didn't expect these two variables to be changed around.. so we should test with a private repo.

PR #414, opened 2018-05-23
Add support in templates for building native modules i.e. C/C++

ivanayov:
This change adds a `build-option` flag to `faas-cli build`,
e.g. `faas-cli  build  -f  stack.yml --build-option=dev`.

If  the function  language  template  has  applied  `build_options` specification in its `template.yml` file, i.e.
```
build_options:
  - name: dev
    packages:
      - musl-dev
      - gcc
      - make
```
then `ADDITIONAL_PACKAGE` agrument will be passed by `build-arg`.

This allows the listed packages to be built by the template's `Dockerfile` if it is edited to support the feature.

## Motivation and Context

[templates/#28](https://github.com/openfaas/templates/issues/28)

## How Has This Been Tested?
1. Cloned @kenfdev's function from [kenfdev/faas-pandas-example](https://github.com/kenfdev/faas-pandas-example)
2. Changed python3 template with the changes from [templates/#30](https://github.com/openfaas/templates/pull/30)
3. Successfully built the function with `faas-cli-darwin build --lang=python3 --build-option dev`
4. Successfully built the function with `faas-cli-darwin build --build-arg "ADDITIONAL_PACKAGE=make automake gcc g++ subversion python3-dev"`
5. Successfully built a simple python3 function without `--build-option` and `--build-arg "ADDITIONAL_PACKAGE=<packages>"`
6. Successfully built faas-pandas with `faas-cli-darwin build --lang=python3 --build-option dev --build-arg "ADDITIONAL_PACKAGE=make automake gcc g++"` where `template.yml` contains only `subversion`,  `python3-dev` and  `musl-dev`.
7. Successfully built a simple python3 function with `faas-cli-darwin build -f test.yml --build-arg TEST=somevalue`
8. Successfully built faas-pandas with `faas-cli-darwin build --build-arg TEST=somevalue --build-arg "ADDITIONAL_PACKAGE=make automake gcc g++ subversion python3-dev"`
9. Built with `bcrypt` in `requirements.txt`. Requires `libffi-dev` module.
10. Built with `bcrypt` and yaml config / yaml + cmd flag / yaml + `--build-arg ADDITIONAL_PACKAGE=libffi-dev` / yaml with a wrong (not available in `template.yml`) build option value / cmd flag with a wrong build option value

## Types of changes

## Checklist:

- [ ] I have updated the documentation accordingly. - TODO

## How to test

- `cd ~/go/src/github.com/openfaas/faas-cli && git fetch origin pull/414/head:dev_build`
- `git checkout dev_build`
- `./build_redist.sh`
- `cd ~/go/src/github.com/openfaas/tempates && git fetch origin pull/30/head:py3_build_native_c_modules`
- `git checkout py3_build_native_c_modules`
- `git clone git@github.com:kenfdev/faas-pandas-example.git`
- `rm -rf faas-pandas-example/template && cp template faas-pandas-example/template && cd faas-pandas-example`
- `~/go/src/github.com/openfaas/faas-cli/faas-cli-darwin build --lang=python3 --build-option dev`

You can also test with 4-10 from "How Has This Been Tested?"

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -12,12 +12,17 @@ import (
 	"github.com/openfaas/faas-cli/stack"
 )
 
+// AdditionalPackageBuildArg holds the special build-arg keyname for use with build-opts.
+// Can also be passed as a build arg hence needs to be accessed from commands
+const AdditionalPackageBuildArg = "ADDITIONAL_PACKAGE"
+
 // BuildImage construct Docker image from function parameters
-func BuildImage(image string, handler string, functionName string, language string, nocache bool, squash bool, shrinkwrap bool, buildArgMap map[string]string) {
+func BuildImage(image string, handler string, functionName string, language string, nocache bool, squash bool, shrinkwrap bool, buildArgMap map[string]string, buildOptions []string) {
 
 	if stack.IsValidTemplate(language) {
 
 		var tempPath string
+
 		if strings.ToLower(language) == "dockerfile" {
 
 			if shrinkwrap {
@@ -53,7 +58,14 @@ func BuildImage(image string, handler string, functionName string, language stri
 			}
 		}
 
-		flagSlice := buildFlagSlice(nocache, squash, os.Getenv("http_proxy"), os.Getenv("https_proxy"), buildArgMap)
+		buildOptPackages, bopErr := getBuildOptionPackages(buildOptions, language)
+
+		if bopErr != nil {
+			fmt.Println(bopErr)
+			return
+		}
+
+		flagSlice := buildFlagSlice(nocache, squash, os.Getenv("http_proxy"), os.Getenv("https_proxy"), buildArgMap, buildOptPackages)
 		spaceSafeCmdLine := []string{"docker", "build"}
 		spaceSafeCmdLine = append(spaceSafeCmdLine, flagSlice...)
 		spaceSafeCmdLine = append(spaceSafeCmdLine, "-t", image, ".")
@@ -95,7 +107,7 @@ func createBuildTemplate(functionName string, handler string, language string) s
 	return tempPath
 }
 
-func buildFlagSlice(nocache bool, squash bool, httpProxy string, httpsProxy string, buildArgMap map[string]string) []string {
+func buildFlagSlice(nocache bool, squash bool, httpProxy string, httpsProxy string, buildArgMap map[string]string, buildOptionPackages []string) []string {
 
 	var spaceSafeBuildFlags []string
 
@@ -115,7 +127,16 @@ func buildFlagSlice(nocache bool, squash bool, httpProxy string, httpsProxy stri
 	}
 
 	for k, v := range buildArgMap {
-		spaceSafeBuildFlags = append(spaceSafeBuildFlags, "--build-arg", fmt.Sprintf("%s=%s", k, v))
+
+		if k != AdditionalPackageBuildArg {
+			spaceSafeBuildFlags = append(spaceSafeBuildFlags, "--build-arg", fmt.Sprintf("%s=%s", k, v))
+		} else {
+			buildOptionPackages = append(buildOptionPackages, strings.Split(v, " ")...)
+		}
+	}
+	if len(buildOptionPackages) > 0 {
+		buildOptionPackages = deDuplicate(buildOptionPackages)
+		spaceSafeBuildFlags = append(spaceSafeBuildFlags, "--build-arg", fmt.Sprintf("%s=%s", AdditionalPackageBuildArg, strings.Join(buildOptionPackages, " ")))
 	}
 
 	return spaceSafeBuildFlags
@@ -127,4 +148,92 @@ func ensureHandlerPath(handler string) error {
 	}
 
 	return nil
+}
+
+func getBuildOptionPackages(requestedBuildOptions []string, language string) ([]string, error) {
+
+	var buildPackages []string
+
+	if len(requestedBuildOptions) > 0 {
+
+		var allFound bool
+
+		availableBuildOptions, err := getBuildOptionsFor(language)
+
+		if err != nil {
+			return nil, err
+		}
+
+		buildPackages, allFound = getPackages(availableBuildOptions, requestedBuildOptions)
+
+		if !allFound {
+			err = fmt.Errorf("Error: You're using a build option unavailable for %s. Please check /template/%s/template.yml for supported build options", language, language)
+			return nil, err
+		}
+
+	}
+	return buildPackages, nil
+}
+
+func getBuildOptionsFor(language string) ([]stack.BuildOption, error) {
+
+	var buildOptions = []stack.BuildOption{}
+
+	pathToTemplateYAML := "./template/" + language + "/template.yml"
+	if _, err := os.Stat(pathToTemplateYAML); os.IsNotExist(err) {
+		return buildOptions, err
+	}
+
+	var langTemplate stack.LanguageTemplate
+	parsedLangTemplate, err := stack.ParseYAMLForLanguageTemplate(pathToTemplateYAML)
+
+	if err != nil {
+		return buildOptions, err
+	}
+
+	if parsedLangTemplate != nil {
+		langTemplate = *parsedLangTemplate
+		buildOptions = langTemplate.BuildOptions
+	}
+
+	return buildOptions, nil
+}
+
+func getPackages(availableBuildOptions []stack.BuildOption, requestedBuildOptions []string) ([]string, bool) {
+	var buildPackages []string
+
+	for _, requestedOption := range requestedBuildOptions {
+
+		requestedOptionAvailable := false
+
+		for _, availableOption := range availableBuildOptions {
+
+			if availableOption.Name == requestedOption {
+				buildPackages = append(buildPackages, availableOption.Packages...)
+				requestedOptionAvailable = true
+				break
+			}
+		}
+		if requestedOptionAvailable == false {
+			return buildPackages, false
+		}
+	}
+
+	return deDuplicate(buildPackages), true
+}
+
+func deDuplicate(buildOptPackages []string) []string {
+
+	seenPackages := map[string]bool{}
+	retPackages := []string{}
+
+	for _, packageName := range buildOptPackages {
+
+		if _, alreadySeen := seenPackages[packageName]; !alreadySeen {
+
+			seenPackages[packageName] = true
+			retPackages = append(retPackages, packageName)
+		}
+	}
+	return retPackages
 }
--- a/commands/build.go
+++ b/commands/build.go
@@ -18,16 +18,14 @@ import (
 
 // Flags that are to be added to commands.
 var (
-	nocache     bool
-	squash      bool
-	parallel    int
-	shrinkwrap  bool
-	buildArgs   []string
-	buildArgMap map[string]string
-	buildOption string
+	nocache      bool
+	squash       bool
+	parallel     int
+	shrinkwrap   bool
+	buildArgs    []string
+	buildArgMap  map[string]string
+	buildOptions []string
 )
-
-const additionalPackageBuildArg = "ADDITIONAL_PACKAGE"
 
 func init() {
 	// Setup flags that are used by multiple commands (variables defined in faas.go)
@@ -38,15 +36,11 @@ func init() {
 
 	// Setup flags that are used only by this command (variables defined above)
 	buildCmd.Flags().BoolVar(&nocache, "no-cache", false, "Do not use Docker's build cache")
-	buildCmd.Flags().BoolVar(&squash, "squash", false, `Use Docker's squash flag for smaller images
-						 [experimental] `)
+	buildCmd.Flags().BoolVar(&squash, "squash", false, `Use Docker's squash flag for smaller images [experimental] `)
 	buildCmd.Flags().IntVar(&parallel, "parallel", 1, "Build in parallel to depth specified.")
-
 	buildCmd.Flags().BoolVar(&shrinkwrap, "shrinkwrap", false, "Just write files to ./build/ folder for shrink-wrapping")
-
 	buildCmd.Flags().StringArrayVarP(&buildArgs, "build-arg", "b", []string{}, "Add a build-arg for Docker (KEY=VALUE)")
-
-	buildCmd.Flags().StringVar(&buildOption, "build-option", "", "Set a build option, e.g. dev")
+	buildCmd.Flags().StringArrayVarP(&buildOptions, "build-option", "o", []string{}, "Set a build option, e.g. dev")
 
 	// Set bash-completion.
 	_ = buildCmd.Flags().SetAnnotation("handler", cobra.BashCompSubdirsInDir, []string{})
@@ -86,14 +80,6 @@ via flags.`,
 func preRunBuild(cmd *cobra.Command, args []string) error {
 	language, _ = validateLanguageFlag(language)
 
-	if buildOption != "" {
-		arg, validBuildOption, err := validateBuildOption(buildOption, language)
-
-		if validBuildOption && err == nil {
-			buildArgs = append(buildArgs, arg)
-		}
-	}
-
 	mapped, err := parseBuildArgs(buildArgs)
 
 	if err == nil {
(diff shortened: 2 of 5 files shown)


alexellis:
Hi Ivana, can you also include a repo with a Dockerfile or template which has this change applied? I assume that would be needed for us to try out the change?

iyovcheva:
@alexellis This is a WIP PR for the python3 template - [templates/#30](https://github.com/openfaas/templates/pull/30)

rgee0:
`faas-cli build -f stack.yml --build-option=dev --build-args ADDITIONAL-PACKAGE=jq`

Is this a supported use case?

Which takes precedence?  dev packages, or jq, or do you get both?

alexellis:
I would say this is a good point. What is the user trying to do? Perhaps they are wanting to get the union of both sets of packages?

rgee0:
As `buildOption` will ultimately become a `buildArg`, I wonder whether we ought to convert the `buildOption` to a `buildArg` in `preRunBuild()` which, I think, would lessen the rework beyond that.

Is the intention for users to pass a single buildOption, or multiple?

alexellis:
For the first version - a single buildOption. i.e. including dev dependencies.

builder/build.go
```diff
@@ -128,3 +137,62 @@ func ensureHandlerPath(handler string) error {
...
+
+	if buildOption != "" {
+		buildOptions, err := deriveBuildOptions(language)
+
+		if err != nil {
+			return result, err
+		}
+
+		switch buildOption {
+		case "dev":
```
rgee0: We cant have a case/select that covers each buildOption as we'll have to open the code on every template change and 3rd party templates will be constrained to officially supported options.

builder/build.go
```diff
@@ -95,7 +95,7 @@ func createBuildTemplate(functionName string, handler string, language string) s
 	return tempPath
 }
 
-func buildFlagSlice(nocache bool, squash bool, httpProxy string, httpsProxy string, buildArgMap map[string]string) []string {
+func buildFlagSlice(nocache bool, squash bool, httpProxy string, httpsProxy string, buildArgMap map[string]string, buildOption string, language string) []string {
```
rgee0: The `buildFlagSlice()` signature could remain the same if the `buildOption` is converted to BuildArgs earlier.

builder/build.go
```diff
@@ -118,6 +118,15 @@ func buildFlagSlice(nocache bool, squash bool, httpProxy string, httpsProxy stri
 		spaceSafeBuildFlags = append(spaceSafeBuildFlags, "--build-arg", fmt.Sprintf("%s=%s", k, v))
 	}
 
+	if buildOption != "" {
+		buildOptionPackages, err := getBuildOptionPackages(buildOption, language)
+
+		//TODO is this ok?
+		if err == nil {
+			spaceSafeBuildFlags = append(spaceSafeBuildFlags, "--build-arg", fmt.Sprintf("ADDITIONAL_PACKAGE=%s", buildOptionPackages))
```
rgee0: We need to handle situations where the user has passed a `--build-arg` with the same name.

iyovcheva:
Thanks @rgee0 
I addressed your comments and tested the changes. Can you please review?

iyovcheva:
Thanks for the comments @rgee0 . Yes, I absolutely agree - it's much more clear this way.

alexellis:
@rgee0 thanks for working with @iyovcheva on this

Ivana, please could you provide a few commands that we can type in to build and make use of the work so far including the template and checking out your PR? https://help.github.com/articles/checking-out-pull-requests-locally/

ericstoekl:
Would I be able to include build options in the `stack.yml`, if they are missing from the `template.yml` in the template folder? This would be useful for including additional dependencies in functions deployed to Openfaas-Cloud.

In other words, could I include a 

```
build-options:
  - jq
```

 list in the `stack.yml` file, then just run `faas build` instead of `faas build --build-args ADDITIONAL-PACKAGES=jq`?

alexellis:
Via CLI args you could add a custom override with this design. I think we should get it in and try it out with some of the users who need native modules.

iyovcheva:
@alexellis I updated the PR body with `How to test` section.

alexellis:
Please can we also make `stack.yml` support `build_options`, but not with arbitrary values, it should only accept something defined in `template.yml`, for example: `dev`.

For this comparison I've invented a fake package called `debug`. Do we prefer a nested or flat structure? [Docker Compose uses a nested structure](https://docs.docker.com/compose/compose-file/#build)

```
provider:
  name: faas

functions:
  test-fn:
    lang: python3
    handler: ./t
    image: test-fn:0.1
    build_options:
      - dev
      - debug
```

Or with some nesting structure:

```
provider:
  name: faas

functions:
  test-fn:
    lang: python3
    handler: ./t
    image: test-fn:0.1
    build:
       options:
         - dev
         - debug
```

iyovcheva:
IMO a nested structure is better - looks well organised and more readable in future, when new flags appear. And it follows Docker's standards

rgee0:
`buildOption` as a flag and `build\options` in the stack?

I don't mind as long as these things are consistent.  A user shouldn't have to learn different options for each mode of operation.

Are the options going to be cumulative too? so I could specify `buildOption=debug` on the CLI and
```
build:
  options:
    - debug
```
in my stack.yaml

And perhaps `build-arg ADDITIONAL-PACKAGES=burt`

What is the thinking behind warn over error if a bad `buildOption` or  `build\options` is/are supplied?

iyovcheva:
@rgee0 I would say an error. Using a bad `buildOption` means that the function will not build as the user expect, so it should better fail.

Failing with a good error message can give better visibility on the problem and point the user to
1. Check the language `template.yml` for an appropriate build option
2. Add new build option to the language template, if there's no available one

Any thoughts?

iyovcheva:
@alexellis @rgee0 I added `stack.yaml` support. Can you please review?

rgee0:
I agree on erroring - that was my original direction.  @alexellis suggested warn, hence my query.

alexellis:
Hey Richard which of the comments need to be resolved to release this as a preview feature? Is it useful as it is?

rgee0:
If we’d have taken a cut before the scope broadened to include yaml then I think we’d have had a preview feature with the cli. Its arguable that the cli and yaml changes could have been two separate features, especially given the drift mentioned above - are you sure about the different notations being adopted?

The de-duplication needs largest focus - a test might help in this regard. The other changes are readability and optimising.

commands/build.go
```diff
@@ -109,12 +125,102 @@ func parseBuildArgs(args []string) (map[string]string, error) {
...
+func extendBuildArgMap(newBuildArg string) {
+	separator := strings.Index(newBuildArg, "=")
+	values := []string{newBuildArg[0:separator], newBuildArg[separator+1:]}
+
+	argKey := strings.TrimSpace(values[0])
+	argValue := strings.TrimSpace(values[1])
+
+	if packagesArg, hasPackagesArg := buildArgMap[argKey]; hasPackagesArg {
+		/* Check if the same arg already contains the same value */
+		if !strings.Contains(packagesArg, argValue) {
```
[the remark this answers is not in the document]
iyovcheva: I will remove this check. You're right it's not harmful and I did test duplicating the packages.
This needs too much logic/parsing to put in for a not important use-case.

alexellis:
We have this scenario already for --label / --env where multiple values can be specified. Please see how that is implemented. There's no need to encode the values within a single flag when using Cobra for parsing arguments.

alexellis:
@rgee0 

>  are you sure about the different notations being adopted?

Can you give the context of this?

alexellis:
Yes I think you're right - if we had released with just the CLI flag we could have finished the PR and raised the YAML part for afterwards.

@iyovcheva to merge we'll need some unit tests on the logic you've written. I'd also recommend checking that code like `extendBuildArgMap` isn't duplicating existing code used to handle the other maps and arrays that we handle for labels/env/secrets etc.

rgee0:
> Can you give the context of this?
I'm referring to `--build-option` in one mode and `build.options` in the other

alexellis:
So for OpenFaaS Cloud you will not be able to specify `--build-option` since it's observed by a controller, in that case we have to define this declaratively. So if you want native modules with OpenFaaS Cloud or a CI/CD job in another tool like Jenkins then it'll be the only way to specify it.

rgee0:
That looks like reason for implementing both modes (not in question), rather than reasoning for differing nomenclature.

iyovcheva:
Travis build fails with an unrelated error
```
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install docker-ce" failed and exited with 100 during .
```

Tested the build locally with `./build_redist.sh`

Yaml support is now moved to [dev_build_yml](https://github.com/iyovcheva/faas-cli/tree/dev_build_yml) and unit test is added for the initial implementation.

alexellis:
@iyovcheva I'm seeing the build pass now 👍

alexellis:
@iyovcheva

I've had a quick look at the code and it looks fine to me, I know you've already had feedback from @rgee0 on this. Here are three things I'd like to see before merging the feature.

* We need a documentation PR before this can be merged over at https://github.com/openfaas/docs/ - we will review the use of English before merging so don't worry if you're not able to get it perfect on the first attempt. It should describe objectively in the CLI heading how to use this new feature.

* We'll also need a PR for at least one template to add the build option into its `template.yml` and Dockerfile.

* The third thing will be to do some edge-case testing. What if someone tries this with the Node.js template for instance? What happens then?

Finally have someone else validate the change by using it - both in a happy path i.e. with your Pythion change and the non-happy path - i.e. Node (where we don't have the change yet). The CLI is very user-facing.

iyovcheva:
@alexellis those are the following PRs:

Docs update: [#32](https://github.com/openfaas/docs/pull/32)

Updated templates:
* [python](https://github.com/openfaas/templates/pull/30) - all versions
* [ruby](https://github.com/openfaas/templates/pull/34)

iyovcheva:
Node.js works with a normal build.
Tested with `npm install gtest`, `npm install v8`

```
Step 15/24 : RUN npm install gtest
 ---> Running in 4b8a279773cf

+ gtest@0.0.0
added 1 package in 2.476s
```

```
Step 15/24 : RUN npm install v8
 ---> Running in bdd598cf49e3

+ v8@0.1.0
added 1 package in 0.513s
```

rgee0:
On the topic of test cases, can we see how the dev build option for Python operates with a function with this in its requirements.txt:
```
numpy
scipy
scikit-learn
```

alexellis:
@iyovcheva please can you provide an update on each point of my comment:

https://github.com/openfaas/faas-cli/pull/414#issuecomment-396293793

@rgee0 if you know the context feel free to fill in some of the detail

I don't see that anyone else has validated or tested the change yet. This is blocking the merge. 

Alex

rgee0:
AFAIK (@iyovcheva correct me):

- **Documentation** - https://github.com/openfaas/docs/pull/32

- **Additional Templates** - Node and Python template PRs linked above

- **Edge cases** - idea provided above

- **3rd party validation** - still to be done

BurtonR:
I've been running through testing this this morning. 
When running the build as: 

`./faas-cli build -f tester.yml --build-arg "ADDITIONAL_PACKAGE=git"`

The `git` package is added and available as expected.

Also, verified multiple packages do get included in the build image:

`./faas-cli build -f tester.yml --build-arg "ADDITIONAL_PACKAGE=git vim"`

However, I am unable to get the `--build-option` flag to work. I've updated the node template myself to include the `build_options` parameter, and I've also copied down the python templates from the PR linked, but none of those seem to be taking the `--build-option` flag's value

`tester.yml` is generated from `./faas-cli new tester --lang python` with the updated `template.yml` and `Dockerfile` from the template PR

I've tried:
`./faas-cli build -f tester.yml --build-option dev`
`./faas-cli build -f tester.yml --build-option blah`

For the `dev` build-option, the packages listed in the `template.yml` `build_option` list are not included

Also, I expect the `blah` to fail (as it's not defined), but there is no warning or error, and the build completes.

After some debugging, I found that the `build-option` flag requires the `--lang` flag to be included in the build command. This caused me quite a bit of confusion as I would expect the lang to get picked up by the setting in the function's .yml file, rather than having to declare it twice (once when generating the function, and again when building it).

This allows users to specify 2 different languages for each command. Create a function from the python template, but call build with the node language which could cause issues, or at the least unexpected behaviour.

alexellis:
Thank you for looking at this Burton. This is the reason why I ask for a second person to validate some changes, especially when they are user-facing.

I'm not sure why the build option needs a lang flag - I think we need to sort these quirks out before merging and releasing.

Ivana let me know if you need help with anything and an ETA for when these changes will be finished.

rgee0:
I've been reworking this a little, not least because the language quirk was a result of one of my initial suggestions.  @iyovcheva it'd be worth us syncing on Slack when you're online next.

There are some cases around this that I'm not sure we've fully considered yet.  For instance, what happens if there are multiple functions in the yaml and the `--build-option` switch is added on the CLI? Particularly if there are multiple languages and they don't all support the stated build option?

alexellis:
There are three primary examples:

* YAML file:
  * single function
  * multiple functions

* CLI args only

The problematic example is YAML file with multiple functions being targeted (remember we can also apply a `--filter` arg) In that instance it is unclear what behavior to use for any flag - let alone `build_options`.

alexellis:
What happens now if we use ` --build-option` with Node.js? We should get an error - this is one of the edge-cases that we need to validate.

alexellis:
Could someone respond to the comment by Burton? https://github.com/openfaas/faas-cli/pull/414#issuecomment-397824696

alexellis:
Please can someone respond to my comment too? https://github.com/openfaas/faas-cli/pull/414#issuecomment-397844656

alexellis:
Let's see the "How to test" example updated to use an example with Pillow in Python for image editing and a quoted text block we can copy/paste to test.

```sh
this is a
quoted
block
```

alexellis:
We should also try this on ARM, but not with a heavy-weight library like numpy which takes a very long time to compile on that platform.

rgee0:
Response to question about when option isn't available for a language. This is the current output:
```
./faas build -f tester.yml --build-option dev --no-cache --build-arg ADDITIONAL_PACKAGE=jq --build-arg ADDITIONAL_PACKAGE=
icu --build-option debug
[0] > Building tester.
Clearing temporary build folder: ./build/tester/
Preparing ./tester/ ./build/tester/function
Building: tester with node template. Please wait..
Error: You're using a build option unavailable for node. Please check /template/node/template.yml for supported build options
[0] < Building tester done.
[0] worker done.
```

rgee0:
@burtonr The double language feature has been addressed, can you retest, please?

burtonr:
Just tested this again and I like it :smiley: 
Ran through a couple scenarios, and they all behaved as I would expect:
(where `tester` is  a python template _with_ the build_options available)
* `./faas-cli build -f tester.yml --build-option dev`
    * built function _with_ extra packages included
* `./faas-cli build -f tester.yml --build-option blah`
    * did _not_ build function with error saying build option not available
* `./faas-cli build -f tester.yml --build-arg "ADDITIONAL_PACKAGE=git"`
    * built function with the `git` package included
* `./faas-cli build -f node-tester.yml --build-option dev`
    * did _not_ build function with error saying build option not available
* Modified the node template to include a build_options section in the `template.yml`
* `./faas-cli build -f node-tester.yml --build-option dev`
    * built function with additional packages included as expected

One thing I did notice though, is that when using the `--build-arg "ADDITIONAL_PACKAGE=git"` on a template that has not been modified, doesn't do anything. It builds normally, and the additional package is not included. This is pretty obvious because the template doesn't have the build arg defined, but perhaps that would be a bit surprising to a user not intently watching the build output.

Not sure if that's an issue, or really how to resolve it beyond parsing the Dockerfile for the build-arg before building. Just something I noticed...

Looks good to me though! Thanks @ivanayov  for working through this month-long PR!

rgee0:
Thanks Burton.

I think we can mitigate the `--build-arg "ADDITIONAL_PACKAGE=git"` item you mention by updating __all__ the official template Dockerfiles to include the `ARG` and the call to install the packages.  This can be done separately to defining build-options for the templates. @ivanayov thoughts on broadening your templates PR to accommodate?

alexellis:
I think I'm OK with `--build-arg "ADDITIONAL_PACKAGE=git"` not doing anything. It's when we specifically request the new feature via a flag or YAML that we should warn.

alexellis:
The Python and Ruby template PRs have been merged, which should be no-harm changes. They will need the `git` package though as some pips do clone source to build.

alexellis:
Thanks all for your work on this. I've merged and released under https://github.com/openfaas/faas-cli/releases/tag/0.6.11

If we need further changes they can be done in another point release.

Once CI is complete and the binaries are uploaded I will merge the documentation PR. https://github.com/openfaas/docs/pull/32

Issue #432, opened 2018-07-09
Feature: Generate unique image tag on build


stefanprodan:
Faas-cli `build` should produce unique image tags as in `gitbranch-gitsha` besides the `latest` one. 
Faas-cli `push` should push the unique tag along with `latest`.

## Expected Behaviour

The faas-cli CI pipeline should:
- detect Git
- get the current branch name and commit SHA
- tag the function image with `gitbranch-gitsha`
- build, push and deploy the  `function:gitbranch-gitsha` image

## Current Behaviour
Only the `latest` tag is applied and pushed to image registry.

## Possible Solution

Detect Git and compose the image tag based on branch and commit.

## Context

Using `latest` for production systems is problematic, there is no way to trace what version of a function you are running. 

Having unique image tags for each build enables:
- easy rollbacks 
- automatic update with tools like Weave Flux
- canary and A/B testing with Istio

stefanprodan:
Derek add label: proposal

alexellis:
We do something like this with OpenFaaS cloud, this is what we have done in OpenFaaS cloud:

Example from OpenFaaS Cloud via the functions for @kenfdev: 

```json
[...]

PR #447, opened 2018-07-19
Generate unique image tag on build

ivanayov:
With this change `faas-cli build --tag-sha` produces unique image tag as `gittag-gitsha` or `latest-gitsha` besides the `latest` one and pushes both to the Docker registry.

The reason behind this change is that using only `latest` for production systems can be problematic, as there is no way to trace what version of the function is running

## Motivation and Context

#432 

## How Has This Been Tested?

Building function on a branch without tag for latest commit
```
$ faas-cli-darwin build -f shatest.yml --tag-sha

$ docker image ls | grep sha
REPOSITORY                                 TAG                                            IMAGE ID            CREATED             SIZE
docwareiy/shatest                          latest                                         46e0fb95e9b8        About an hour ago   94MB
docwareiy/shatest                          latest-c568a42                                 46e0fb95e9b8        About an hour ago   94MB
```

Building function on a branch without tag for latest commit
```
$ faas-cli-darwin build --tag-sha

$ docker image ls | grep sha
REPOSITORY                                 TAG                                            IMAGE ID            CREATED             SIZE
docwareiy/swquotes                         0.1.0-b8c4fcd                                  97dafe67abd3        3 weeks ago         94MB
docwareiy/swquotes                         latest                                         97dafe67abd3        3 weeks ago         94MB
```

Then tested `faas push` and checked DockerHub. Functions were pushed with both latest and SHA tags.

## Types of changes

## Checklist:

- [ ] I have updated the documentation accordingly. - TODO
- [ ] I have added tests to cover my changes. - TODO


builder/build.go
```diff
@@ -17,7 +17,7 @@ import (
 const AdditionalPackageBuildArg = "ADDITIONAL_PACKAGE"
 
 // BuildImage construct Docker image from function parameters
-func BuildImage(image string, handler string, functionName string, language string, nocache bool, squash bool, shrinkwrap bool, buildArgMap map[string]string, buildOptions []string) {
+func BuildImage(image string, handler string, functionName string, language string, nocache bool, squash bool, shrinkwrap bool, buildArgMap map[string]string, buildOptions []string, tagsha bool) {
```
alexellis: `tagsha` has poor casing for Go, this should be tagSHA, or just `sha` (I prefer `sha`).

builder/build.go
```diff
@@ -237,3 +240,18 @@ func deDuplicate(buildOptPackages []string) []string {
 	}
 	return retPackages
 }
+
+func getVersion() string {
```
alexellis: What about when there is no git folder available? I think we should check that before running exec. You can find out if you're in a git repo by looking for a `.git` folder.

commands/build.go
```diff
@@ -41,6 +42,7 @@ func init() {
 	buildCmd.Flags().BoolVar(&shrinkwrap, "shrinkwrap", false, "Just write files to ./build/ folder for shrink-wrapping")
 	buildCmd.Flags().StringArrayVarP(&buildArgs, "build-arg", "b", []string{}, "Add a build-arg for Docker (KEY=VALUE)")
 	buildCmd.Flags().StringArrayVarP(&buildOptions, "build-option", "o", []string{}, "Set a build option, e.g. dev")
+	buildCmd.Flags().BoolVar(&tagsha, "tag-sha", false, "Tag function image with gitsha. This doesn't replase latest tag, but provides second with versioning")
```
alexellis: `tag-sha` was my suggestion, but I don't know if it's a good option. Do you have other suggestions?

`--auto-tag` or just `--tag` ? If users use this a lot we need it to be easy to use. Also You mention `gitsha` and I think you mean "Git SHA".

The comment `replase` is a typo. Let's just dump the second sentence.

ivanayov: I would suggest `--auto-tag` as it's more descriptive.

builder/build.go
ivanayov: It's addressed. The change is on the next line and will not show as outdated.

alexellis:
Please see note: https://github.com/openfaas/faas-cli/issues/432#issuecomment-406283488

alexellis:
Commit structure looks OK to me now.

alexellis:
Merged via other commits. Will be in 0.6.16.

Issue #452, opened 2018-07-19
Registry auth with `-send-registry-auth` fails for nested images in custom registry (Gitlab)


johnmccabe:
## Expected Behaviour

It should be possible to use a private registry by setting the `-send-registry-auth` flag to extract an auth header from the docker credentials manager.

## Current Behaviour

This currently fails on Gitlab when using an image thats nested under the parent repo.

For example this works:

    registry.gitlab.com/johnmccabe/myfunctions

But this fails to find any credentials:

    registry.gitlab.com/johnmccabe/myfunctions/coolfn

Using `docker pull/push` etc all work as expected with the private registries.

## Possible Solution

Fix the getRegistryAuth domain extraction function in `commands/deploy.go`.

## Steps to Reproduce (for bugs)

1. Create a private repo in Gitlab, for example: `johnmccabe/test-registry`
2. Push an image nested under that repo, for example: `registry.gitlab.com/johnmccabe/test-registry/gitlabfn`
3. Ensure the image pushed above is deleted locally
4. Docker login to the gitlab repo
5. Attempt to deploy, you should see the function fail to pull the image (`docker service ls`).

## Context

Number of users have been encountering issues with custom registries.

## Your Environment

* FaaS-CLI version ( Full output from: `faas-cli version` ): 
```
  ___                   _____           ____
 / _ \ _ __   ___ _ __ |  ___|_ _  __ _/ ___|
| | | | '_ \ / _ \ '_ \| |_ / _` |/ _` \___ \
| |_| | |_) |  __/ | | |  _| (_| | (_| |___) |
 \___/| .__/ \___|_| |_|_|  \__,_|\__,_|____/
      |_|

CLI:
 commit:  6c28891cf648621ef120c802d9160cae7506b977
 version: 0.6.14

Gateway
 uri:     http://127.0.0.1:8080

Provider
 name:          faas-swarm
 orchestration: swarm
 version:       0.3.5
 sha:           b8cc399b5eadfb59956b8341a44ea3ba4acf1e7f
```

* Docker version ( Full output from: `docker version` ):
```
Client:
 Version:      18.05.0-ce
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   f150324
 Built:        Wed May  9 22:12:05 2018
 OS/Arch:      darwin/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.05.0-ce
  API version:  1.37 (minimum version 1.12)
[...]

PR #453, opened 2018-07-19
Fix registry auth for nested images in custom registry (Gitlab)

johnmccabe:
## Description
This PR fixes the getRegistryAuth function so that it works correctly with images that exist in subdirs on custom docker registries, for example Gitlab.

Before this change the following image works as expected:

    registry.gitlab.com/johnmccabe/myfunctions

But an image nested under that namespace would fail to find any auth from the docker cached creds, for example this would fail:

    registry.gitlab.com/johnmccabe/myfunctions/coolfn

This fix address the underlying issue in the registry domain extraction so that both cases now work, I've tested with private registries on both Gitlab and Docker Hub successfully.

Have also added a test to catch the nested subdir image case.

## Motivation and Context

- [x] I have raised an issue to propose this change (#452)

## How Has This Been Tested?

Tested **on OSX** with private registry on both Gitlab (hosted), and Docker Hub.

## Types of changes

## Checklist:

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -464,7 +464,7 @@ func getRegistryAuth(config *configFile, image string) string {
 		slashes := strings.Count(image, "/")
 		if slashes > 1 {
 			regS := strings.Split(image, "/")
-			registry = strings.Join(regS[:len(regS)-2], ", ")
+			registry = regS[0]
 		}
 
 		if registry != "" {
--- a/commands/deploy_test.go
+++ b/commands/deploy_test.go
@@ -70,6 +70,22 @@ func Test_getRegistryAuth_CustomRegistry_Found(t *testing.T) {
 	}
 }
 
+func Test_getRegistryAuth_NestedGitlabRegistry_Found(t *testing.T) {
+	wantAuth := "alexellis2-auth-str"
+	configFile1 := configFile{
+		AuthConfigs: map[string]authConfig{
+			"registry.gitlab.com": authConfig{Auth: wantAuth},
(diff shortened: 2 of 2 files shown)


johnmccabe:
@tarunmangukiya this should fix the issue for you on Gitlab, can you run this build of the cli on your Windows env and let us know please.

johnmccabe:
@alexellis ^ do you have a K8s setup with a private registry to hand that you could sanity check this on.

tarunmangukiya:
Sure, I'll test it asap.

tarunmangukiya:
@johnmccabe @alexellis Tested successfully. Working for my private GitLab repo.

commands/deploy.go
```diff
@@ -464,7 +464,7 @@ func getRegistryAuth(config *configFile, image string) string {
 		slashes := strings.Count(image, "/")
 		if slashes > 1 {
 			regS := strings.Split(image, "/")
-			registry = strings.Join(regS[:len(regS)-2], ", ")
+			registry = regS[0]
 		}
```
alexellis: Do we need to check this index first?

alexellis:
LGTM, wonder if this affected the issue on Nexus too?

commands/deploy.go
alexellis: @johnmccabe

johnmccabe: Nope, since you've checked that the number of slashes is > 0 above

alexellis: ty

PR #409, opened 2018-05-13
Add --tls-no-verify to commands

affix:
From Issue #376 add a --tls-no-verify flag

## Description
Adds a --tls-no-verify flag to commands. This defaults to false to that TLS/SSL certificates must be valid.

Append the --tls-no-verify flag to your command to skip TLS/SSL Verification.

## Motivation and Context
solves issue #376

## How Has This Been Tested?
Tested manually agains my faas-nettes setup with an nginx proxy

**Valid SSL**
- Can Login Successfuly without --tls-no-verify  :white_check_mark:
- Can Login Successfuly with --tls-no-verify :white_check_mark:
- Can Deploy Successfuly without --tls-no-verify :white_check_mark:
- Can Deploy Successfuly with --tls-no-verify :white_check_mark:
- Can List Successfuly without --tls-no-verify :white_check_mark:
- Can List Successfuly with --tls-no-verify :white_check_mark:
- Can Invoke Successfuly without --tls-no-verify :white_check_mark:
- Can Invoke Successfuly with --tls-no-verify :white_check_mark:

**No SSL**
- Can Login Successfuly without --tls-no-verify :white_check_mark:
- Can Login Successfuly with --tls-no-verify :white_check_mark:
- Can Deploy Successfuly without --tls-no-verify :white_check_mark:
- Can Deploy Successfuly with --tls-no-verify :white_check_mark:
- Can List Successfuly without --tls-no-verify :white_check_mark:
- Can List Successfuly with --tls-no-verify :white_check_mark:
- Can Invoke Successfuly without --tls-no-verify :white_check_mark:
- Can Invoke Successfuly with --tls-no-verify :white_check_mark:

**Invalud SSL**
- Can Not Login Successfuly without --tls-no-verify :white_check_mark: 
- Can Login Successfuly with --tls-no-verify :white_check_mark:
- Can Not Deploy Successfuly without --tls-no-verify :white_check_mark:
- Can Deploy Successfuly with --tls-no-verify :white_check_mark:
- Can Not List Successfuly without --tls-no-verify :white_check_mark:
- Can List Successfuly with --tls-no-verify :white_check_mark:
- Can Not Invoke Successfuly without --tls-no-verify :white_check_mark:
- Can Invoke Successfuly with --tls-no-verify :white_check_mark:

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

[...]

PR #460, opened 2018-07-22
Rebase/review for 409

alexellis:
Rebase and review of #409

- rebased onto master
- "true" moved to named variable in all tests
- "true" set to "false" in store list - should not be set to true
by default, but false.

Signed-off-by: Keiran Smith <contact@keiran.scot>


alexellis:
@ewilde @LucasRoesler unfortunately there are some contending changes around this area. This PR is from much earlier and needs merging first. I think Lucas' work is already in but Ed's may need a rebase.

The signature for deploy function is growing too large, so I think we need to introduce a struct into which to set the arguments. This will make multiple-clashing patches easier to merge too.

Any thoughts?

Alex

Issue #472, opened 2018-07-24
Feature: "watch" mode -> automatic build/push/deploy on save


alexellis:
## Expected Behaviour

New feature: file-system watch mode will detect a change in a function handler folder and then automatically build/push/deploy the affected function.

This would be done via the existing OpenFaaS API meaning all backends are supported - faas-fargate, faas-swarm, faas-netes and openfaas-operator.

Similar things are done in entire projects such as [skaffold](https://github.com/GoogleContainerTools/skaffold) from Google. I know Stefan @stefanprodan  is a fan of this project. I believe Skaffold needs a Dockerfile in a set place, we can workaround this by using "faas-cli build --shrinkwrap" or maybe a combination of all of the above.

The basic idea is to allow a "watch" / auto-reloader experience for function development.

Thoughts?

alexellis:
I was chatting to @johnmccabe about this, we're thinking of adding `faas-cli up` with an optional `--watch` flag which would be an alias for the whole workflow of `build/push/deploy`

alexellis:
@johnmccabe how did the initial attempt go with this? How does Docker-Compose handle de-bounce?

johnmccabe:
Initial PoC worked well, I'll pick this up after work tomorrow and get something pushed up for eyeballs.

alexellis:
Awesome John.

johnmccabe:
Yeah, needs dusted off and pushed (got side-tracked helping with the minikube stuff) - I can swing back round to this on Sunday after I get VMUG and PHP Scotland out of the way.

alexellis:
Do you still have a local copy? Anything we should be aware of or watch for?

PR #503, opened 2018-08-16
Add `up` alias for build, push, deploy

johnmccabe:
## Description
Add `up` alias for `build` ➡︎ `push` ➡︎ `deploy`.

faas-up

faas-up-skip-push

~~This reuses all the flags/logic from the respective commands which are chained together, it adds a single local flag `--push` which is used to include the push step.~~

~~Alternatives to setting `--push` would be to do so for non-`localhost`/`127.0.0.1` gateways perhaps. @alexellis wdyt?~~

This reuses all the flags/logic from the respective commands which are chained together, it adds a single local flag `--skip-push` which is used to exclude the push step.

## Motivation and Context

- [x] I have raised an issue to propose this change (#472)

## How Has This Been Tested?

Built and tested against a local gateway with both a local and docker hub image push/skip-push.

No unit tests added as it just reuses the underlying commands.

## Types of changes

## Checklist:

### DIFF
--- a/commands/up.go
+++ b/commands/up.go
@@ -1,4 +1,4 @@
-// Copyright (c) Alex Ellis 2017. All rights reserved.
+// Copyright (c) OpenFaaS Author(s) 2018. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 package commands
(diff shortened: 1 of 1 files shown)


alexellis:
Let's just have `faas-cli push` refuse if there is no prefix on the image. I think that's a separate change though. cc @viveksyngh

commands/up.go
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Alex Ellis 2017. All rights reserved.
```
viveksyngh: Please use `OpenFaaS Author(s) 2018` for copyright. One of the feedback from @alexellis  on my previous PR :)

alexellis: +1 for this

johnmccabe: yup, its a WIP pr was going to address this later - funnily enough I was looking for that header text when raising the PR and noticed this - https://github.com/openfaas/faas-cli/blob/master/commands/update_gitignore.go, you'll need to update that one as well :)

burtonr:
Awesome @ivanayov  I was wondering the same thing, but haven't had time to test it out yet. Good to know!

Issue #481, opened 2018-07-27
Suggestion: Configurable function directory name


dmcclory:
It would be nice if a user could specify the output directory for the function code when they are creating a new function.

## Expected Behaviour

I'm imagining there would be a new flag:

```
faas new name-with-hyphens --lang=python3 --function-dir=simplename
```

the resulting tree would be:
```
.
├── simplename
│   ├── __init__.py
│   ├── handler.py
│   └── requirements.txt
├── name-with-hyphens.yml
```

and the stack yaml file would be:

```
provider:
  name: faas
  gateway: http://127.0.0.1:8080

functions:
  name-with-hyphens:
    lang: python3
    handler: ./simpledirname
    image: name-with-hyphens
```

## Current Behaviour

Right now, people need to make those changes manually.

## Possible Solution
It seems like it would require a few changes to `commands/new_function.go`

## Context
The only characters that are allowed in a function name are "a-z, 0-9 and dashes". I'm in an environment where I'd like be able to have function names that use those characters, but also guarantee that the output dir will be a valid python module name. I know that that is probably a very specific use case, but there may be other edge cases that this would handle.

viveksyngh:
Derek add label: proposal

alexellis:
Seems like a fair request, @dmcclory would you be willing to work on this yourself?

alexellis:
`--function-dir` could be more a little more concise, otherwise I think this is fine.

dmcclory:
Yeah, I'll give it a go! I'll see if I can think of a better name for the flag too.

alexellis:
Awesome. Make sure to join Slack too https://docs.openfaas.com/community

dmcclory:
Hello!

I've almost got a PR ready, it took me a little while to get the tests passing.

Before I put it up, I've also been thinking about the feedback on the name of the flag. I thought of a few alternatives, `--handler-dir` or `--handler`. (We could also have a short abbreviation like `-d` or `-H`)

alexellis:
Hi, glad you're making progress. I don't think you can use capital letters with Cobra for the short variant.

`--handler` sounds fine, maybe just stick with that for the first version/PR?

Alex

PR #504, opened 2018-08-16
Configurable handler directory

dmcclory:
This PR implements #481 

## Description
The fass-cli can now write the handler code to a directory that is different from the function name, by passing the `--handler` flag

## Motivation and Context

My original motivation was being able to define the handler code as a python-importable module, in order to write some tests for it. But I realized that other people might have other requirements, and this is flexible enough to accommodate them.

## How Has This Been Tested?

I added a new test case to `commands/new_function_test.go`.

## Types of changes

## Checklist:

### DIFF
--- a/commands/new_function.go
+++ b/commands/new_function.go
@@ -27,6 +27,7 @@ var (
 func init() {
 	newFunctionCmd.Flags().StringVar(&language, "lang", "", "Language or template to use")
 	newFunctionCmd.Flags().StringVarP(&gateway, "gateway", "g", defaultGateway, "Gateway URL to store in YAML stack file")
+	newFunctionCmd.Flags().StringVar(&handlerDir, "handler", "", "directory the handler will be written to")
 	newFunctionCmd.Flags().StringVarP(&imagePrefix, "prefix", "p", "", "Set prefix for the function image")
 
 	newFunctionCmd.Flags().BoolVar(&list, "list", false, "List available languages")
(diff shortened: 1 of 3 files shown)


dmcclory:
I think I just needed to rebase off the latest master. (It's "pretty nice" when that's all it takes!)

commands/new_function.go
```diff
@@ -140,23 +141,27 @@ func runNewFunction(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	if _, err := os.Stat(functionName); err == nil {
-		return fmt.Errorf("folder: %s already exists", functionName)
+	if len(handlerDir) == 0 {
+		handlerDir = functionName
```
martindekov: Hey there! I see only one problem here, every time we append new function with the current change we add `./` in front of the handler path so with the current change every time we use the `--append` flag in front of the `handler: ./<handler_folder>` in .yml file the `./` is appended so it becomes `././<handler_folder>` on 2nd append and `./././<handler_folder>` on 3rd append etc .. because if we look down we see `handler: ./{{ $function.Handler }}`.Got to find a way around that, but your code is very clean and good looking, great job!

alexellis: @martindekov is a change needed from @dmcclory ?

martindekov: Yes, with the normal faas-cli I don't have the same problem.

ewilde: Another way to go, is not to prefix the commandline `--handler` argument with `./` but default, if not given i.e.:
```golang
if len(handlerDir) == 0 {
	handlerDir = fmt.Sprintf("./%s", functionName)
}
```

And remove prefixing elsewhere i.e.

```golang
{{- range $name, $function := .Functions }}
  {{ $name }}:	  {{ $name }}:
    lang: {{ $function.Language }}
  handler: ./{{ $function.Handler }} <--- remove ./
    image: {{ $function.Image }}
{{- end }}	{{- end }}
```

```golang
if _, err := os.Stat("./" + dirName); os.IsNotExist(err) {      <--- remove ./
	t.Fatalf("%s/ directory was not created", dirName)
}
```
etc...

dmcclory: Does it matter if much if the `handler` value in the yaml doesn't start with a `./`?

I was thinking of doing this:

```
  {{- range $name, $function := .Functions }}
    {{ $name }}:
      lang: {{ $function.Language }}
      handler: {{ $function.Handler }}
      image: {{ $function.Image }}
  {{- end }}
```

and just adding the prefix here:
```
    function := stack.Function{
      Name:     functionName,
      Handler:  "./" + handlerDir,
      Language: language,
      Image:    imageName,
    }
```

alexellis:
Given the comments I don't think we're ready to merge for today's release.

alexellis:
Derek set milestone: 0.7.1

ewilde:
@alexellis @dmcclory happy to I think this commit https://github.com/openfaas/faas-cli/commit/e730b6b5f9dca02a5ada0cd5ab40aaa7c549da1b  message meets the guideline in the contrib guide https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md#commit-messages

dmcclory:
I've rebased it into 1 commit & updated the commit message using e730b6b as a model.

alexellis:
Awesome, thank you so much for doing that.

Once the chaps have had another look if they give the 👍 I'll merge ready for 0.7.1

alexellis:
Derek set milestone: 0.7.1

alexellis:
Thanks @martindekov. for re-testing. Seems like the other reviewers don't have time to comment or are OK with the changes.

@dmcclory thank you for your first PR to the faas-cli, I've merged the change and it will be live in 0.7.2

Alex

alexellis:
Derek set milestone: 0.7.2

PR #506, opened 2018-08-21
Get license-check via get.sh

alexellis:
Uses utility script which is multi-arch meaning we can use
one Dockerfile for varying platforms.


alexellis:
CI passed and tested locally, merging.

Issue #446, opened 2018-07-18
Add a new command to display a function information 


viveksyngh:
Add a new command in CLI which will show information related to a function (same as what we display in Gateway UI for a function)

## Expected Behaviour

Have a new command with the name `describe`/`get`/`inspect` which can be invoked as

```
faas-cli  describe  <function_name> -g <gateway>

```

which will display function information like `Status`, `Replicas`, `Invocations`, `Image`, `URL`

## Current Behaviour

NA

## Possible Solution

Invoke gateway service to get details of a function and display it. 

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* FaaS-CLI version ( Full output from: `faas-cli version` ):

* Docker version ( Full output from: `docker version` ):

* Are you using Docker Swarm (FaaS-swarm ) or Kubernetes (FaaS-netes)?

* Operating System and version (e.g. Linux, Windows, MacOS):

* Link to your project or a code example to reproduce issue:

burtonr:
How would the output look?

Would it be a table, similar to the `faas-cli list` command?

Or, should we go with a `json` object return? 
This may be more expected if using the `describe` command (as is the result with Swarm and K8s using that same command).

burtonr:
Derek add label: proposal

viveksyngh:
I think with table structure we will have the challenge in putting everything into one line in the terminal as `URL` length can be varied. 

We can display in `json` structure like `describe` or `inspect`.  

cc @alexellis

rgee0:
What the user knows:
the gateway address in order to deploy, which can bee obtained from `—version`
the function name, which can also be found in the existing list command.

The url of a function is http://$OPENFAAS_URL/function/<function-name>. With a variation for async.

What is the problem we are trying to address here? Ultimately understanding that will drive the solution.

alexellis:
I think this would act like the store and print in ASCII. Take a look at the store subcommand, there's something available like "info".

A table doesn't make sense because an argument would be necessary so we're only showing one item. Not sure what the idea about JSON was, but this is already available via the API.

alexellis:
I think Vivek's original intent was to get a URL for a specific function and some other data that we hold. Kind of like an info/inspect/get function in Docker or kubectl. Is that about right?

viveksyngh:
@alexellis Yes, Idea was to display the function URL and other information related to a function in CLI as we are displaying Gateway UI.  

[...]

PR #514, opened 2018-08-25
Add describe verb for a function

viveksyngh:
This commit adds a new verb `describe` to display information of an
OpenFaaS function.

Fixes: #446

## Motivation and Context

- [x] I have raised an issue to propose this change ([required]
(https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

issue #446 

## How Has This Been Tested?

This has been tested on my local system. 

```
./faas-cli describe haveibeenpwned
Name:			haveibeenpwned
Status:			Not Ready
Replicas:		1
Available replicas:	0
Invocations:		0
Image:			functions/haveibeenpwned:0.1
Function process:
URL:			http://127.0.0.1:8080/function/haveibeenpwned
Async URL:		http://127.0.0.1:8080/async-function/haveibeenpwned
➜  faas-cli git:(describe_verb) ✗ ./faas-cli describe env
Name:			env
Status:			Ready
Replicas:		1
Available replicas:	1
Invocations:		0
Image:			functions/alpine:latest
Function process:	env
URL:			http://127.0.0.1:8080/function/env
Async URL:		http://127.0.0.1:8080/async-function/env

./faas-cli describe figlet
Server returned unexpected status code: 404 -

```

## Types of changes

## Checklist:

### DIFF
--- a/commands/describe.go
+++ b/commands/describe.go
@@ -62,6 +62,19 @@ func runDescribe(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	//To get correct value for invocation count from /system/functions endpoint
+	functionList, err := proxy.ListFunctions(gatewayAddress, tlsInsecure)
+	if err != nil {
+		return err
+	}
+
+	var invocationCount float64
+	for _, fn := range functionList {
+		if fn.Name == function.Name {
+			invocationCount = fn.InvocationCount
+		}
+	}
+
 	var status string
 	if function.AvailableReplicas > 0 {
 		status = "Ready"
@@ -73,7 +86,7 @@ func runDescribe(cmd *cobra.Command, args []string) error {
 	fmt.Printf("%s:\t\t\t%s\n", "Status", status)
 	fmt.Printf("%s:\t\t%d\n", "Replicas", function.Replicas)
 	fmt.Printf("%s:\t%d\n", "Available replicas", function.AvailableReplicas)
-	fmt.Printf("%s:\t\t%v\n", "Invocations", function.InvocationCount)
+	fmt.Printf("%s:\t\t%v\n", "Invocations", invocationCount)
 	fmt.Printf("%s:\t\t\t%s\n", "Image", function.Image)
 	fmt.Printf("%s:\t%s\n", "Function process", function.EnvProcess)
 	fmt.Printf("%s:\t\t\t%s\n", "URL", getFunctionURL(gatewayAddress, functionName))
--- a/commands/describe.go
+++ b/commands/describe.go
@@ -1,0 +1,93 @@
+// Copyright (c) OpenFaaS Author(s) 2018. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package commands
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/openfaas/faas-cli/proxy"
+	"github.com/openfaas/faas-cli/stack"
+
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	describeCmd.Flags().StringVar(&functionName, "name", "", "Name of the function")
+	describeCmd.Flags().StringVarP(&gateway, "gateway", "g", defaultGateway, "Gateway URL starting with http(s)://")
+	describeCmd.Flags().BoolVar(&tlsInsecure, "tls-no-verify", false, "Disable TLS validation")
+	faasCmd.AddCommand(describeCmd)
+}
+
+var describeCmd = &cobra.Command{
+	Use:   "describe FUNCTION_NAME [--gateway GATEWAY_URL]",
+	Short: "Describe an OpenFaaS function",
+	Long:  `Display details of an OpenFaaS function`,
+	Example: `faas-cli describe figlet 
+faas-cli describe env --gateway http://127.0.0.1:8080
+faas-cli describe echo -g http://127.0.0.1.8080`,
+	PreRunE: preRunDescribe,
+	RunE:    runDescribe,
+}
+
+func preRunDescribe(cmd *cobra.Command, args []string) error {
+	return nil
+}
+
+func runDescribe(cmd *cobra.Command, args []string) error {
+	if len(args) < 1 {
+		return fmt.Errorf("please provide a name for the function")
+	}
+	var yamlGateway string
+	var services stack.Services
+	functionName = args[0]
+
+	if len(yamlFile) > 0 {
+		parsedServices, err := stack.ParseYAMLFile(yamlFile, regex, filter)
+		if err != nil {
+			return err
+		}
+
+		if parsedServices != nil {
+			services = *parsedServices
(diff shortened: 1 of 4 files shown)


rgee0:
I cant get a non-zero value for Invocations.  The UI says 3 the describe output says 0.

alexellis:
Perhaps we should ping both endpoints and munge together the results in the output then?

commands/describe.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		if parsedServices != nil {
+			services = *parsedServices
+			yamlGateway = services.Provider.GatewayURL
+		}
+	}
+	gatewayAddress := getGatewayURL(gateway, defaultGateway, yamlGateway, os.Getenv(openFaaSURLEnvironment))
+
+	function, err := proxy.GetFunctionInfo(gatewayAddress, functionName, tlsInsecure)
```
alexellis: Invocations are not returned from this endpoint, so we may want to overlay it with a call to `/system/functions`.

viveksyngh: Update it to get invocation count from `/system/functions`. is there any reason why we can not get correct value from `/system/function/<function-name>` endpoint ?

alexellis: Mainly because of the cost of combining data with Prometheus and how the status endpoint is used.

alexellis:
Please can one of you set the milestone for this using Derek. Pick the latest active one.

commands/describe.go
rgee0: How did invocations make it into the `/system/function/<function-name>` endpoint?  Having it always appearing as 0 in there could be a bit misleading for somebody.

alexellis:
We should try to set it to nil. Happy to take a PR. I suspect it's just because it uses the same or similar structure as the other endpoint?

commands/describe.go
```diff
@@ -0,0 +1,106 @@
...
+	//To get correct value for invocation count from /system/functions endpoint
+	functionList, err := proxy.ListFunctions(gatewayAddress, tlsInsecure)
+	if err != nil {
+		return err
+	}
+
+	var invocationCount float64
+	for _, fn := range functionList {
+		if fn.Name == function.Name {
+			invocationCount = fn.InvocationCount
```
rgee0: Should we break here?  We've got the value we want so why go through the rest of the values?

commands/describe.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+
+	var status string
+	if function.AvailableReplicas > 0 {
+		status = "Ready"
+	} else {
+		status = "Not Ready"
+	}
```
rgee0: Since we definitely need a value then we could save some assignments by:
```
status := "Not Ready"

if function.AvailableReplicas > 0 {
	status = "Ready"
}
```

proxy/describe.go
```diff
@@ -0,0 +1,61 @@
...
+	"time"
+
+	"github.com/openfaas/faas/gateway/requests"
+)
+
+//GetFunctionInfo get an OpenFaaS function information
+func GetFunctionInfo(gateway string, functionName string, tlsInsecure bool) (requests.Function, error) {
+	var result requests.Function
+
+	gateway = strings.TrimRight(gateway, "/")
```
rgee0: Should we use `url.Parse` here and then use `path.Join` to assemble the endpoint to ensure we get a clean url?

proxy/describe.go
```diff
@@ -0,0 +1,61 @@
...
+			return result, fmt.Errorf("cannot parse result from OpenFaaS on URL: %s\n%s", gateway, jsonErr.Error())
+		}
+	case http.StatusUnauthorized:
+		return result, fmt.Errorf("unauthorized access, run \"faas-cli login\" to setup authentication for this server")
+	default:
+		bytesOut, err := ioutil.ReadAll(res.Body)
+		if err == nil {
+			return result, fmt.Errorf("server returned unexpected status code: %d - %s", res.StatusCode, string(bytesOut))
+		}
+	}
```
rgee0: Are these the only two status codes we should expect?

commands/describe.go
```diff
@@ -0,0 +1,106 @@
...
+	fmt.Printf("%s:\t\t%s\n", "Async URL", getFunctionAsyncURL(gatewayAddress, functionName))
+
+	return nil
+}
+
+func getFunctionURL(gateway string, functionName string) string {
+	gateway = strings.TrimRight(gateway, "/")
+	return gateway + "/function/" + functionName
+}
+
```
rgee0: Since `getFunctionURL` and `getFunctionAsyncURL` effectively do the same thing, could we return sync and async from the same function?

commands/describe.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	fmt.Printf("%s:\t\t\t%s\n", "Name", function.Name)
+	fmt.Printf("%s:\t\t\t%s\n", "Status", status)
+	fmt.Printf("%s:\t\t%d\n", "Replicas", function.Replicas)
+	fmt.Printf("%s:\t%d\n", "Available replicas", function.AvailableReplicas)
+	fmt.Printf("%s:\t\t%v\n", "Invocations", invocationCount)
+	fmt.Printf("%s:\t\t\t%s\n", "Image", function.Image)
+	fmt.Printf("%s:\t%s\n", "Function process", function.EnvProcess)
+	fmt.Printf("%s:\t\t\t%s\n", "URL", getFunctionURL(gatewayAddress, functionName))
```
rgee0: Do we have a struct that would fit this data?  Could we populate that struct then use a method to output?  This could then be tested.

commands/describe.go
alexellis: @viveksyngh PTAL

commands/describe.go
alexellis: Please amend as per above

commands/describe.go
alexellis: At least in this version use a tab writer or construct a string instead of using `fmt.Printf`.

Look at the store commands for examples.

proxy/describe.go
viveksyngh: It is a `GET` request then I think this should be only two status codes we should check. Do you think any other HTTP codes need to be checked?

viveksyngh:
I have updated this PR to incorporate review comments.  Now output looks as below.

```
               Name: figlet
             Status: Ready
           Replicas: 1
 Available replicas: 1
        Invocations: 0
              Image: functions/figlet:0.1
   Function process:
                URL: http://127.0.0.1:8080/function/figlet
          Async URL: http://127.0.0.1:8080/async-function/figlet
```

proxy/describe.go
alexellis: Just add a default clause in a follow-up PR - you may get 502 or something else or a timeout.

alexellis:
I'm going to merge this and I'd like @viveksyngh to add anything else in a new PR. 

Alex

alexellis:
Derek set milestone: 0.7.3

PR #525, opened 2018-09-03
Fix intermittent travis build

viveksyngh:
This commit fixes intermittent Travis build issue for the CLI. It was
failing intermittently due to an intermittent error in CRD YAML test cases.

Updated test cases which causing this error because when
YAML structured in parsed with multiple function definition it can
produce output in different order.

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/commands/generate_test.go
+++ b/commands/generate_test.go
@@ -13,7 +13,7 @@ import (
 var generateTestcases = []struct {
 	Name       string
 	Input      string
-	Output     string
+	Output     []string
 	Format     schema.BuildFormat
 	APIVersion string
 	Namespace  string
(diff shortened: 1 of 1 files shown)


alexellis:
@rgee0 requesting your eyes.

rgee0:
Funnily enough I was just looking at this

commands/generate_test.go
```diff
@@ -170,9 +190,18 @@ func Test_generateCRDYAML(t *testing.T) {
...
 			t.Fatalf("%s failed: ouput is not as expected: %s", testcase.Name, generatedYAML)
 		}
 	}
 
 }
+
+func stringInSlice(a string, list []string) bool {
+	for _, b := range list {
+		if b == a {
+			return true
```
rgee0: I might be misreading this but aren't we likely to exit on the first item in the range?

I'd have thought we'd need to switch the logic and return false in the range iteration if `!=` and then return true at the end if we made it through the list?

Issue #527, opened 2018-09-10
Task: Test templating during CI


alexellis:
## Task:

Create a new function in a language such as python3 during CI and test that it works. Do not use the default handler, this must be overridden with some custom code to prove the templating engine is working end to end.

## Work:

Write a bash script to be called during CI to do the following:

- create a function using a template
- override the default message or handler code
- build using the cli
- run using Docker (not faas-cli deploy) with a set port
- access the port to make sure you get the correct value back

## Status

This is high priority because we had a regression last week in one of the CLI versions.

ivanayov:
Derek assign: me

alexellis:
Derek close: resolved

PR #528, opened 2018-09-11
Add integration test for templating during CI

ivanayov:

## Description

This change intends to prevent regression in the new CLI releases.
It makes a build using the CLI and tests how running container
from the function image behaves

## Motivation and Context

#527

## How Has This Been Tested?

`make test-integration`

Tested on Ubuntu and Mac OS.
Tested on ARM by @rdimitrov 

## Types of changes

- [x] New test

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ after_success:
 
 before_deploy:
   - make build_redist
+  - make test-templating
   - ./ci/hashgen.sh
 
 deploy:
--- a/Makefile
+++ b/Makefile
@@ -30,3 +30,11 @@ ci-armhf-push:
 ci-armhf-build:
 	(./build.sh $(TAG)-armhf)
 
+.PHONY: test-templating
+PORT?=38080
+FUNCTION?=templating-test-func
+FUNCTION_UP_TIMEOUT?=30
+.EXPORT_ALL_VARIABLES:
+test-templating:
+	./build_integration_test.sh
+
--- a/build_integration_test.sh
+++ b/build_integration_test.sh
@@ -1,0 +1,124 @@
+#!/bin/bash
+
+cli="./faas-cli"
+template="python3"
+
+get_package() {
+    uname=$(uname)
+
+    echo "Getting faas-cli package for $uname..."
+
+    case $uname in
+    "Darwin")
+    cli="./faas-cli-darwin"
+    ;;
+    "Linux")
+        arch=$(uname -m)
+        echo $arch
+        case $arch in
+        "armv6l" | "armv7l")
+        cli="./faas-cli-armhf"
+        template="python3-armhf"
+        ;;
+        esac
+    ;;
+    esac
+
+    echo "Using package $cli"
+    echo "Using template $template"
+}
+
+build_faas_function() {
+
+    function_name=$1
+
+    eval $cli new $function_name --lang $template
+
+cat << EOF > $function_name/handler.py
+def handle(req):
+
+    return "Function output from integration testing: Hello World!"
+EOF
+
+    eval $cli build -f $function_name.yml
+}
+
+wait_for_function_up() {
+    function_name=$1
+    port=$2
+    timeout=$3
+
+    function_up=false
+    for (( i=1; i<=$timeout; i++ ))
+    do
+        echo "Checking if 127.0.0.1:$port is up.. ${i}/$timeout"
+        status_code=$(curl 127.0.0.1:$port/_/health -o /dev/null -sw '%{http_code}')
+        if [ $status_code == 200 ] ; then
+            echo "Function $function_name is up on 127.0.0.1:$port"
+            function_up=true
+            break
+        else
+            echo "Waiting for function $function_name to run on 127.0.0.1:$port"
+            sleep 1
+        fi
+    done
+
+    if [ "$function_up" != true ] ; then
+        echo "Failed to reach function on 127.0.0.1:$port.. Service timeout"
+
+        echo "Removing container..."
+        docker rm -f $id
+
+        echo "Removing function image $function_name:latest"
+        docker rmi -f $function_name:latest
+
+        exit 1
+    fi
+}
+
+run_and_test() {
+    function_name=$1
+    port=$2
+    timeout=$3
+
+    if lsof -Pi :$port -sTCP:LISTEN -t >/dev/null ; then
+        echo "Port $port is already allocated.. Cannot use this port for testing.
+        Exiting test..."
+
+        echo "Removing function image $function_name:latest"
+        docker rmi -f $function_name:latest
+        exit 1
+    fi
+
+    id=$(docker run --env fprocess="python3 index.py" --name $function_name -p $port:8080 -d $function_name:latest)
+
+    wait_for_function_up $function_name $port $timeout
+
+    curl -s 127.0.0.1:$port > got.txt
+
+cat << EOF > want.txt
+Function output from integration testing: Hello World!
+EOF
+
+    if cmp got.txt want.txt ; then
+        echo "SUCCESS testing function $function_name"
+    else
+        echo "FAILED testing function $function_name"
+    fi
+
+    echo "Removing container..."
+    docker rm -f $id
+
+    echo "Removing function image $function_name:latest"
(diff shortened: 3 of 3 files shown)


build_integration_test.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/sh
+
+cli="./faas-cli"
+if [ $(uname) == 'Darwin' ] ; then
```
alexellis: What would happen on an ARMHF platform for this test?

build_integration_test.sh
```diff
@@ -0,0 +1,48 @@
...
+    cli="./faas-cli-darwin"
+fi
+
+eval $cli new btest --lang python3
```
alexellis: what does `btest` mean? Perhaps we can come up with a more descriptive name?

build_integration_test.sh
```diff
@@ -0,0 +1,48 @@
...
+def handle(req):
+
+    return "Function output from integration testing: Hello World!"
+EOF
+
+eval $cli build -f btest.yml
+
+docker run -d -p $PORT:$PORT --name btest btest:latest
+
+</dev/tcp/localhost/$PORT
```
alexellis: What does this line do?

build_integration_test.sh
```diff
@@ -0,0 +1,48 @@
...
+if [ $(echo $?) -eq 0 ] ; then
+    echo "Container btest successfully running on port $PORT"
+else
+    echo "Container btest is not running on port $PORT"
+    docker container stop btest
+    docker container rm btest
+    exit 1
+fi
+
+output=$(docker exec btest /bin/sh -c "echo '' | python3 index.py")
```
alexellis: This is not what I expected. Can't we visit the function via HTTP?

build_integration_test.sh
ivanayov: it's updated on the next lines

Makefile
```diff
@@ -30,3 +30,10 @@ ci-armhf-push:
 ci-armhf-build:
 	(./build.sh $(TAG)-armhf)
 
+.PHONY: test-integration
+PORT?=8083
```
alexellis: This port I think may clash with AlertManager or similar, how about using a port we know like 8080, then add a high letter such as `38080`? (just check that's a valid port first)

build_integration_test.sh
```diff
@@ -0,0 +1,41 @@
...
+    cli="./faas-cli-darwin"
+elif [ -f "/etc/debian_version" ] ; then
+    if [ $(dpkg --print-architecture) == 'armhf' ] ; then
+        cli="./faas-cli-armhf"
+    elif [ $(dpkg --print-architecture) == 'arm64' ] ; then
+        cli="./faas-cli-arm64"
+    fi
+fi
+
+eval $cli new $FUNCTION --lang python3
```
alexellis: Now that we've got things parameterized let's define a bash function for this whole block.

build_integration_test.sh
```diff
@@ -0,0 +1,41 @@
...
+
+id=$(docker run --env fprocess="python3 index.py" --name $FUNCTION -p 38080:8080 -d $FUNCTION:latest)
+
+curl -s 127.0.0.1:38080 > got.txt
+
+cat << EOF > want.txt
+Function output from integration testing: Hello World!
+EOF
+
+if cmp got.txt want.txt ; then
```
alexellis: Wonder if this should use the `[ ]` syntax? @ewilde can you review please as the resident bash expert?

Makefile
ivanayov: yep, I'm actually not using it, but forgot to edit.

build_integration_test.sh
ewilde: @alexellis I agree more consistent to stick with the [ ] syntax see: https://ryanstutorials.net/bash-scripting-tutorial/bash-if-statements.php#test

build_integration_test.sh
```diff
@@ -0,0 +1,59 @@
...
+    eval $cli build -f $function_name.yml
+}
+
+run_and_test() {
+    function_name=$1
+    port=$2
+
+    id=$(docker run --env fprocess="python3 index.py" --name $function_name -p $port:8080 -d $function_name:latest)
+
+    curl -s 127.0.0.1:$port > got.txt
```
ewilde: is the service guaranteed to be up at this point, what about on a slow machine like travis. I've seen in other CI openfaas scripts, error handling and waiting for services to start see: https://github.com/openfaas/faas/blob/master/contrib/ci.sh#L11

what do you think?

alexellis: Good point.. some kind of retry would be ideal here.

alexellis: You can curl to `/_/health` which will indicate `200` when ready.

ivanayov:
@ewilde @rdimitrov I addressed your comments + some additional checks and clean-up.
Can you please review?

rdimitrov:
All good from my side :+1:

build_integration_test.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
+
+cli="./faas-cli"
+template="python3"
+if [ $(uname) == 'Darwin' ] ; then
+    cli="./faas-cli-darwin"
+elif [ -f "/etc/debian_version" ] ; then
```
alexellis: What about the approach we use in the CLI get script? I think it might be more portable across platforms?

build_integration_test.sh
```diff
@@ -0,0 +1,98 @@
...
+
+        echo "Removing function image $function_name:latest"
+        docker rmi -f $function_name:latest
+        exit 1
+    fi
+
+    id=$(docker run --env fprocess="python3 index.py" --name $function_name -p $port:8080 -d $function_name:latest)
+
+    function_up=false
+    for (( i=1; i<=$timeout; i++ ))
```
alexellis: Does this work in place?

```
for i in {1...$timeout}
```

build_integration_test.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
+
+cli="./faas-cli"
+template="python3"
+if [ $(uname) == 'Darwin' ] ; then
```
alexellis: This should probably be a function too

build_integration_test.sh
```diff
@@ -0,0 +1,98 @@
...
+        docker rmi -f $function_name:latest
+        exit 1
+    fi
+
+    id=$(docker run --env fprocess="python3 index.py" --name $function_name -p $port:8080 -d $function_name:latest)
+
+    function_up=false
+    for (( i=1; i<=$timeout; i++ ))
+    do
+        echo "Checking if 127.0.0.1:$port is up.. ${i}/$timeout"
```
alexellis: Can this be a function too?

build_integration_test.sh
ivanayov: It's not working with variable, referring to the documentation and it failed on practice as well

build_integration_test.sh
ivanayov: It's addressed

ivanayov:
Thank you @rdimitrov 
Can I ask you for a final run on ARM with the comments addressed (as you already have the environment)?
I verified on Ubuntu and OsX

build_integration_test.sh
```diff
@@ -0,0 +1,124 @@
...
+    eval $cli build -f $function_name.yml
+}
+
+wait_for_function_up() {
+    function_name=$1
+    port=$2
+    timeout=$3
+
+    function_up=false
+    for (( i=1; i<=$timeout; i++ ))
```
alexellis: What happened to my suggestion here? I can't see my comment anymore

This would be simpler:

```
for i in {1...10}
```

.travis.yml
```diff
@@ -41,6 +41,7 @@ after_success:
 
 before_deploy:
   - make build_redist
+  - make test-integration
```
alexellis: `test-integration` should be something more specific such as `test-templating`

build_integration_test.sh
ivanayov: Yep, reply is hidden as outdated. 
`{1...$timeout}` does not work with variables, referring to the docs and after manual testing

PR #560, opened 2018-11-13
Update Go version to 1.11

alexellis:
Updates Go version to 1.11 

## Description

This is the current version of Golang, but could do with more exploratory testing from the community before merging. @stefanprodan may also have other input having commented about this version and static binaries in the past.

## Motivation and Context

https://github.com/openfaas/faas/issues/875

## How Has This Been Tested?

Unit tests pass and ran some exploratory tested with
build_redist.sh and binaries used on MacOS, Arch Linux (x86_64)
and Raspbian (armhf).

## Types of changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,7 @@
 sudo: required
 
-language: go
-
-go:
-  - 1.9.x
-install:
-  - echo "Please don't go get"
-  
+language: generic
+ 
 services:
   - docker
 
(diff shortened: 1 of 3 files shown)


alexellis:
They need to be moved to 3.8 per https://github.com/openfaas/faas/issues/875

stefanprodan:
I would change the go version in travis file also so that we'll use the same version everywhere.

alexellis:
I am wondering if we need the Go runtime at all for this build? 

https://github.com/openfaas/faas-cli/blob/master/.travis.yml

We seem to do everything in Docker.. @ewilde what do you think?

ewilde:
Yeah good point, worth testing on your feature branch without travis go support

alexellis:
It worked 👍

PR #576, opened 2018-12-28
Implement pinned template repo clone

### LINKED ISSUE #573: Allow pinning of templates to version via SHA/branch or tag (release)
## Expected Behaviour

I should be able to pin a template to a specific version via SHA/branch or tag (release) so that I can control my build pipeline and keep it reproducible.

## Current Behaviour

I can clone and checkout a templates repo manually, but not via `faas-cli template pull`.

## Possible Solution

```
faas-cli template pull https://cdn.com/repo.git#0.1.0

faas-cli template pull https://cdn.com/repo.git#dev

faas-cli template pull https://cdn.com/repo.git#d22af7cf74b7e24b37f46f8c60cb912751e93eab
```

If no `#` is given then the behaviour should default to the current functionality.

LucasRoesler:
Update `template pull` command to parse a git reference name from the
trailing hash/url fragment in the git url. This allows the user to pin the template to a specific revision or tag or even update from a specific branch.

## Description

- Add parsing utilities to check for a pinning value in the git url
- Updates git clone command to specify the branch flag
- Adds git command to validate the reference name
- Updates and adds unit tests for the new parsing

## Motivation and Context

Resolves #573 

## How Has This Been Tested?

I have manually tested it locally with a private template repository.   I have used `faas-cli template pull` with my private repository, with a tag and without a tag, as well as empty with the default OF template repo.

Additionally, unit tests have been added to validate the url parsing.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

Because the pin separator is `#`, this was originally validated as an valid URL but not parsed in any specific way. This fragment is not handled by the `git` CLI directly, but could possibly be handled by a remote git server.  It is important to note that GitHub does not handle a fragment in the url, it will cause a 404. By using the `#` character, it is possible that this could break for a user with a very special private git server configuration. The vast majority of users will not see any issues since most git servers can't parse or wont accept the url with a fragment. 

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,9 +100,9 @@ export OPENFAAS_URL=127.0.0.1:31112
 
 Advanced commands:
 
-* `faas-cli template pull` - pull in templates from a remote GitHub repository [Detailed Documentation](guide/TEMPLATE.md)
+* `faas-cli template pull` - pull in templates from a remote git repository [Detailed Documentation](guide/TEMPLATE.md)
 
-The default template URL of `https://github.com/openfaas/templates.git` can be overriden in two places including an environmental variable
+The default template URL of `https://github.com/openfaas/templates.git` can be overridden in two places including an environmental variable
 
 * 1st priority CLI input
 * 2nd priority `OPENFAAS_TEMPLATE_URL` environmental variable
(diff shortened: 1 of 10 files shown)


alexellis:
The tests look very thorough. Should we update the command text to give an example of using the `#` branch/tag reference?

LucasRoesler:
@alexellis just updated the relevant docs

alexellis:
@LucasRoesler please set the milestone for this feature as the next release via Derek.

alexellis:
Merged, thank you so much 🥇

LucasRoesler:
Derek set milestone: 0.8.3

Issue #577, opened 2018-12-30
Implement secret management commands


LucasRoesler:
## Expected Behaviour

The CLI needs to implement secret management commands as specified in openfaas/faas#807 These commands would talk to the OpenFaaS gateway that would then work with the configured provider to Create, Update, or List the secrets 

## Possible Solution

Implement a new `secrets` command with the following subcommands

```sh
$ faas secret create --name secret-name --value secret-value
Secret `secret-name` created.
$ faas secret delete secret-name
Secret `secret-name` deleted.
$ faas secret replace secret-name --value new-secret-value
Secret `secret-name` updated.
$ faas secret list
secret-name-1
secret-name-2
secret-name-3
```

Some other ideas to consider

1. Support a `--from-file` flag: `faas secret create --name secret-name --from-file /path/to/secret/file`
2. The List command could have a verbose mode with extra information such as creation/modification date

## Context

This will help simplify end user deployment of secure functions and simplify documentation such as 
- [Lab 10 - Advanced feature - Secrets](https://github.com/openfaas/workshop/blob/master/lab10.md)
- [OpenFaaS Docs: Secrets](https://docs.openfaas.com/reference/secrets/)

viveksyngh:
Derek assign me

alexellis:
Derek set milestone: 0.8.3

alexellis:
I'm happy for this work to be broken up - i.e. for someone who has time now to do an initial PR to add the "faas-cli secret list` command first and get that merged.

LucasRoesler:
I would like to do this one

Derek assign: me

alexellis:
Go for it

viveksyngh:
I have started work on this should be able to push initial version today.

alexellis:
Can you both work together on Slack do coordinate?

ivanayov:
@viveksyngh @LucasRoesler @bartsmykla did someone started working on `create`/`delete`/`replace`? Can I take one of them?

bartsmykla:
I didn't

alexellis:
Please check with Vivek in the members channel. He has an open PR already which should be linked.

leodido:
Would like to help implementing one command.

alexellis:
I would say whoever does invoke needs to do replace/update too since it will be the same code, let's not duplicate effort on that one. @viveksyngh did you say you've started this?

I've just put a PR up for `delete/remove`

Alex

viveksyngh:
[...]

PR #578, opened 2019-01-05
Add secret list command to faas-cli

viveksyngh:
This commit adds `faas-cli secret list` command to the faas-cli which
will be used to get secret list from the cluster.

## Motivation and Context

#577 

## How Has This Been Tested?

Tested on local Docker for Mac with faas-swarm.

```
./faas-cli secret ls -g http://127.0.0.1:8081

NAME
test
```

## Types of changes

## Checklist:

### DIFF
--- a/commands/secret_list.go
+++ b/commands/secret_list.go
@@ -4,10 +4,13 @@
 package commands
 
 import (
+	"bytes"
 	"fmt"
 	"os"
+	"text/tabwriter"
 
 	"github.com/openfaas/faas-cli/proxy"
+	"github.com/openfaas/faas-cli/schema"
 	"github.com/spf13/cobra"
 )
 
@@ -15,7 +18,7 @@ import (
 var secretListCmd = &cobra.Command{
 	Use:   "list",
 	Short: "List all secrets",
-	Long:  `List all secrets names and metadata`,
+	Long:  `List all secrets`,
 	Example: `faas-cli secret list
 faas-cli secret list --gateway=http://127.0.0.1:8080`,
 	RunE:    runSecretList,
@@ -53,9 +56,26 @@ func runSecretList(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	for _, secret := range secrets {
-		fmt.Println(secret.Name)
+	if len(secrets) == 0 {
+		fmt.Println("No secret found.")
+		return nil
 	}
+	fmt.Print(renderSecretList(secrets))
 
 	return nil
+}
+
+func renderSecretList(secrets []schema.Secret) string {
+	var b bytes.Buffer
+	w := tabwriter.NewWriter(&b, 0, 0, 1, ' ', 0)
+	fmt.Fprintln(w)
+	fmt.Fprintln(w, "NAME")
+
+	for _, secret := range secrets {
+		fmt.Fprintf(w, "%s\n", secret)
+	}
+
+	fmt.Fprintln(w)
+	w.Flush()
+	return b.String()
 }
--- a/commands/secret_list.go
+++ b/commands/secret_list.go
@@ -72,7 +72,7 @@ func renderSecretList(secrets []schema.Secret) string {
 	fmt.Fprintln(w, "NAME")
 
 	for _, secret := range secrets {
-		fmt.Fprintf(w, "%s\n", secret)
+		fmt.Fprintf(w, "%s\n", secret.Name)
 	}
 
 	fmt.Fprintln(w)
--- a/commands/secret_list.go
+++ b/commands/secret_list.go
@@ -1,0 +1,61 @@
+// Copyright (c) OpenFaaS Author(s) 2018. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package commands
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/openfaas/faas-cli/proxy"
+	"github.com/spf13/cobra"
+)
+
+// secretListCmd represents the secretCreate command
+var secretListCmd = &cobra.Command{
+	Use:   "list",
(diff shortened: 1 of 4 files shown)


LucasRoesler:
I just tested this with I also tested this with the latest build of gateway (openfaas/faas@a65df4795bc66147c41161c48bfd4c72f60c7434) and openfaas-incubator/openfaas-operator#70 and it worked as expected 

```bash
$ ./faas-cli-darwin secret list --gateway=http://localhost:31112
test
test2
```

I still need to read through the code

alexellis:
Should we have a heading like with the other commands?

viveksyngh:
What are the other informations that we want to display here ?

LucasRoesler:
At the moment there is no other information besides `name`.  Right now the gateway won't expose any other information. We could add more fields in the future, once we know they are needed and supported.

proxy/secret.go
```diff
@@ -0,0 +1,60 @@
...
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/openfaas/faas-cli/schema"
+)
+
+//GetSecretList get secrets list
+func GetSecretList(gateway string, tlsInsecure bool) ([]schema.Secret, error) {
+	var results []schema.Secret
```
LucasRoesler: We probably want a "warn insecure" block here like we do in the deploy code https://github.com/openfaas/faas-cli/blob/200359ce2237dd8b55e156683957c44ab26b2ef9/proxy/deploy.go#L81-L85

alexellis:
I think we should have a heading of Name.

alexellis:
All the commands in the future should have some kind of --no-header flag.

LucasRoesler:
@alexellis Sorry for being dense, when you say `--no-header`, you mean something like `docker images -q` so that the "Name" header is excluded from the output, right?

alexellis:
Yes 👍

proxy/secret.go
alexellis: +1 @viveksyngh

commands/secret_list.go
```diff
@@ -0,0 +1,61 @@
...
+
+func runSecretList(cmd *cobra.Command, args []string) error {
+	var gatewayAddress string
+	gatewayAddress = getGatewayURL(gateway, defaultGateway, "", os.Getenv(openFaaSURLEnvironment))
+
+	secrets, err := proxy.GetSecretList(gatewayAddress, tlsInsecure)
+	if err != nil {
+		return err
+	}
+
```
alexellis: Can we build up a string instead? Have a look how it's done in other commands. This will allow us to test the output.

Please also add the header of "Name"

commands/secret_list.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/openfaas/faas-cli/proxy"
+	"github.com/spf13/cobra"
+)
+
+// secretListCmd represents the secretCreate command
+var secretListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List all secrets",
+	Long:  `List all secrets names and metadata`,
```
alexellis: We have no metadata now, so omit this.

commands/secret_list.go
```diff
@@ -0,0 +1,81 @@
...
+	if len(secrets) == 0 {
+		fmt.Println("No secret found.")
+		return nil
+	}
+	fmt.Print(renderSecretList(secrets))
+
+	return nil
+}
+
+func renderSecretList(secrets []schema.Secret) string {
```
alexellis: Nicely done

proxy/secret.go
```diff
@@ -0,0 +1,66 @@
...
+
+	if tlsInsecure {
```
alexellis: I think we only show this if tlsInsecure is off? @LucasRoesler?

proxy/secret.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot connect to OpenFaaS on URL: %s", gateway)
+	}
+
+	if res.Body != nil {
+		defer res.Body.Close()
+	}
+
+	switch res.StatusCode {
+	case http.StatusOK:
```
alexellis: Please add http.StatusAccepted too.

alexellis:
This is looking great now Vivek. I did add one query on the tlsInsecure behavior though.. do you get a warning when you run this against http://localhost ?

proxy/secret.go
LucasRoesler: yeah, I think the check should be inverted here.  Relatedly, the  doc string on this function should document the args to make it clearer how `tlsInsecure` should behave

viveksyngh:
I will update it, I could not get version running locally which gives `secrets` endpoint. @LucasRoesler 
 can you help me with the testing?

proxy/secret.go
viveksyngh: will add it. just a doubt will`GET` method return status`http.StatusAccepted`?

commands/secret_list.go
```diff
@@ -0,0 +1,81 @@
...
+	"text/tabwriter"
+
+	"github.com/openfaas/faas-cli/proxy"
+	"github.com/openfaas/faas-cli/schema"
+	"github.com/spf13/cobra"
+)
+
+// secretListCmd represents the secretCreate command
+var secretListCmd = &cobra.Command{
+	Use:   "list",
```
ivanayov: I think it's good to have the `ls` alias as well, because it's common for most commands and people are used to using it

viveksyngh:
Tested on local Docker for Mac with `faas-swarm`. 

```
./faas-cli secret ls -g http://127.0.0.1:8081

NAME
test
```

alexellis:
(Which commands do we want the warning for? All of them or just where we may be writing confidential data?)

PR #580, opened 2019-01-10
Add create command for secrets

viveksyngh:
This commit adds command to create new secrets from the CLI

## Motivation and Context

#577 

## How Has This Been Tested?

Tested on my local for docker for mac and Kubernetes
```
./faas-cli secret create test-secret --value=test
Creating secret: test-secret

server returned unexpected status code: 500 - secrets "test-secret" already exists

./faas-cli secret create test-secret --from-file ~/Downloads/openfaas-cloud.2018-04-11.private-key.pem
Creating secret: test-secret

server returned unexpected status code: 500 - secrets "test-secret" already exists

 ./faas-cli secret create of-cloud --from-file=/Users/viveks/Downloads/openfaas-cloud.2018-04-11.private-key.pem
Creating secret: of-cloud

Created: 202 Accepted

./faas-cli secret create cli-test --value=test
Creating secret: cli-test

Created: 202 Accepted

/faas-cli secret create of-cloud --from-file=/Users/viveks/Downloads/openfaas-cloud.2018-04-11.private-key.pem --value=test
Please either provide value or from-file for the secret

./faas-cli secret list

NAME
cli-test
of-cloud
test-secret


./faas-cli secret create  --from-file=/Users/viveks/Downloads/openfaas-cloud.2018-04-11.private-key.pem
Please provide secret name

```

## Types of changes

## Checklist:

### DIFF
--- a/commands/secret_create.go
+++ b/commands/secret_create.go
@@ -63,6 +63,11 @@ func runSecretCreate(cmd *cobra.Command, args []string) error {
 	}
 
 	if len(secretValue) == 0 && len(secretFile) == 0 {
+		stat, _ := os.Stdin.Stat()
+		if (stat.Mode() & os.ModeCharDevice) != 0 {
+			fmt.Fprintf(os.Stderr, "Reading from STDIN - hit (Control + D) to stop.\n")
+		}
+
 		secretStdin, err := ioutil.ReadAll(os.Stdin)
 		if err != nil {
 			return err
--- a/commands/secret_create.go
+++ b/commands/secret_create.go
@@ -14,9 +14,8 @@ import (
 )
 
 var (
-	secretName  string
-	secretValue string
-	secretFile  string
+	literalSecret string
+	secretFile    string
 )
 
 // secretCreateCmd represents the secretCreate command
@@ -33,7 +32,7 @@ cat /path/to/secret/file | faas-cli secret create secret-name`,
 }
 
 func init() {
-	secretCreateCmd.Flags().StringVar(&secretValue, "from-literal", "", "Value of the secret")
+	secretCreateCmd.Flags().StringVar(&literalSecret, "from-literal", "", "Value of the secret")
 	secretCreateCmd.Flags().StringVar(&secretFile, "from-file", "", "Path to the secret file")
 	secretCreateCmd.Flags().BoolVar(&tlsInsecure, "tls-no-verify", false, "Disable TLS validation")
 	secretCreateCmd.Flags().StringVarP(&gateway, "gateway", "g", defaultGateway, "Gateway URL starting with http(s)://")
@@ -42,27 +41,37 @@ func init() {
 
 func preRunSecretCreate(cmd *cobra.Command, args []string) error {
 
+	if len(secretFile) > 0 && len(literalSecret) > 0 {
+		return fmt.Errorf("please provide secret using only one option from --from-literal, --from-file and STDIN")
+	}
+
 	return nil
 }
 
 func runSecretCreate(cmd *cobra.Command, args []string) error {
-	var gatewayAddress string
-	var err error
+	var (
+		gatewayAddress string
+		err            error
+		secretValue    string
+	)
 
 	if len(args) < 1 {
 		return fmt.Errorf("please provide secret name")
 	}
-	secretName = args[0]
 
-	if len(secretValue) == 0 && len(secretFile) > 0 {
+	secretName := args[0]
+
+	switch {
+	case len(literalSecret) > 0:
+		secretValue = literalSecret
+
+	case len(secretFile) > 0:
 		secretValue, err = readSecretFromFile(secretFile)
 		if err != nil {
 			return err
 		}
 
-	}
-
-	if len(secretValue) == 0 && len(secretFile) == 0 {
+	default:
 		stat, _ := os.Stdin.Stat()
 		if (stat.Mode() & os.ModeCharDevice) != 0 {
 			fmt.Fprintf(os.Stderr, "Reading from STDIN - hit (Control + D) to stop.\n")
@@ -72,9 +81,9 @@ func runSecretCreate(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
-
 		secretValue = string(secretStdin)
 	}
+
 	secretValue = strings.TrimSpace(secretValue)
 
 	if len(secretValue) == 0 {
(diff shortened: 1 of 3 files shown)


commands/secret_create.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func init() {
+	secretCreateCmd.Flags().StringVar(&secretValue, "value", "", "Value of the secret")
+	secretCreateCmd.Flags().StringVar(&secretFile, "from-file", "", "Path to the secret file")
+	secretCreateCmd.Flags().BoolVar(&tlsInsecure, "tls-no-verify", false, "Disable TLS validation")
+	secretCreateCmd.Flags().StringVarP(&gateway, "gateway", "g", defaultGateway, "Gateway URL starting with http(s)://")
+	secretCmd.AddCommand(secretCreateCmd)
+
+	// Here you will define your flags and configuration settings.
```
alexellis: Let's remove this comment

commands/secret_create.go
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) OpenFaaS Author(s) 2018. All rights reserved.
```
alexellis: Can we put 2019?

alexellis:
Good start. Just sent some comments.

alexellis:
1) So if you do:

```
faas-cli secret create api-key
cat ~/Downloads/derek.pem | faas-cli secret create api-key
```

It should read from stdin. See how this is done in `faas-cli invoke` as an example

2) If you want to pass a value then do:

```
faas-cli secret create api-key --from-literal="04385e5c413c10ed68afb010ebe8c5dd706aa20a"
```

3) And finally for files:

```
faas-cli secret create api-key --from-file=~/Downloads/derek.pem
```

viveksyngh:
Got it .. what should be the preference if user provides more than option, I know it is unlikely to option but just a thought ?

LucasRoesler:
@viveksyngh when the cmd reads from stdin, i wonder if we should include a comment line that instructs the user to `CTRL-D` to indicate then end of the content? For example,

```bash
./faas-cli secret create of-cloud-1
Type CTRL-D to end input
hello
Creating secret: of-cloud-1
```

alexellis:
Yes we should, that's what I meant by looking at the invoke command originally. We have a sort of hybrid between the invoke and login commands.

Do we also need a `--trim` flag to be used with this command?

commands/secret_create.go
```diff
@@ -0,0 +1,91 @@
...
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/openfaas/faas-cli/proxy"
+	"github.com/spf13/cobra"
+)
+
+var (
+	secretName  string
```
leodido: Imho this is better declared within the `runSecretCreate` function (the only place where it is used).

ivanayov:
@viveksyngh please see #582 so not to duplicate efforts

commands/secret_create.go
```diff
@@ -0,0 +1,96 @@
...
+
+	if len(secretValue) == 0 && len(secretFile) > 0 {
+		secretValue, err = readSecretFromFile(secretFile)
+		if err != nil {
+			return err
+		}
+
+	}
+
+	if len(secretValue) == 0 && len(secretFile) == 0 {
```
ivanayov: What will happen if someone passes both from literal and file?
This is possible with the flags, but we don’t cover the use-case.

commands/secret_create.go
```diff
@@ -0,0 +1,96 @@
...
+	var gatewayAddress string
+	var err error
+
+	if len(args) < 1 {
+		return fmt.Errorf("please provide secret name")
+	}
+	secretName = args[0]
+
+	if len(secretValue) == 0 && len(secretFile) > 0 {
+		secretValue, err = readSecretFromFile(secretFile)
```
ivanayov: I’m not 100% sure but AFAIR in Kubernetes the file name was taken into account. Didn’t check the documentation, but remember that testing Derek was once failing because the secret file wasn’t named as shown in the docs.

commands/secret_create.go
alexellis: function or file?

LucasRoesler: I agree, each of these variables should be local variables of the `runSecretCreate` method, similar to the `gatewayAddress` variable.

commands/secret_create.go
LucasRoesler: I am not sure I understand the comment. Reading the value form a file here in the CLI shouldn't impact how it is created in the provider.

commands/secret_create.go
LucasRoesler: This seems like a good place for a `switch` (i would also suggest changing the name fo the variable for the `from-literal` flag to something like `literalSecret`)

```golang
var secretValue string
switch {
case len(literalSecret) > 0:
    // handle input
case len(secretFile) > 0:
   // handle parsing from the file
default:
   // handle parsing from stdin
}
```

This will set a clear order of precedence and also make it clear that reading directly from stdin is the default behavior

commands/secret_create.go
alexellis: We haven't done this anywhere else, so I'm not considering this a blocker. Perhaps a nice refactoring exercise for someone?

LucasRoesler: ok, sounds good, i will create a ticket

commands/secret_create.go
alexellis: I think what Ivana means is that if you do --from-file - the name of the secret is usually the name of the file. Here we just allow that to be overridden

commands/secret_create.go
alexellis: Seems fair as a comment. Can it be done as a refactoring exercise?

commands/secret_create.go
viveksyngh: Got it, that specific to kubernetes? I think for swarm we need to pass the secret name.

commands/secret_create.go
viveksyngh: Yeah, that can be done. I had already left a comment what should be preference between different flags ?

LucasRoesler:
I just tested this locally and everything checked out

- passing the value in multiple ways produced a validation error
- each input method worked as expected

I tested against the openfaas-operator provider

proxy/secret.go
```diff
@@ -66,7 +66,7 @@ func GetSecretList(gateway string, tlsInsecure bool) ([]schema.Secret, error) {
 	return results, nil
 }
 
-// RemoveSecret remove a secret via the OpenFaaS API by name
+//RemoveSecret remove a secret via the OpenFaaS API by name
```
alexellis: What tool is removing the space between the comment and `//`? I use vscode and its built-in formatters that keep putting it back.

Issue #581, opened 2019-01-10
Refactor TLS warning into const


alexellis:
## Expected Behaviour

The following string is repeated in several places, especially when we merge all the remaining secrets features such as #580 

```
"WARNING! Communication is not secure, please consider using HTTPS. Letsencrypt.org offers free SSL/TLS certificates.
```

## Possible Solution

We should extract these messages into a const within the package they live in. If they live in both, then maybe they should go in a common existing package?

viveksyngh:
Like this there are few more error messages that we can refactor, they are mostly part of the proxy functions.

mario-mazo:
I would like to take this one

PR #585, opened 2019-01-12
Moved TLS warnings to its own package

mario-mazo:
Signed-off-by: Mario Mazo <buster.mazo@gmail.com>

Moved TLS warnings to its own package to solve #581 

## Description
I have moved the TLS warnings to its own package. I have chosen the root level because
the message is spread across several packages. Also I have used the `faaserrors` instead of
the more widely used _errors_ to avoid confusion.

## Motivation and Context
avoiding updating error/warning messages in several places

## How Has This Been Tested?
Tested against OpenFaaS deployed in EKS. CLI seems to work as expected

## Types of changes

## Checklist:

### DIFF
--- a/commands/login.go
+++ b/commands/login.go
@@ -11,9 +11,9 @@ import (
 	"strings"
 	"time"
 
-	"github.com/openfaas/faas-cli/proxy"
-
 	"github.com/openfaas/faas-cli/config"
+	"github.com/openfaas/faas-cli/faaserrors"
+	"github.com/openfaas/faas-cli/proxy"
 	"github.com/spf13/cobra"
 )
 
@@ -118,7 +118,7 @@ func validateLogin(gatewayURL string, user string, pass string) error {
 	}
 
 	if res.TLS == nil {
-		fmt.Println("WARNING! Communication is not secure, please consider using HTTPS. Letsencrypt.org offers free SSL/TLS certificates.")
+		fmt.Println(faaserrors.NoTLSWarn)
 	}
 
 	switch res.StatusCode {
--- a/faaserrors/errors.go
+++ b/faaserrors/errors.go
@@ -1,0 +1,6 @@
+package faaserrors
+
+const (
+	// NoTLSWarn Warning thrown when no SSL/TLS is used
+	NoTLSWarn = "WARNING! Communication is not secure, please consider using HTTPS. Letsencrypt.org offers free SSL/TLS certificates."
+)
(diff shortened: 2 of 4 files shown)


commands/login.go
```diff
@@ -11,9 +11,9 @@ import (
 	"strings"
 	"time"
 
-	"github.com/openfaas/faas-cli/proxy"
-
 	"github.com/openfaas/faas-cli/config"
+	"github.com/openfaas/faas-cli/faaserrors"
```
LucasRoesler: I think it makes sense to just name the package `errors` and then import at 

```
faaserrors "github.com/openfaas/faas-cli/errors"
```

or

```
clierrors "github.com/openfaas/faas-cli/errors"
```

especially since the only error is not a generic OpenFaaS error statement but a specific cli error

faaserrors/errors.go
```diff
@@ -0,0 +1,6 @@
+package faaserrors
```
LucasRoesler: The name of this file really should be `warnings.go` or `warn.go` since the only variable is not of error type.

commands/login.go
mario-mazo: @LucasRoesler  clierrors seems like a more sensitive name. Will change

alexellis:
Do we need a new package for one string? If it's only consumed by the proxy package then can we leave it there.

mario-mazo:
@alexellis its actually in 2 packages (proxy and command), we could copy paste... that's also not nice

alexellis:
I'm curious if we need it in both?

alexellis:
I think I'd rather see all the warnings in the commands package and the const kept there. The proxy should be reusable since it's public. 

Alex

mario-mazo:
@alexellis sorry. I am not sure if I understand you correctly. so the idea is to:

- Create the warning const in the commands pkg
- In proxy pkg call `commands.NoTLSWarn` ?

alexellis:
I'm not sure that the warnings belong in the proxy package at all because it could be used by a third party in their code. Therefore it fits better in the command package.

mario-mazo:
@alexellis After checking the proxy pkg in more detail I see 2 things:
in `secret.go` there is 4 times this exact code

```go
if !tlsInsecure {
		if !strings.HasPrefix(gateway, "https") {
			fmt.Println("WARNING! Communication is not secure, please consider using HTTPS. Letsencrypt.org offers free SSL/TLS certificates.")
		}
	}
```

and then there is an 1 check in `deploy.go` in `Deploy()` 

so Does it make sense for you if we move the checks from proxy/secret and proxy/deploy  to the command pkg

also that `if !tlsInsecure` reads weird I would rather read it `if tlsInsecure` or even `if !tlsSecure`

PR #593, opened 2019-01-27
Proxy TLS warnings refactoring

### LINKED ISSUE #581: Refactor TLS warning into const

mario-mazo:
Proxy TLS warnings refactoring

## Description

I have made all the TLS warnings from the proxy pkg to the commands. pkg

## Motivation and Context

Having the TLS checks in the proxy made it more difficult to reuse. So having the TLS checks on the client is more flexible. Also there was some code duplicating we could simplify

fixes #581 
- [x ] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

## How Has This Been Tested?

deploy 
```
Deploying: hello-fork.

Deployed. 202 Accepted.
URL: http://127.0.0.1:8080/function/hello-fork
```

list
```
$ faas-cli secret ls

NAME
```
## Types of changes

## Checklist:

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -14,16 +14,14 @@ import (
 	"path/filepath"
 	"strings"
 
-	"gopkg.in/yaml.v2"
-
 	"github.com/docker/docker-credential-helpers/client"
 	homedir "github.com/mitchellh/go-homedir"
 	"github.com/openfaas/faas-cli/builder"
 	"github.com/openfaas/faas-cli/proxy"
 	"github.com/openfaas/faas-cli/schema"
 	"github.com/openfaas/faas-cli/stack"
-
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
 )
 
 // DeployFlags holds flags that are to be added to commands.
@@ -283,6 +281,7 @@ Error: %s`, fprocessErr.Error())
 				TLSInsecure:             tlsInsecure,
 			}
 
+			fmt.Println(checkTLSInsecure(deploySpec.Gateway, deploySpec.TLSInsecure))
 			statusCode := proxy.DeployFunction(deploySpec)
 
 			if badStatusCode(statusCode) {
@@ -376,6 +375,7 @@ func deployImage(
 		TLSInsecure:             tlsInsecure,
 	}
 
+	fmt.Println(checkTLSInsecure(deploySpec.Gateway, deploySpec.TLSInsecure))
 	statusCode = proxy.DeployFunction(deploySpec)
 
 	return statusCode, nil
--- a/commands/errors.go
+++ b/commands/errors.go
@@ -1,0 +1,24 @@
+// Copyright (c) OpenFaaS Author(s) 2019. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package commands
+
+import (
+	"strings"
+)
+
+const (
+	// NoTLSWarn Warning thrown when no SSL/TLS is used
+	NoTLSWarn = "WARNING! Communication is not secure, please consider using HTTPS. Letsencrypt.org offers free SSL/TLS certificates."
+)
+
+// checkTLSInsecure returns a warning message if the given gateway does not have https.
+// Use tsInsecure to skip validations
+func checkTLSInsecure(gateway string, tlsInsecure bool) string {
+	if !tlsInsecure {
+		if !strings.HasPrefix(gateway, "https") {
+			return NoTLSWarn
+		}
+	}
+	return ""
+}
--- a/commands/secret_list.go
+++ b/commands/secret_list.go
@@ -40,6 +40,7 @@ func preRunSecretListCmd(cmd *cobra.Command, args []string) error {
 func runSecretList(cmd *cobra.Command, args []string) error {
 	var gatewayAddress string
 	gatewayAddress = getGatewayURL(gateway, defaultGateway, "", os.Getenv(openFaaSURLEnvironment))
+	fmt.Println(checkTLSInsecure(gatewayAddress, tlsInsecure))
 
 	secrets, err := proxy.GetSecretList(gatewayAddress, tlsInsecure)
 	if err != nil {
--- a/commands/errors_test.go
+++ b/commands/errors_test.go
@@ -1,0 +1,29 @@
+// Copyright (c) OpenFaaS Author(s) 2019. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+package commands
+
+import "testing"
+
+func Test_checkTLSInsecure(t *testing.T) {
+	type args struct {
+		gateway     string
+		tlsInsecure bool
+	}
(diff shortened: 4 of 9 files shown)


commands/errors.go
```diff
@@ -0,0 +1,24 @@
...
+	"strings"
+)
+
+const (
+	// NoTLSWarn Warning thrown when no SSL/TLS is used
+	NoTLSWarn = "WARNING! Communication is not secure, please consider using HTTPS. Letsencrypt.org offers free SSL/TLS certificates."
+)
+
+// checkTLSInsecure checks if the given gateway has https. Use tsInsecure to skip validations
+// Note: This function does not stop the flow only prints a warning
```
LucasRoesler: This note is slightly inaccurate, it does not actually print anything.

commands/errors.go
```diff
@@ -0,0 +1,24 @@
...
+func checkTLSInsecure(gateway string, tlsInsecure bool) string {
+	if !tlsInsecure {
+		if !strings.HasPrefix(gateway, "https") {
+			return (NoTLSWarn)
```
LucasRoesler: We can remove the parentheses 
```suggestion
			return NoTLSWarn
```

commands/secret_list.go
```diff
@@ -40,6 +40,7 @@ func preRunSecretListCmd(cmd *cobra.Command, args []string) error {
 func runSecretList(cmd *cobra.Command, args []string) error {
 	var gatewayAddress string
 	gatewayAddress = getGatewayURL(gateway, defaultGateway, "", os.Getenv(openFaaSURLEnvironment))
+	fmt.Print(checkTLSInsecure(gatewayAddress, tlsInsecure), "\n")
```
LucasRoesler: use `fmt.Println()`
```suggestion
	fmt.Println(checkTLSInsecure(gatewayAddress, tlsInsecure))
```

commands/errors.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+const (
+	// NoTLSWarn Warning thrown when no SSL/TLS is used
+	NoTLSWarn = "WARNING! Communication is not secure, please consider using HTTPS. Letsencrypt.org offers free SSL/TLS certificates."
+)
+
+// checkTLSInsecure checks if the given gateway has https. Use tsInsecure to skip validations
+// Note: This function does not stop the flow only prints a warning
+func checkTLSInsecure(gateway string, tlsInsecure bool) string {
```
LucasRoesler: You could add the print statement here and make it testable by using the `io.Writer` interface.  It would look like this

```go
func warnTLSInsecure(out io.Writer, gateway string, tlsInsecure bool) {
	if !tlsInsecure {
		if !strings.HasPrefix(gateway, "https") {
			fmt.Fprintln(out, NoTLSWarn)
		}
	}
}
```

All of the usages would then look like 

```go
warnTLSInsecure(cmd.OutOrStdout(), gatewayAddress, tlsInsecure)
```

No need for print statements.

This is still testable because the `bytes.Buffer` implements `io.Writer`, so the test look like

```go
buf := &bytes.Buffer{}

warnTLSInsecure(buf, tt.args.gateway, tt.args.tlsInsecure)
got := buf.String()

if got != tt.want {
    t.Errorf("checkTLSInsecure() = %v, want %v", got, tt.want)
}
```

Also, important, Cobra commands also have `cmd.OutOrStderr()` which behaves the same way but the output goes to stderr, as implied.  These helpers are useful because it ensures that you get the right output buffers, even if it is being configured to go somewhere else by the user in someway.

alexellis: I agree with this, but would see it as part of a follow-up change. The scope of this PR is only to move the location of the warnings. @mario-mazo is also a new contributor to the project so if we work together more incrementally I think it would make the experience smoother?

LucasRoesler: i agree, we can definitely take this as next steps.  The other three comments are smaller and would be good changes though, I think.

commands/secret_list.go
mario-mazo: oh the `Println` was a mistake I had something different and forgot to revert back to `println` I will fix

commands/errors.go
mario-mazo: I was not aware of `cmd.OutOrStdout()` I will implement this

alexellis: Please don't.

This PR is only to move the message, not to refactor the code. Do that in a follow-up PR.

commands/secret_list.go
alexellis: 👍

commands/errors.go
alexellis: I'd suggest:

// checkTLSInsecure will return a warning if the gateway URL is not using TLS and the
// tlsInsecure flag is set to false.

alexellis:
Let's squash both commits into one please since we didn't accept the original solution, there's no point having it in Git history.

mario-mazo:
squashed and pushed

commands/deploy.go
```diff
@@ -283,6 +281,7 @@ Error: %s`, fprocessErr.Error())
 				TLSInsecure:             tlsInsecure,
 			}
 
+			fmt.Println(checkTLSInsecure(deploySpec.Gateway, deploySpec.TLSInsecure))
```
alexellis: My final "nit" here is `fmt.Println` will print a new-line even if the message is `""`

Shouldn't this be:

```
if msg:= checkTLSInsecure(....); len(msg) >0 {
   fmt.Println(msg)
}
```

? cc @LucasRoesler

LucasRoesler: yes, that is right, which is why i originally proposed moving into the method.  We will need this if block everywhere to avoid the empty line

mario-mazo: or the option that i had initially works as well
```
fmt.Print(checkTLSInsecure(...), "\n")
```
no need for ifs just append the `\n` if needed

LucasRoesler: Perhaps I am misunderstanding something, but

```go
fmt.Print(checkTLSInsecure(...), "\n")
```

should always print a newline. So in the case when the url starts with `https`, there will be an empty line printed.  @alexellis's comment is that we don't want an empty line printed when the TLS warning message is empty.

mario-mazo: Oh yes  right. I was thinking about something else.
what I can do is print the \n in the function when the warning is there and return ""
this way we can use `fmt.Print()` here and in case of no warning no new line should be printed

alexellis: I've merged the PR to bring a conclusion to the conversation, I then fixed things up in #595 along with some missing help commands.

Let's review and raise a new issue if we want further changes. 

Mario thanks for your PR. Looking forward to the next one.

Issue #550, opened 2018-10-19
Suggestion: Support an OPENFAAS_PREFIX env var


rgee0:
Having to repeatedly supply `--prefix` was mentioned in #479 and #496 covers the broader idea of having a concept similar to the AWS CLI where profiles can be specified to save the user having to repeatedly supply details.  Should we support an env var which when present would negate the need to provide `--prefix`, in much the same way as OPENFAAS_URL negates the need to provide `--gateway/-g`

## Expected Behaviour
```
$ export  OPENFAAS_PREFIX=rgee0
$ faas-cli new --lang node newFn
$ cat newFn.yml
provider:
  name: faas
  gateway: http://127.0.0.1:8080
functions:
  fn1:
    lang: node
    handler: ./newFn
    image: rgee0/newFn:latest
```
## Current Behaviour
Provide `--prefix` on every call to new

## Possible Solution
Follow similar approach to OPENFAAS_URL with a precedence order established so that if `--prefix` is present it will over ride the env var.

## Context
Improved UX

## Your Environment

N/A

martindekov:
I think this is a very good idea. Most of the time I am pushing to my dockerhub and it is annoying to write the prefix every time. Is it approved? If so I can start working on it

ivanayov:
+1 for this

I wonder if it's better to make the name more descriptive, but  just`OPENFAAS_PREFIX` is also fine.
IMO  it should be taken into account only by `faas new` and prioritised after the `--prefix` flag.

rgee0:
> How would we expect this to work if the YAML file already has a prefix?

I was anticipating a precedence order approach similar to that which we use in the `getGateway()` method.  Whether the order itself is the same as that in the gateway decision is open to discussion.

> I wonder if it's better to make the name more descriptive, but justOPENFAAS_PREFIX is also fine.

It mirrors the flag, which the user should already be familiar with.  Happy to deviate on the suggested name, but in doing so we should consider whether the flag is affected in the same way.

martindekov:
Is there any movement on this Issue?

alexellis:
I think it would make more sense as `OPENFAAS_USER_PREFIX` - for context - `--prefix` was only chosen for the CLI for brevity.

I think there are two problems here so let's make sure we are solving the right one first:

## 1. Users want to hack on a stack with a prefix such as `openfaas` or `rgee0` but have to manually hack on the YAML and keep that in sync

## 2. Some people find adding `--prefix` to `faas new` tedious

[...]

Issue #417, opened 2018-06-07
Allow for --image to be defined for faas-cli push as it is for build and deploy.


IainColledge:
When using faas-cli in a CICD pipeline it is advantageous to override the developer defined image in the associated config.yml file.

For example a workflow may have a developer locally developing with a local image of hello-python but when the last git commit is pushed to the remote repository and subsequent CICD build, push and deploy activities take place they may use a remote docker repository.

Like faas-cli build and faas-cli deploy the use of --image IMAGE_NAME should be supported in order to decouple developer workflow and configuration from CICD pipeline steps and configuration.

## Expected Behaviour

Given an example config yml file hello-python.yml:

```
provider:
  name: faas
  gateway: http://127.0.0.1:8080

functions:
  hello-python:
    lang: python
    handler: ./hello-python
    image: hello-python
```

Executing:

`faas-cli build -f hello-python.yml --image REGISTRY_PREFIX/hello-python`

followed by:

`faas-cli push -f hello-python.yml --image REGISTRY_PREFIX/hello-python`

Should result in the image REGISTRY_PREFIX/hello-python being pushed to the registry defined by REGISTRY_PREFIX.

## Current Behaviour

faas-cli push fails with `Unknown flag: --image`

## Possible Solution

Change the code for faas-cli push to allow for --image flag usage.

## Context

Allowing developers to develop and test functions as they see fit without having to be coupled to the CICD build and deploy process.

## Your Environment

* FaaS-CLI version ( Full output from: `faas-cli version` ):

Commit: 6532f3c66967b3a872208a29b1b60a873b7d2490
Version: 0.6.9

* Docker version ( Full output from: `docker version` ):

Client:
 Version:      17.03.1-ce-rc1
 API version:  1.26 (downgraded from 1.27)
 Go version:   go1.7.5
 Git commit:   3476dbf
 Built:        Fri Mar 17 00:27:41 2017
 OS/Arch:      linux/amd64

Server:
 Version:      1.13.1
 API version:  1.26 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   092cba3
 Built:        Wed Feb  8 06:50:14 2017
 OS/Arch:      linux/amd64
 Experimental: false

* Are you using Docker Swarm (FaaS-swarm ) or Kubernetes (FaaS-netes)?

Kubernetes (Azure aks)

* Operating System and version (e.g. Linux, Windows, MacOS):

`uname -a 
[...]

PR #614, opened 2019-03-24
Add the option to override the image prefix

kamaln7:
This is an attempt to implement the functionality described in #550: add the option to set an image prefix though a CLI argument or environment variable.
## Description
This provides the option to override the image prefix in the stack yaml in the following preference order:

CLI Argument (`--prefix`), fall back to environment variable (`OPENFAAS_IMAGE_PREFIX`), fall back to the `stack.yml` image name.

If `stack.yml` contains an image name _without_ a prefix, but either the CLI arg or Env var (or both) are set, the prefix will be prepended to the image name.
If `stack.yml` contains an image name _with_ a prefix, and either the CLI arg or Env var (or both) are set, the prefix with the highest preference will override the one in `stack.yml`.

This change is added to `new`, `build`, and `push` like so:

* `faas-cli new` allows setting the prefix in an environment variable in
addition to the cli arg
* `faas-cli build` and `faas-cli push` now allow overriding the prefix set in the stack YAML,
provided by a cli arg or env var

Signed-off-by: Kamal Nasser <hello@kamal.io>

## Motivation and Context
With this change, people can easily re-use other people's `stack.yml` files without having to modify them to either add their own prefix or replace the `stack.yml`'s prefix.

## How Has This Been Tested?
Tests were added to the files that were modified.

## Types of changes

## Checklist:

I'd like to make sure that the functionality (e.g. overriding existing prefixes) is OK before adding documentation.

### DIFF
--- a/commands/push.go
+++ b/commands/push.go
@@ -5,6 +5,7 @@ package commands
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"sync"
 
@@ -20,7 +21,7 @@ func init() {
 
 	pushCmd.Flags().IntVar(&parallel, "parallel", 1, "Push images in parallel to depth specified.")
 	pushCmd.Flags().StringVar(&tag, "tag", "", "Override latest tag on function Docker image, takes 'sha' or 'branch'")
-
+	pushCmd.Flags().StringVarP(&imagePrefix, "prefix", "p", "", "Set prefix for the function image")
 }
 
 // pushCmd handles pushing function container images to a remote repo
@@ -54,6 +55,14 @@ func runPush(cmd *cobra.Command, args []string) error {
 		if parsedServices != nil {
 			services = *parsedServices
 		}
+	}
+
+	for _, function := range services.Functions {
+		if function.SkipBuild {
+			continue
+		}
+
+		function.Image = overrideImagePrefix(function.Image)
 	}
 
 	if len(services.Functions) > 0 {
@@ -140,4 +149,31 @@ func validateImages(functions map[string]stack.Function) []string {
 		}
 	}
 	return invalidImages
+}
+
+func overrideImagePrefix(yamlImageName string) string {
+	if yamlImageName == "" {
+		return ""
+	}
+
+	imageNameParts := strings.Split(yamlImageName, "/")
+	var (
+		yamlPrefix    string
+		yamlHasPrefix = (len(imageNameParts) > 1)
+	)
+	if yamlHasPrefix {
+		// prefix is set in yaml
+		yamlPrefix = imageNameParts[0]
+	}
+
+	imagePrefix = getImagePrefix(imagePrefix, os.Getenv(imagePrefixEnvironment), yamlPrefix)
+	if yamlHasPrefix {
+		// update prefix in place
+		imageNameParts[0] = imagePrefix
+	} else if imagePrefix != "" {
+		// prepend prefix to image name
+		imageNameParts = append([]string{imagePrefix}, imageNameParts...)
+	}
+
+	return strings.Join(imageNameParts, "/")
 }
--- a/commands/build.go
+++ b/commands/build.go
@@ -35,6 +35,7 @@ func init() {
 	buildCmd.Flags().StringVar(&handler, "handler", "", "Directory with handler for function, e.g. handler.js")
 	buildCmd.Flags().StringVar(&functionName, "name", "", "Name of the deployed function")
 	buildCmd.Flags().StringVar(&language, "lang", "", "Programming language template")
+	buildCmd.Flags().StringVar(&imagePrefix, "prefix", "", "Image prefix to optionally override the one set in the stack yaml")
 
 	// Setup flags that are used only by this command (variables defined above)
 	buildCmd.Flags().BoolVar(&nocache, "no-cache", false, "Do not use Docker's build cache")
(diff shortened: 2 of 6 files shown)


kamaln7:
I'm having second thoughts. What if the preference order was:

1. CLI arg
2. YAML file
3. Env var

Overriding based on an env var might cause unexpected behavior if someone forgot that they had one set. However overriding based on an explicitly passed prefix would make sense to me.

commands/push.go
```diff
@@ -141,3 +150,30 @@ func validateImages(functions map[string]stack.Function) []string {
 	}
 	return invalidImages
 }
+
+func overrideImagePrefix(yamlImageName string) string {
+	if yamlImageName == "" {
+		return ""
+	}
+
+	imageNameParts := strings.Split(yamlImageName, "/")
```
alexellis: This should be done without using split. There are other string methods available

alexellis:
I spoke with @rgee0 and we'd like to see a test with a private repo

alexellis:
You'll also need to test this with `faas-cli up`

commands/push.go
LucasRoesler: It might be worth looking at the docker distribution repo, they have a package dedicated to  parsing docker image names.  Many of the public objects aren't going to be useful, but the regex could be helpful or the logic in some of the parsers. 

Importantly, I think this current approach will have issues if the image name contains a domain, see the grammar section for https://github.com/docker/distribution/blob/6d62eb1d4a3515399431b713fde3ce5a9b40e8d5/reference/reference.go#L7

Adding a test case fo something like `quay.io/repo/func` would be good

alexellis:
On the comment from @LucasRoesler we do have that package vendored already for the `--send-registry-auth` functionality. @burtonr and @johnmccabe worked on it.

commands/push.go
```diff
@@ -141,3 +150,30 @@ func validateImages(functions map[string]stack.Function) []string {
...
+	)
+	if yamlHasPrefix {
+		// prefix is set in yaml
+		yamlPrefix = imageNameParts[0]
+	}
+
+	imagePrefix = getImagePrefix(imagePrefix, os.Getenv(imagePrefixEnvironment), yamlPrefix)
+	if yamlHasPrefix {
+		// update prefix in place
+		imageNameParts[0] = imagePrefix
```
ivanayov: In this case we end up with the same value (as the third parameter in `getImagePrefix` is with the biggest priority).
Can't we skip the replacement with the same value in that case?

alexellis:
@kamaln7 I've not seen any activity on this for several weeks. What's its current status?

Given the new env-var subst feature do you feel this is still required? cc @rgee0 

https://github.com/openfaas/faas-cli#environment-variable-substitution

kamaln7:
Hey @alexellis - thanks for the feedback & review everyone. Sorry, I haven't had time to address the comments.

The new envvar substitution feature looks great. It might be a good idea to just use that as it's more generic but that would restrict this to only stack yaml files that explicitly use it.

What do you think?

commands/push.go
kamaln7: Yes, that would be redundant in that case but I thought there wasn't need for special handling of it since it would ultimately be the same, just with more code

commands/push.go
kamaln7: Good call, thanks. I'll change it to use the package you linked for (un)marshalling

alexellis:
What I was saying was that I'm not sure we need this PR given the new envsubst feature for the YAML. 

Using the CLI to deploy without YAML is an edge-case.

kamaln7:
The envsubst feature needs the YAML to be edited, eg:

```
functions:
  url-ping:
    lang: python
    handler: ./sample/url-ping
    image: ${DOCKER_USER:-exampleco}/faas-url-ping:0.2
```

You can't use it with almost all of the yaml files out there on github/etc since they all look like so- that's what I meant

```
functions:
  url-ping:
    lang: python
    handler: ./sample/url-ping
    image: exampleco/faas-url-ping:0.2
```

alexellis:
Generally as a rule the contribution guide states that PRs are not worked on until there is approval that the feature is wanted and that the maintainers are happy with the proposed design. 

Env-var substitution is brand new so it's unlikely that any YAML files have it yet. Which YAML files are you looking at currently as an example?

alexellis:
I'm going to ask us to go back to the define / design phase on the issue.

Feel free to keep your branch around so that it could potentially be re-used once we have agreement on the design / feature from the core team.

CleverCoder:
I was going to open a new issue, but ran into a similar situation. I also saw the closed issue here, quite related: https://github.com/openfaas/faas-cli/issues/417

_The pre-condition: the function configuration file contains the gateway and repository name that's used in production. (this is a requirement for now)_

I have a setup where we want to have a configuration file with multiple properties (not just gateway, image, lang, handler), and I want to be able to run a set of CLI options to perform build/push/deploy where I override the image name that's used.
I can use Docker commands to create a tag locally similar to `localhost:5001/some-image:tag`, but I'd really like to issue the command:
```
faas up -f <myfunction>.yml -g $OPENFAAS_GATEWAY -i localhost:5001/some-image:tag --build-arg <...>
```
This would also be useful in CI pipelines where manipulation of the repo and image name is needed (not just the tag).
My only workaround is to modify the function.yml file and remember to switch things back to our ECR repo... not ideal.
Unless there's a better way to do this that I'm missing?

alexellis:
Hi thanks for your interest in OpenFaaS

When you say "we x y z", we (openfaas team) don't know you. Since we're giving you free technical support, could you please introduce the "we" briefly?

This is already a solved problem and you should use our recommended approach:

https://www.openfaas.com/blog/custom-environments/

This is also explained in the docs

Alex

CleverCoder:
That was fast!! I appreciate the quick reply. Regarding the free tech support bit, yeah.. I get it. "We" is an employer, and I'm trying to develop and test functions that are deployed to OpenFaas. Not sure what more you need there. I'm not expecting much, and understand the workload.

I reviewed the link you provided, and the only relevant part is regarding environment variable substitution. The secrets management works fine. Even then, it's not super clear how I can override the Docker image name while running in the CLI.

I also re-reviewed the documentation to find what's explained there (per your comment), and all I could find was more details around variable substitution (https://docs.openfaas.com/reference/yaml/#yaml-environment-variable-substitution). This is handy! That would imply then that we (our devops folks) would need to make changes to the configuration files to inject a proper docker image URL for prod builds (work for them), and then I could leverage the variable substitution. (I'm guessing this is what you meant)

All I was really hoping for was to be able to override the image from the CLI. Also, that post seems to focus on overriding environment variables, and I don't see how that can alter the Docker image repository:tag that's used during the build. Am I missing something? Again.. thanks for the quick response!

alexellis:
The request is to "override the image prefix" why? For different environments.

We've provided a feature to do this along with documentation, therefore this issue is complete and closed.

We expect employees to disclose what company we're supporting and to send their info into [ADOPTERS.md](https://github.com/openfaas/faas/blob/master/ADOPTERS.md)

If you don't like that arrangement, the code is open source and you're welcome to read/modify it as you require.

Issue #622, opened 2019-04-03
Incorrect unit test: Test_buildFlagSlice


alexellis:
## Expected Behaviour

There is a test case which should fail in `Test_buildFlagSlice`, but the code is not checking want/got values.

PR #575 added a `--build-label` flag for the `build` command, yet we have a test merged for `--label` which is a flag which doesn't exist and is invalid - yet the test still passes.

```
            `expectedSlice: []string{"--build-arg", "muppets=burt and ernie", "--build-arg", "playschool=Jemima", "--label", "org.label-schema.name=test function"},`
```

## Possible Solution

1) Update the test logic running the test table so that the above scenario *fails*
2) Update the fixture / test case to use the proper flag `--build-label`

```go
for _, test := range buildFlagOpts {

		t.Run(test.title, func(t *testing.T) {

			flagSlice := buildFlagSlice(test.nocache, test.squash, test.httpProxy, test.httpsProxy, test.buildArgMap, test.buildPackages, test.buildLabelMap)
			fmt.Println(flagSlice)
			if len(flagSlice) != len(test.expectedSlice) {
				t.Errorf("Slices differ in size - wanted: %d, found %d", len(test.expectedSlice), len(flagSlice))
			}

			//map iteration is non-deterministic so dont compare values if the buildArgMap is > 1
			if len(test.buildArgMap) <= 1 {
				for i, val := range flagSlice {

					if val != test.expectedSlice[i] {
						t.Errorf("Slices differ in values - wanted: %s, found %s", test.expectedSlice[i], val)
					}

				}
			}

		})
	}
```

PR #627, opened 2019-04-08
Fix unit test for builder

viveksyngh:
This commit fixes issue with unit tests in builder package.

Fixes: #622

## Motivation and Context

#622 

## How Has This Been Tested?

Since it is only unit tests update .. No need to test the functionalities. 

## Types of changes

## Checklist:

### DIFF
--- a/builder/build_test.go
+++ b/builder/build_test.go
@@ -297,9 +297,8 @@ func Test_buildFlagSlice(t *testing.T) {
 
 }
 
-func compareSliceValues(expectedSlice, outputSlice []string) bool {
+func compareSliceValues(expectedSlice, actualSlice []string) bool {
 	var expectedValueMap = make(map[string]int)
-
 	for _, expectedValue := range expectedSlice {
 		if _, ok := expectedValueMap[expectedValue]; ok {
 			expectedValueMap[expectedValue]++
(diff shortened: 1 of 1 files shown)


builder/build_test.go
```diff
@@ -287,20 +287,49 @@ func Test_buildFlagSlice(t *testing.T) {
...
+		if _, ok := outputValueMap[outputValue]; ok {
+			outputValueMap[outputValue]++
+		} else {
+			outputValueMap[outputValue] = 1
+		}
+	}
+
+	for key, expectedValue := range expectedValueMap {
+
+		if value, ok := expectedValueMap[key]; ok {
```
LucasRoesler: Should this use the `outputValueMap` instead? 

```
		if value, ok := outputValueMap[key]; ok {
```

You iterate over the `expectedValueMap` and then select again form `expectedValueMap`, it seems tautological.


You could also reduce the amount of nesting and make the code a little more symantec by doing something like this 

```go
for key, expectedValue := range expectedValueMap {
	value, exists := outputValueMap[key]
	if !exists || value != expectedValue  {
		return false
	}
}
```
I think this if statement reads closer to how your would describe the code, "if the value does not exist or is not the expected value, return false"  The order becomes important in the if statement, because you only want to check the value after you know the value exists.


I think you need to include a length check to ensure that there are no extra values in the output map. This should probably be the first thing that happens.  

Alternatively, this can also be verified while you construct the `outputValueMap`. If you find a value that does not exist in the `expectedValueMap`, then you can return before you even check the check the values in this loop. 


Lastly, the role of the values might be clearer if we use `actual*` instead of `output*`.   This will align closer to the error messages we use in other tests.

viveksyngh: Thanks for the feedback. Yes that should be `outputValueMap '. 

I have updated with all changes from review comments.

zeerorg:
Yep seems like it was passing false positives with previous commit. Sorry should have tested it more thoroughly. The current commit fixes the issue.

viveksyngh:
@zeerorg Yes we can add a different unit test function to test negative case.

alexellis:
Derek set milestone: 0.8.10

PR #632, opened 2019-04-28
Add url-only flag to describe command

rgee0:
Signed-off-by: Richard Gee <richard@technologee.co.uk>

## Description
This change adds the --url-only flag to describe.  Defaulting to false, when this flag is set the runDescribe function will return as soon as the URL is known.

## Motivation and Context
Describe currently passes out a nice table of various details.  There may be occasions - the workshop being one - where the user only requires the URL element of this table.  Further, by returning a naked URL the describe command can be cleanly embedded into other commands, for example: curl -d 'hello' $(faas describe go-echo --url-only).

### Existing Command
```bash
$ ./faas-cli-darwin describe go-echo
Name:                go-echo
Status:              Ready
Replicas:            1
Available replicas:  1
Invocations:         4
Image:               rgee0/go-echo:latest
Function process:    ./handler
URL:                 http://127.0.0.1:8080/function/go-echo
Async URL:           http://127.0.0.1:8080/async-function/go-echo
Labels:              com.openfaas.function : go-echo
                     function : true
```

### New Option

```bash
$ ./faas-cli-darwin describe go-echo --url-only
http://127.0.0.1:8080/function/go-echo
```

### New Option Embedded
Note: Intended to demonstrate embedding, in reality a user would use the invoke command for the same effect.  

```bash
$ curl -d 'hello' $(./faas-cli-darwin describe go-echo --url-only)
Hello world, input was: hello
```

```bash
$ faas-cli-darwin describe --help
Display details of an OpenFaaS function

Usage:
  faas-cli describe FUNCTION_NAME [--gateway GATEWAY_URL] [flags]

Examples:
faas-cli describe figlet 
faas-cli describe env --gateway http://127.0.0.1:8080
faas-cli describe echo -g http://127.0.0.1.8080

Flags:
      --envsubst         Substitute environment variables in stack.yml file (default true)
  -g, --gateway string   Gateway URL starting with http(s):// (default "http://127.0.0.1:8080")
  -h, --help             help for describe
      --name string      Name of the function
      --tls-no-verify    Disable TLS validation
      --url-only         Return only the function URL

Global Flags:
      --filter string   Wildcard to match with function names in YAML file
      --regex string    Regex to match with function names in YAML file
  -f, --yaml string     Path to YAML file describing function(s)
```

## Types of changes

## Checklist:

### DIFF
--- a/commands/describe.go
+++ b/commands/describe.go
@@ -17,11 +17,16 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	urlOnly bool
+)
+
 func init() {
 	describeCmd.Flags().StringVar(&functionName, "name", "", "Name of the function")
 	describeCmd.Flags().StringVarP(&gateway, "gateway", "g", defaultGateway, "Gateway URL starting with http(s)://")
 	describeCmd.Flags().BoolVar(&tlsInsecure, "tls-no-verify", false, "Disable TLS validation")
 	describeCmd.Flags().BoolVar(&envsubst, "envsubst", true, "Substitute environment variables in stack.yml file")
+	describeCmd.Flags().BoolVar(&urlOnly, "url-only", false, "Return only the function URL")
 
 	faasCmd.AddCommand(describeCmd)
 }
@@ -87,6 +92,11 @@ func runDescribe(cmd *cobra.Command, args []string) error {
 	}
 
 	url, asyncURL := getFunctionURLs(gatewayAddress, functionName)
+
+	if urlOnly == true {
+		fmt.Printf("%s\n", url)
+		return nil
+	}
 
 	funcDesc := schema.FunctionDescription{
 		Name:              function.Name,


alexellis:
I like this feature :+1: could it work any briefer like as `--url`?

rgee0:
I opted to stay away from `--url` as we have used it elsewhere with a different meaning

alexellis:
OK, how about a short-version to go with it like we have -f and --yaml - what would the short flag be for this command?

LucasRoesler:
What if we did `-f, --format=(table|url|json|...)`? if we decide to add more output formats, it will be easier to extend this way versus adding yet another flag.  

The formats I listed are just for examples, not a suggestion of which ones we should include.

rgee0:
@alexellis: what do you suggest? `-u` has been used as a shortcut to `--url`. 

@LucasRoesler: I thought about something similar while I was adding this - many of the CLIs offer `-f (parseable|standard|json)`.  However, I think that is a different feature - to me, a format flag is about specifying the format in which the information should be provided, rather than specifying what should be contained within the information.

LucasRoesler:
@rgee0 that is fair, although, I will softly counter that with `kubectl`'s current options

```
[(-o|--output=)json|yaml|wide|custom-columns=...|custom-columns-file=...|go-template=...|go-template-file=...|jsonpath=...|jsonpath-file=...]
```

three of the options allow you to customize the content in addition to the shape of the output.

I don't have strong feelings on this though. 

I am in favor of `--url` as a short flag for this.  I don't think there will be much confusion about the re-use of the flag in `faas-cli describe go-echo --url`  since it reads pretty semantically as "describe the url",

rgee0:
@LucasRoesler the first suggestion was `format` and the counter is `output`.  These words have different meanings.

I'll close this while we think it through.

LucasRoesler:
very very fair, if we went that direction output is definitely better than format.

PR #655, opened 2019-07-04
Handle '409 CONFLICT' during secret creation.'

bmcstdio:
## Description

This PR adds support for handling `409 CONFLICT` status codes returned by the OpenFaaS gateway when trying to create a secret that already exists. It is a companion PR to https://github.com/openfaas/faas-netes/pull/468.

## Motivation and Context

I believe https://github.com/openfaas/faas-netes/pull/468 contains all the required details.

## How Has This Been Tested?

I've tested it by extending the existing unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/proxy/secret.go
+++ b/proxy/secret.go
@@ -223,6 +223,9 @@ func CreateSecretToken(gateway string, secret schema.Secret, tlsInsecure bool, t
 	case http.StatusUnauthorized:
 		output += fmt.Sprintln("unauthorized access, run \"faas-cli login\" to setup authentication for this server")
 
+	case http.StatusConflict:
+		output += fmt.Sprintf("secret with the name %q already exists\n", secret.Name)
+
 	default:
 		bytesOut, err := ioutil.ReadAll(res.Body)
 		if err == nil {
--- a/proxy/secret_test.go
+++ b/proxy/secret_test.go
@@ -169,3 +169,22 @@ func Test_CreateSecret_Unauthorized401(t *testing.T) {
 		t.Fatalf("Error not matched: %s", output)
 	}
 }
+
+func Test_CreateSecret_Conflict409(t *testing.T) {
+	s := test.MockHttpServerStatus(t, http.StatusConflict)
+
+	secret := schema.Secret{
+		Name:  "secret-name",
+		Value: "secret-value",
+	}
+	status, output := CreateSecret(s.URL, secret, true)
+
+	if status != http.StatusConflict {
+		t.Errorf("want: %d, got: %d", http.StatusConflict, status)
+	}
+
+	r := regexp.MustCompile(`(?m:secret with the name "` + secret.Name + `" already exists)`)
+	if !r.MatchString(output) {
+		t.Fatalf("Error not matched: %s", output)
+	}
+}


proxy/secret.go
```diff
@@ -223,6 +223,9 @@ func CreateSecretToken(gateway string, secret schema.Secret, tlsInsecure bool, t
 	case http.StatusUnauthorized:
 		output += fmt.Sprintln("unauthorized access, run \"faas-cli login\" to setup authentication for this server")
 
+	case http.StatusConflict:
+		output += fmt.Sprintln("a secret with the same name already exists, run \"faas-cli secret update\" instead")
```
rgee0: I wonder whether we should keep the message objective. Updating a pre-existing secret could have undesired effects in some scenarios.  The user may instead use a different name.

alexellis: What would you suggest should happen in this case?

rgee0: That the user should find their own path since they know better the context of their environment than the CLI does.

Informing the user that the secret already exists helps them understand the cause of the failure without leading them towards one particular action.

Either that or add “doing so may have undesired effects on existing functions that are already using this secret”

LucasRoesler: I agree with @rgee0 , there are two equally valid options for the use (1) use a different name or (2) update the existing secret.  Why should the CLI direct you to the second option? I think it would be good enough to just say `a secret with the same name already exists`

alexellis: I'm good with what @rgee0 and @LucasRoesler said.

@bmcstudio thank you for handling the specific error, could you return the message such as mentioned above?

We'll let the user decide what to do when they have a secret that already exists rather than directing them to the *update* command.

bmcstdio: @alexellis @LucasRoesler @rgee0 thank you, I've updated the error message.

proxy/secret.go
```diff
@@ -223,6 +223,9 @@ func CreateSecretToken(gateway string, secret schema.Secret, tlsInsecure bool, t
 	case http.StatusUnauthorized:
 		output += fmt.Sprintln("unauthorized access, run \"faas-cli login\" to setup authentication for this server")
 
+	case http.StatusConflict:
+		output += fmt.Sprintln("a secret with the same name already exists")
```
alexellis: I would prefer: `secret with the name "X" already exists`

proxy/secret_test.go
```diff
@@ -169,3 +169,22 @@ func Test_CreateSecret_Unauthorized401(t *testing.T) {
...
+	s := test.MockHttpServerStatus(t, http.StatusConflict)
+
+	secret := schema.Secret{
+		Name:  "secret-name",
+		Value: "secret-value",
+	}
+	status, output := CreateSecret(s.URL, secret, true)
+
+	if status != http.StatusConflict {
+		t.Errorf("expected: %d, got: %d", http.StatusConflict, status)
```
alexellis: We use `want`/`got` rather than `expected`. Could you amend please?

bmcstdio:
@alexellis no worries, I've amended the commit according to your review. :+1:

alexellis:
Thank you Bruno

PR #682, opened 2019-09-01
WIP Proposal use git describe as build tag

### LINKED ISSUE #677: Proposal build and push with git describe value as the tag
The options to the `--tag` flag should be expanded to include `'git-describe'`

For CI builds, it would be really helpful to use the output of `git describe --tags` as the docker image tag.  Per the man page

```
The command finds the most recent tag that is reachable from a commit. If the tag points to the commit, then only the tag is shown. Otherwise, it suffixes the tag name with the number of additional commits  on top of the tagged object and the abbreviated object name of the most recent commit.
```

Tags would look like `v1.2.2-1-g3443110`.

## Expected Behaviour

Accepts `git-describe` or `describe` as an option for `--tags` and use the output of `git-describe --tags` for the value.

## Current Behaviour

Currently, it supports tagging the function with the git SHA or the git BRANCH
```sh
--tag string                 Override latest tag on function Docker image, takes 'sha' or 'branch'
```

## Possible Solution

Accepts `git-describe` or `describe` as an option for `--tags` and use the output of `git-describe --tags` for the value.

## Context

I am currently working on an example workflow with Github Actions and it occurred to me that I could not create a workflow that meaningfully used git tags (not without a lot of bash to parse outputs and retagging things).  It would significantly reduce the amount of string parsing required in CI if  the `faas-cli` could be made tag aware.

LucasRoesler:
Add a new option for the `--tag` flag, `describe`.  This will use the output of the `git describe --tags --always` as the docker image tag.  For example:

```sh
$  git describe --tags --always
v0.1.0
$ faas-cli build --tag=describe
[0] > Building classify.
Clearing temporary build folder: ./build/classify/
Preparing ./classify/ ./build/classify//function
Building: theaxer/classify:latest-v0.1.0 with pydatascience template. Please wait..
```

## Description

This PR is split into _TWO_ commits to separate a minimal implementation and then a subsequent refactor to cleanup some of the imports and hopefully improve code reusibility.

1) daeea84  Adds new GetGitDescribe ouput command, use this command in the build,
deploy, generate, and push commands.

2) 231e5e7  The refactor
    - Create and move exec commands to a new exec package, this will make it
easier to reuse in the future "install" command
    - Move git related commands to the versioncontrol package to keep the
git related commands together and easy to find.
    - Implement flags interfaces for the BuildFormat so that the flags
package can set and validate this value directly and then we can
directly pass this value around to improve type safety in the throughout
    - Renames the `tag` variable to `tagFormat` to improve readability, the
value is not directly the tag value and this could be confusing before.

I have marked this as a WIP because it is simply a demonstration to test out #677 Additionally, I plan to add some more tests and I think we could consider renaming `BuildFormat` to `TagFormat` or `ImageTagFormat` because the current name seems fairly ambiguous to me. 

## Motivation and Context

Resolves #677

## How Has This Been Tested?

I ran the impacted commands locally: `faas-cli build`, `faas-cli deploy`, `faas-cli generate`, and `faas-cli push`.   I verified the `--help` text was correct and I verified that the function image was built and tagged as expected.

## Types of changes

## Checklist:


schema/image.go
```diff
@@ -1,31 +1,84 @@
...
+		*l = DescribeFormat
+	default:
+		return fmt.Errorf("unknown image tag format: '%s'", value)
+	}
+	return nil
+}
+
 // BuildImageName builds a Docker image tag for build, push or deploy
-func BuildImageName(format BuildFormat, image string, SHA string, branch string) string {
+func BuildImageName(format BuildFormat, image string, version string, branch string) string {
```
LucasRoesler: @alexellis i am curious how you would feel about adding another option that allows us strip the `latest`.  In my ideal workflows I want images that are tagged `foo:v1.0.1` or `foo:a7afdl` not `foo:latest-v1.0.1`

alexellis:
FYI this would need to work with OpenFaaS Cloud's tag / formatting mechanism as per: https://github.com/openfaas/openfaas-cloud/blob/master/git-tar/function/ops.go#L180

LucasRoesler:
@alexellis this change only extends `BuildImageName`, it is fully backwards compatible and would not require any change in OpenFaaS Cloud

schema/image.go
alexellis: The text `latest` is the value from the stack.yml file:

```yaml
functions:
  git-tar:
    lang: go
    handler: ./git-tar
    image: git-tar:0.11.0
  buildshiprun:
    lang: go
    handler: ./buildshiprun
    image: buildshiprun:0.5.0

# repo: 0.9.0
```


It is effectively the static image tag as specified in the file, which may well be different to the tag of the Git repository. Take OpenFaaS cloud for instance. Each function has an independent version of the Git repo release.

PR #727, opened 2019-11-19
Allow for specifying the '--ssh' flag in 'faas build'.

bmcustodio:
## Description

This PR allows for passing the `--ssh` flag to `docker build`, to be used with BuildKit as described in the "SSH" section of [Build secrets and SSH forwarding in Docker 18.09](https://medium.com/@tonistiigi/build-secrets-and-ssh-forwarding-in-docker-18-09-ae8161d066).

## Motivation and Context

## How Has This Been Tested?

By running the included unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -20,6 +20,7 @@ import (
 
 // Flags that are to be added to commands.
 var (
+	ssh           string
 	nocache       bool
 	squash        bool
 	parallel      int
@@ -41,6 +42,7 @@ func init() {
 	buildCmd.Flags().StringVar(&language, "lang", "", "Programming language template")
 
 	// Setup flags that are used only by this command (variables defined above)
+	buildCmd.Flags().StringVar(&ssh, "ssh", "", `(Experimental) The value to be passed to the '--ssh' flag of 'docker build'.`)
 	buildCmd.Flags().BoolVar(&nocache, "no-cache", false, "Do not use Docker's build cache")
 	buildCmd.Flags().BoolVar(&squash, "squash", false, `Use Docker's squash flag for smaller images [experimental] `)
 	buildCmd.Flags().IntVar(&parallel, "parallel", 1, "Build in parallel to depth specified.")
@@ -170,7 +172,7 @@ func runBuild(cmd *cobra.Command, args []string) error {
 		if len(functionName) == 0 {
 			return fmt.Errorf("please provide the deployed --name of your function")
 		}
-		err := builder.BuildImage(image, handler, functionName, language, nocache, squash, shrinkwrap, buildArgMap, buildOptions, tagFormat, buildLabelMap)
+		err := builder.BuildImage(image, handler, functionName, language, ssh, nocache, squash, shrinkwrap, buildArgMap, buildOptions, tagFormat, buildLabelMap)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


commands/build.go
```diff
@@ -41,6 +42,7 @@ func init() {
 	buildCmd.Flags().StringVar(&language, "lang", "", "Programming language template")
 
 	// Setup flags that are used only by this command (variables defined above)
+	buildCmd.Flags().StringVar(&ssh, "ssh", "", `(Experimental) The value to be passed to the '--ssh' flag of 'docker build'.`)
```
LucasRoesler: If we add this flag, I think it would be good to extend this doc string to mention that BUILDKIT is required, maybe even directly mentioning the `DOCKER_BUILDKIT=1` var

alexellis: Oh does this *only* work with buildkit?

alexellis: +1 to Lucas' comment.

If so, it might be best to add the `DOCKER_BUILDKIT` env to the build command.

LucasRoesler: not only does it require buildkit, but you have use the experimental syntax in the Dockerfile `# syntax = docker/dockerfile:1.0-experimental`  and update the relevant RUN commands like this https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md#run---mounttypessh

bmcstdio: It doesn't _require_ Buildkit...

```
$ cat Dockerfile
FROM golang:1.13.4 AS builder
COPY . .
RUN go build -o /app main.go

FROM gcr.io/distroless/base:nonroot
COPY --from=builder /app /app
CMD /app

$ docker build --ssh default -t foo:bar .
(...)
Successfully tagged foo:bar
```

LucasRoesler: It won't fail without buildkit, but you must use the experimental mode/syntax to utilize the ssh feature, right?

bmcstdio: Yeah, right—hence the default being `false` 🙂

alexellis:
> Allow for specifying the '--ssh' flag in 'faas build'

I have looked at this a couple of times whilst at KubeCon, so I might missing something.

The `--ssh` flag is not about performing the entire build over `ssh` with a remote builder, but about a bespoke flag for `docker build` that enables secrets to be fetched over `ssh` at build-time?

bmcstdio:
> The --ssh flag is not about performing the entire build over ssh with a remote builder, but about a bespoke flag for docker build that enables secrets to be fetched over ssh at build-time? 

@alexellis in short—yes. I wonder if everyone has read the link I've provided in the description...

alexellis:
Getting secrets from SSH sounds very useful if we are adding the flag to the CLI, can you provide a minimal example that we can use to test it and document it for other users?

Also whilst `--ssh` is the flag for docker, it may be confusing to use the same for `faas-cli build`. Are there any other variants you think could fit?

commands/build.go
LucasRoesler: To be clear, I don't want to block accepting this PR, only improve the doc string to provide a stronger hint to people that experimental here is refers directly to an experimental, docker/buildkit feature.

This isn't _our_ experimental feature, it is Docker's experimental feature that we are exposing.

ehengao:
I have add another [pull request](https://github.com/openfaas/faas-cli/pull/786) similar with this change and tested locally after the built. 
Perhaps we can align on the naming and descriptions.

PR #732, opened 2019-12-06
[WIP] Consolidate exec package with go-execute

alexellis:
## Description

[WIP] Consolidate exec package with go-execute

## Motivation

This patch consolidates the exec package for docker build so that
it uses the the go-execute package used in other OpenFaaS projects.

The aim is to allow for conditional printing of stdio whilst also
being able to capture the output.

In a future PR a CLI animation can replace the Docker build, which
will be default, but optional. If an error is found then the
result of the build will be buffered and available to print to the
user.

This change stops Docker from printing progress bars when
downloading layers. Instead a line is printed when pulling and
when a layer is complete.

## How has this been tested

Tested for faas-cli build with multiple functions using the
sample stack.yml and --parallel=1/4.

## Type of change

Refactoring and fixes to error handling behaviour:

- Uses go-execute package
- Buffers stdio for use in error messages, before we didn't have access to this
- Makes it optional to stream to STDIO directly, before there was no choice
- Prints any errors that occur, after the build after all the workers have completed


alexellis:
Before merge, the 'dirty' changes in go-execute need to be upstreamed.

alexellis:
Hi @LucasRoesler I believe that I've resolved the vendor issue I mentioned in the comment above re: StreamStdio.

Alex

LucasRoesler:
I pulled the latest version and everything worked 💯

alexellis:
Thank you.

alexellis:
/set title: Consolidate exec package with go-execute for faas build

PR #741, opened 2019-12-20
Describe function command should include namespace in urls

### LINKED ISSUE #739: describe function command does not produce the correct url for the supplied namespace
The `faas-cli describe` command is namespace aware, but does not print the correct urls that include the `.<namespace>`

## Expected Behaviour

The output should include the url with the namespace

```sh
$ faas-cli describe --namespace beta travel-predictions
Name:                travel-predictions
Status:              Ready
...
Function process:
URL:                 https://example.com/function/travel-predictions.beta
```

## Current Behaviour

```sh
$ faas-cli describe --namespace beta travel-predictions
Name:                travel-predictions
Status:              Ready
...
Function process:
URL:                 https://example.com/function/travel-predictions
```

## Possible Solution

Update the `getFunctionURLs` method

1. Deploy to kuberenets with namespace support, as done in these docs https://docs.openfaas.com/reference/namespaces/
2. Add a namespace `beta` 
```sh
kubectl create ns beta
kubectl annotate ns beta openfaas="true"
```
3. Deploy a function to beta `faas-cli store deploy NodeInfo -n beta`
4. Describe that function `faas-cli describe -n beta nodeinfo`

## Context

Relates to openfaas/faas-netes#511

## Your Environment

`0.11.2`

* Are you using Docker Swarm (FaaS-swarm ) or Kubernetes (FaaS-netes)?
faas-netes in GKE

codegold79:
## Description

Name spaces have been appended to both the url and async url in the command package function getFunctionURLs.

## Motivation and Context

Fixes #739 

## How Has This Been Tested?

I have created a table test that ensures the expected URLs are generated for DNSs with and without ports, as well as for functions with and without namespaces.

I also built the faas-cli app on my MacOS using MiniKube and OpenFaaS installed with k3sup. I then enabled namespaces, a new feature as of this comment. I followed the instructions in the following pages to set up my system:
* https://docs.openfaas.com/deployment/kubernetes
* https://docs.openfaas.com/reference/namespaces/

I ran the describe command with `./faas-client describe <function> -n <namespace>` and `./faas-client describe <function>` and saw the expected URLs. See screenshot:
Screen Shot 2019-12-19 at 5 02 25 PM

## Types of changes

## Checklist:

### DIFF
--- a/commands/describe.go
+++ b/commands/describe.go
@@ -88,7 +88,7 @@ func runDescribe(cmd *cobra.Command, args []string) error {
 		status = "Ready"
 	}
 
-	url, asyncURL := getFunctionURLs(gatewayAddress, functionName)
+	url, asyncURL := getFunctionURLs(gatewayAddress, functionName, functionNamespace)
 
 	funcDesc := schema.FunctionDescription{
 		Name:              function.Name,
@@ -109,9 +109,18 @@ func runDescribe(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func getFunctionURLs(gateway string, functionName string) (string, string) {
+func getFunctionURLs(gateway string, functionName string, functionNamespace string) (string, string) {
 	gateway = strings.TrimRight(gateway, "/")
-	return gateway + "/function/" + functionName, gateway + "/async-function/" + functionName
+
+	url := gateway + "/function/" + functionName
+	asyncURL := gateway + "/async-function/" + functionName
+
+	if functionNamespace != "" {
+		url += "." + functionNamespace
+		asyncURL += "." + functionNamespace
+	}
+
+	return url, asyncURL
 }
 
 func printFunctionDescription(funcDesc schema.FunctionDescription) {
--- a/commands/describe_test.go
+++ b/commands/describe_test.go
@@ -1,0 +1,28 @@
+package commands
+
+import "testing"
+
+func Test_getFunctionURLs(t *testing.T) {
+	cases := []struct {
+		name              string
+		gateway           string
+		functionName      string
+		functionNamespace string
+		expectedURL       string
+		expectedAsyncURL  string
+	}{
+		{"localhost", "http://127.0.0.1:8080", "figlet", "alpha", "http://127.0.0.1:8080/function/figlet.alpha", "http://127.0.0.1:8080/async-function/figlet.alpha"},
+		{"secure site", "https://example.com", "nodeinfo", "beta", "https://example.com/function/nodeinfo.beta", "https://example.com/async-function/nodeinfo.beta"},
+		{"no namespace", "https://example.com:31112", "nodeinfo", "", "https://example.com:31112/function/nodeinfo", "https://example.com:31112/async-function/nodeinfo"},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			url, asyncURL := getFunctionURLs(tc.gateway, tc.functionName, tc.functionNamespace)
+
+			if url != tc.expectedURL || asyncURL != tc.expectedAsyncURL {
+				t.Fatalf("incorrect URL(s), want: %q and %q, got: %q and %q", tc.expectedURL, tc.expectedAsyncURL, url, asyncURL)
+			}
+		})
+	}
+}


codegold79:
Build failed. I'll look at the logs and see what failed tomorrow.

LucasRoesler:
@codegold79 The CI error is 

```
Run "gofmt -s -w" on your Golang code
The command '/bin/sh -c test -z "$(gofmt -l $(find . -type f -name '*.go' -not -path "./vendor/*"))" || { echo "Run \"gofmt -s -w\" on your Golang code"; exit 1; }' returned a non-zero code: 1
Makefile:13: recipe for target 'build' failed
```

Run `make local-fmt` and address whatever it says and the PR should be good

codegold79:
VS Code automatically runs `go fmt` when I save the file, but since all the formatting errors were invisible white spaces, I forgot to save. :doh: And then, I forgot to sign off. Had to force push. Bah. I blame morning brain fog.

Thanks @LucasRoesler for your help in this first PR for this repo. While waiting for code review, I'll go and see if I can wrap my head around what's happening with faas-containerd or openfaas/certifier.

commands/describe_test.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			url, asyncURL := getFunctionURLs(tc.gateway, tc.functionName, tc.functionNamespace)
+
+			if url != tc.expectedURL || asyncURL != tc.expectedAsyncURL {
+				t.Fatalf("incorrect URL(s):\n got %s and %s\n expected %s and %s\n", url, asyncURL, tc.expectedURL, tc.expectedAsyncURL)
+			}
+			t.Log("Expected URLs were generated: " + url + " and " + asyncURL + ".")
```
alexellis: What is the purpose of this log statement?

commands/describe_test.go
```diff
@@ -0,0 +1,29 @@
...
+		{"secure site", "https://example.com", "nodeinfo", "beta", "https://example.com/function/nodeinfo.beta", "https://example.com/async-function/nodeinfo.beta"},
+		{"no namespace", "https://example.com:31112", "nodeinfo", "", "https://example.com:31112/function/nodeinfo", "https://example.com:31112/async-function/nodeinfo"},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			url, asyncURL := getFunctionURLs(tc.gateway, tc.functionName, tc.functionNamespace)
+
+			if url != tc.expectedURL || asyncURL != tc.expectedAsyncURL {
+				t.Fatalf("incorrect URL(s):\n got %s and %s\n expected %s and %s\n", url, asyncURL, tc.expectedURL, tc.expectedAsyncURL)
```
alexellis: Please use verbiage of want/got instead

I think we can do without the newlines too, perhaps use `%q` if you need to see the exact quoted values.

commands/describe_test.go
codegold79: I was thinking having a log of what went well would distinguish it from a test that didn't have tests and passed. I encountered PASS when I mistakenly had no tests, ran the test, and the result was PASS, because nothing failed.

There's a difference between PASS because there was no failure, and actually PASS because the tests were successful.

However, if it's not useful, not idiomatic, nor per the style of this repository, I'll remove the `t.Log`.

commands/describe_test.go
codegold79: Cool, thanks for the tip, I'll fix this Monday.

commands/describe_test.go
codegold79: I looked at several other tests in faas-cli/commands, and I didn't see any that logged the details of what resulted in a test PASS. So, for consistency, I'll remove the line beginning with `t.Log`.

commands/describe_test.go
codegold79: Done.

commands/describe_test.go
alexellis: Thanks

commands/describe.go
```diff
@@ -109,9 +109,16 @@ func runDescribe(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func getFunctionURLs(gateway string, functionName string) (string, string) {
+func getFunctionURLs(gateway string, functionName string, functionNamespace string) (string, string) {
 	gateway = strings.TrimRight(gateway, "/")
-	return gateway + "/function/" + functionName, gateway + "/async-function/" + functionName
+	url := gateway + "/function/" + functionName + "." + functionNamespace
```
alexellis: What happens here when the namespace is `""`?

commands/describe.go
```diff
@@ -109,9 +109,16 @@ func runDescribe(cmd *cobra.Command, args []string) error {
...
+	asyncURL := gateway + "/async-function/" + functionName + "." + functionNamespace
+
+	// Remove dot for functions without namespaces.
+	url = strings.TrimRight(url, ".")
```
alexellis: Might be better to add conditionally, rather than mutating the string and then trimming, this seems like a bit of a hack.

Would something like this work? The existing tests should continue to work. 

```
if len(functionNamespace) > 0 {

}
```

commands/describe.go
codegold79: The URLs end up having a dot at the end if namespace has nil value.

commands/describe.go
codegold79: Yes, I think it would work. Having an if statement also makes the comment "// Remove dot..." unnecessary. I'll remove that comment.

codegold79: I ended up writing a simpler `if functionNamespace == "" {...}`, which doesn't have a `len()` function call.

commands/describe.go
```diff
@@ -109,9 +109,20 @@ func runDescribe(cmd *cobra.Command, args []string) error {
...
 }
 
-func getFunctionURLs(gateway string, functionName string) (string, string) {
+func getFunctionURLs(gateway string, functionName string, functionNamespace string) (string, string) {
+	var url, asyncURL string
+
 	gateway = strings.TrimRight(gateway, "/")
-	return gateway + "/function/" + functionName, gateway + "/async-function/" + functionName
+
+	if functionNamespace == "" {
```
alexellis: I think you could do away with this repetition in each block. What if you just decide the suffix here and have it as an empty string when there is no namespace? That also means you can do away with the else

codegold79: I like your suggestion--I think the code looks much cleaner now.

PR #744, opened 2019-12-24
[Feature] Copy common files into each function's handler folder

### LINKED ISSUE #322: Design review: "I want a copy of folder X in every function"
Some users have asked for a common folder "X" to be copied into every one of their functions at build time.

i.e.

```
functions/
 + /fn1/
 + /fn2/
 + /fn3/
 + /common_folder/
```

In the scenario above - the user wants "common_folder" to be copied into or to be accessible within fn1, fn2 and fn3 and also expects the CLI to do that.

> My take - this is something which is not possible with Docker due to files outside the build context being inaccessible. It doesn't feel like the responsibility of the faas-cli templating system to get involved in additional user files.

Possible workaround: have user make use of a Makefile or batch file to copy functions into a temporary folder and build out the structure the way they want.

@stefanprodan suggested using "grunt" or a JavaScript task runner just like when you build a react website from a "src" folder into a "dest" folder.

LucasRoesler:
## Description
- Extend the yaml schema specification to add `copy_extra_paths` to the stack `configuration` section
- Extend the behavior of `createBuildTemplate` to accept a slice of additional paths that should be copied into the build context
- Add a new flag `copy-extra` per the design request in  https://github.com/openfaas/faas-cli/issues/322#issuecomment-543127450 that allows the user to supply additional build folders to add to the `copy_extra_paths`
- Includes a new unit test to ensure that the extra paths copy can not copy folders from outside the current project. This is the same behavior and nearly the same implementation as is found in the `docker build`

## Motivation and Context

Resolves #322 

## How Has This Been Tested?

The attached [faas-cli-test-binaries.zip](https://github.com/openfaas/faas-cli/files/3998791/faas-cli-test-binaries.zip) can be used to test this behavior. Or it can be built locally using `make local-install`

- I have created a tiny sample project that demonstrates how it behaves in a function https://github.com/LucasRoesler/copy-common-example and I have verified that this works as expected
- Additionally, a new test was added to ensure the the copied paths are scoped correctly
- I have verified the error message is presented correctly. Here is a sample of the error message if the specified path is outside of the project scope

```sh
$ faas-cli build
[0] > Building echo-common.
Clearing temporary build folder: ./build/echo-common/
Preparing ./echo-common/ ./build/echo-common/
Building: theaxer/echo-common:latest with Dockerfile template. Please wait..
[0] < Building echo-common done in 0.00s.
[0] Worker done.

Total build time: 0.00
Errors received during build:
- Forbidden path appears to be outside of the build context: : /private (/private)
```

## Types of changes

## Checklist:

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -23,7 +23,7 @@ import (
 const AdditionalPackageBuildArg = "ADDITIONAL_PACKAGE"
 
 // BuildImage construct Docker image from function parameters
-func BuildImage(image string, handler string, functionName string, language string, nocache bool, squash bool, shrinkwrap bool, buildArgMap map[string]string, buildOptions []string, tagMode schema.BuildFormat, buildLabelMap map[string]string, quietBuild bool) error {
+func BuildImage(image string, handler string, functionName string, language string, nocache bool, squash bool, shrinkwrap bool, buildArgMap map[string]string, buildOptions []string, tagMode schema.BuildFormat, buildLabelMap map[string]string, quietBuild bool, copyExtraPaths []string) error {
 
 	if stack.IsValidTemplate(language) {
 		pathToTemplateYAML := fmt.Sprintf("./template/%s/template.yml", language)
@@ -47,7 +47,7 @@ func BuildImage(image string, handler string, functionName string, language stri
 			return fmt.Errorf("building %s, %s is an invalid path", imageName, handler)
 		}
 
-		tempPath, buildErr := createBuildContext(functionName, handler, language, isLanguageTemplate(language), langTemplate.HandlerFolder)
+		tempPath, buildErr := createBuildContext(functionName, handler, language, isLanguageTemplate(language), langTemplate.HandlerFolder, copyExtraPaths)
 		fmt.Printf("Building: %s with %s template. Please wait..\n", imageName, language)
 		if buildErr != nil {
 			return buildErr
@@ -164,7 +164,7 @@ type dockerBuild struct {
 const defaultHandlerFolder = "function"
 
 // createBuildContext creates temporary build folder to perform a Docker build with language template
-func createBuildContext(functionName string, handler string, language string, useFunction bool, handlerFolder string) (string, error) {
+func createBuildContext(functionName string, handler string, language string, useFunction bool, handlerFolder string, copyExtraPaths []string) (string, error) {
 	tempPath := fmt.Sprintf("./build/%s/", functionName)
 	fmt.Printf("Clearing temporary build folder: %s\n", tempPath)
 
@@ -225,9 +225,53 @@ func createBuildContext(functionName string, handler string, language string, us
 		}
 	}
 
+	for _, extraPath := range copyExtraPaths {
+		extraPathAbs, err := pathInScope(extraPath, ".")
+		if err != nil {
+			return tempPath, err
+		}
+		// Note that if useFunction is false, ie is a `dockerfile` template, then
+		// functionPath == tempPath, the docker build context, not the `function` handler folder
+		// inside the docker build context
+		copyErr := CopyFiles(
+			extraPathAbs,
+			filepath.Clean(path.Join(functionPath, extraPath)),
+		)
+
+		if copyErr != nil {
+			return tempPath, copyErr
+		}
+	}
+
 	return tempPath, nil
 }
 
+// pathInScope returns the absolute path to `path` and ensures that it is located within the
+// provided scope. An error will be returned, if the path is outside of the provided scope.
+func pathInScope(path string, scope string) (string, error) {
+	scope, err := filepath.Abs(filepath.FromSlash(scope))
+	if err != nil {
+		return "", err
+	}
+
+	abs, err := filepath.Abs(filepath.FromSlash(path))
+	if err != nil {
+		return "", err
+	}
+
+	if abs == scope {
+		return "", fmt.Errorf("forbidden path appears to equal the entire project: %s (%s)", path, abs)
+	}
+
+	if strings.HasPrefix(abs, scope) {
+		return abs, nil
+	}
+
+	// default return is an error
+	return "", fmt.Errorf("forbidden path appears to be outside of the build context: %s (%s)", path, abs)
+}
+
+// appears to be unused???
 func dockerBuildFolder(functionName string, handler string, language string) string {
 	tempPath := fmt.Sprintf("./build/%s/", functionName)
 	fmt.Printf("Clearing temporary build folder: %s\n", tempPath)
--- a/builder/build_test.go
+++ b/builder/build_test.go
@@ -2,6 +2,7 @@ package builder
 
 import (
 	"fmt"
+	"path/filepath"
 	"strings"
 	"testing"
 
@@ -467,6 +468,69 @@ func Test_deDuplicate(t *testing.T) {
 
 				if found == false {
 					t.Errorf("Slices differ in values  - wanted: %s, found %s", strings.Join(test.expectedStrings, " "), strings.Join(uniqueStrings, " "))
+				}
+			}
+		})
+	}
+}
+
+func Test_pathInScope(t *testing.T) {
+	root, err := filepath.Abs(".")
+	if err != nil {
+		t.Fatalf("unexpected error during test setup: %s", err)
+	}
+
+	cases := []struct {
+		name         string
+		path         string
+		expectedPath string
+		err          bool
+	}{
+		{
+			name:         "can copy folders without any relative path prefix",
+			path:         "common/models/prebuilt",
+			expectedPath: filepath.Join(root, "common/models/prebuilt"),
+		},
+		{
+			name:         "can copy folders with relative path prefix",
+			path:         "./common/data/cleaned",
+			expectedPath: filepath.Join(root, "./common/data/cleaned"),
+		},
+		{
+			name: "error if path equals the current directory",
+			path: "./",
+			err:  true,
+		},
+		{
+			name: "error if relative path moves out of the current directory",
+			path: "../private",
+			err:  true,
+		},
+		{
+			name: "error if absolute path is outside of the current directory",
+			path: "/private/common",
+			err:  true,
+		},
+		{
+			name: "error if relative path moves out of the current directory, even when hidden in the middle of the path",
+			path: "./common/../../private",
+			err:  true,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			abs, err := pathInScope(tc.path, root)
+			switch {
+			case tc.err && err == nil:
+				t.Fatalf("expected error but got none")
+			case tc.err && !strings.HasPrefix(err.Error(), "forbidden path"):
+				t.Fatalf("expected forbidden path error, got \"%s\"", err)
+			case !tc.err && err != nil:
+				t.Fatalf("unexpected err \"%s\"", err)
+			default:
+				if abs != tc.expectedPath {
+					t.Fatalf("expected path %s, got %s", tc.expectedPath, abs)
 				}
 			}
 		})
--- a/commands/build.go
+++ b/commands/build.go
@@ -28,6 +28,7 @@ var (
 	buildArgs        []string
 	buildArgMap      map[string]string
 	buildOptions     []string
+	copyExtra        []string
 	tagFormat        schema.BuildFormat
 	buildLabels      []string
 	buildLabelMap    map[string]string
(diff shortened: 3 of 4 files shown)


builder/build.go
```diff
@@ -209,9 +209,46 @@ func createBuildTemplate(functionName string, handler string, language string, u
 		}
 	}
 
+	for _, extraPath := range copyExtraPaths {
+		extraPathAbs, err := pathInScope(extraPath, ".")
```
doowb: After this check, should the same logic be used as the above "Infos" loop in each "extra path" folder to ensure a `build` and `template` folder does not get copied in? I'm not sure if the reasoning for that check above really only comes into play when the function handler is the root of the project or if it matters all the time.

builder/build.go
```diff
@@ -209,9 +209,46 @@ func createBuildTemplate(functionName string, handler string, language string, u
...
+		if err != nil {
+			return tempPath, err
+		}
+		copyErr := CopyFiles(
+			extraPathAbs,
+			filepath.Clean(tempPath+"/"+extraPath),
```
doowb: I believe the "extra paths" should be copied into the `functionPath` to ensure they are at the same directory level as the rest of the function handler. This should allow using the extra files without any other changes to existing templates.

```suggestion
			filepath.Clean(functionPath+"/"+extraPath),
```

builder/build.go
```diff
@@ -209,9 +209,46 @@ func createBuildTemplate(functionName string, handler string, language string, u
...
+	if err != nil {
+		return "", err
+	}
+
+	abs, err := filepath.Abs(filepath.FromSlash(path))
+	if err != nil {
+		return "", err
+	}
+
+	if strings.HasPrefix(abs, scope) {
```
doowb: I'm not familiar with `go`, so will this ensure that a user is not trying to copy the same folder as the `scope`? Meaning, should there be a check for `abs !== scope` also?

builder/build_test.go
```diff
@@ -472,3 +473,61 @@ func Test_deDuplicate(t *testing.T) {
...
+		{
+			name: "error if absolute path is outside of the current directory",
+			path: "/private/common",
+			err:  true,
+		},
+		{
+			name: "error if relative path moves out of the current directory, even when hidden in the middle of the path",
+			path: "./common/../../private",
+			err:  true,
+		},
```
doowb: Based on my comment above, I think adding the following might catch that case (_I'm trying to use `suggestion` feature but I'm not sure if it works for multiple lines_)
```suggestion
		},
		{
			name: "error if path is the current directory",
			path: root,
			err:  true,
		},
```

builder/build.go
LucasRoesler: i am not sure that check is needed because the use is explicitly adding these extra paths. In the initial infos loop, we are doing implicit copies, so i think it makes sense to skip.

However, we should definitely check that they don't add `.`, I dont think that makes sense, which is exactly your comment below https://github.com/openfaas/faas-cli/pull/744/files/e93706a1ce32103171da538c01b4bf3857fda6d9#r363798976

I will update that and add a test for it

builder/build.go
LucasRoesler: The only potential issue with this is that there are much more likely to be file path collisions, you could have `data` in the top of the project and then `data` inside your function. The default behavior would result in a merge/overwrite of the existing `function/data`? 

If we all agree that this is what we want, then i am more than happy to change it.

builder/build.go
LucasRoesler: You are right, i will add a check that `abs != scope`

burtonr:
I tried this, but I couldn't get it to work... Forgive me if I missed something, but here's the list of things I tried with more detail below:
- `./faas-cli build -f function.yml --copy-extra ./config`
- `./faas-cli build -f function.yml --copy-extra=./config`
- `./faas-cli build -f function.yml --copy-extra="./config"`
- `./faas-cli build --copy-extra ./config -f faas-echoer.yml`
- using the `copy_extra_paths` as a property of the `configuration` setting in the `yml` file

**Steps I've taken:**

Cloned this branch and built the cli with `go build`.

With an existing function I have, running the cli command: 

```
$ ./faas-cli build -f faas-echoer.yml --copy-extra=./config
```

I end up with the following in the build container/function:

```
$ docker run -it --rm burtonr/faas-echoer:0.0.2 /bin/sh
~ $ ls
fwatchdog   handler     handler.go
```

Next, I used the `yml` file as shown in the example repo

```yaml
version: 1.0
provider:
  name: openfaas

configuration:
  copy_extra_paths:
    - ./config

functions:
  faas-echoer:
    lang: golang-http
    handler: ./faas-echoer
    image: burtonr/faas-echoer:0.0.2
```

Again, the same result. No "config" folder in the container.

Next, I cloned the `copy-common-example` repo, and ran the test as it's documented. It works!
```
$ docker run -it --rm theaxer/echo-common:latest /bin/sh
~ $ ls
common            function          index.py          requirements.txt
```
```shell
2020/01/08 02:55:02 stdout: Serving on http://0.0.0.0:5000
2020/01/08 02:55:02 stdout: 2020-01-08 02:55:02,382:DEBUG:looking for: the coolest
2020/01/08 02:55:02 stdout: 2020-01-08 02:55:02,382:DEBUG:in{"the coolest": "openfaas"}
2020/01/08 02:55:02 POST / - 200 OK - ContentLength: 8
```

Now, to try again but with the "config" directory:
_(ran `docker rmi` and `docker image prune` to ensure nothing was cached)_

```yaml
provider:
  name: openfaas

configuration:
  copy_extra_paths:
    - ./common
    - ./config

functions:
  echo-common:
    lang: Dockerfile
    handler: ./echo-common
    fprocess: python index.py
    image: theaxer/echo-common
    environment:
      write_debug: true
```
No `config` directory in the container
``` shell
$ docker run -it --rm theaxer/echo-common:latest /bin/sh
~ $ ls
common            function          index.py          requirements.txt
```

I noticed that the example repo has the `common` directory being copied as part of the Dockerfile: https://github.com/LucasRoesler/copy-common-example/blob/master/echo-common/Dockerfile#L37

On a positive note, the `pathInScope` function does work and provides a proper error message as expected: 

``` shell
$ ./faas-cli build
[0] > Building echo-common.
Clearing temporary build folder: ./build/echo-common/
Preparing ./echo-common/ ./build/echo-common/
Building: theaxer/echo-common:latest with Dockerfile template. Please wait..
[0] < Building echo-common done in 0.00s.
[0] Worker done.

Total build time: 0.00
Errors received during build:
- Forbidden path appears to be outside of the build context: /home/burtonr/config (/home/burtonr/config)
```

burtonr:
Just wanted to clarify a little after talking with Lucas after my initial testing...

This feature is intended to copy a directory to the `build` directory. Adding the directory to the function container is an exercise left to the user.  The user should create a new custom template that includes a line in the Dockerfile that will `COPY` that directory to the designated location inside the container. 

That said, the example function/template that Lucas provided is what is expected. I can confirm, with my previous test run, is that the feature of the faas-cli *does* work as it should. The `config` directory was properly copied into the `build` directory. What I missed was the need for the custom template Dockerfile to be created and used to take that copied directory and add it to the function container image.

I didn't have enough time tonight to go through the entire scenario again, but I plan to. In the mean time, I can confirm that using Lucas' example repo did work as it is intended.

Once this does get approved and merged, I can write a new section in the docs to explain this behaviour so that others won't misunderstand as I had.

LucasRoesler:
@alexellis, can you also take a look. Ultimately, we need to make a decision about the destination of the additional paths. We can copy into the "/function" or leave it only in the docker build context (as it is now).

builder/build.go
doowb: That is more what I expected and I saw the PR subject saying that it is copying common files into the function's handler folder.

doowb:
I had the same initial expectation as @burtonr, that I could use existing templates and the common folders would be copied into the `function` folder.

My initial thoughts where that this is a way to share code between "functions" without an additional build step. In practice though, I would want that build step to be run while I'm doing development, which would mean this step would also be run before running `faas build`.

I can see how copying common data (I think @LucasRoesler use case was for machine learning models) can be useful also, and I still believe that copying to the `function` folder is the best approach. I'm having a hard time seeing the benefit of only copying to the Docker context and requiring custom templates to make use of the common folders (but I'm probably missing something).

The only reason I can see that being an issue for existing templates is when an existing function has conflicting paths in the `function` folder. But since this new option is `opt-in`, I don't see that being an issue.

fwiw... I only noticed that this was copying to the root of the Docker context and not the `function` path due to looking at the code. I probably wouldn't have noticed it any other way since most of the functions I write use the Dockerfile template which has all my function code at the root of the Docker context anyway.

burtonr:
@doowb After speaking with Lucas about this, I think the "trouble" with this is just that it's new and not documented. With a proper doc page in place to explain how it works (with a new template, or Dockerfile), it think a lot of the user issues can be prevented, or at least answered easily.
We'll wait for it to be merged in it's final form before writing it up though. Just in case something changes in the mean time, like the copy into directory...

LucasRoesler:
@alexellis, @burtonr, and @doowb  I have pushed a commit that copies into the `functionPath`. But, it is important to note that function path has two meanings.  If it is a `dockerfile` template, then `functionPath` == the docker build context. But for any other template then this is the `/function` folder.

This means that my `copy-common` example repo will _still_ copy into the the top level temp folder, not into the `/function`, because it is technically a `dockerfile` template.  I am a little worried about this behavior that changes between the two types of templates.  It is pretty subtle and would be easily missed, but I am not sure it is a common case that we should worry too much about.

I have attached more binaries so that you can try it out, if you like. 

One thing that is interesting to note, the `/function` folder is a convention that we have for the  templates, but I am not sure it is really checked/enforced anywhere. I wonder if we want to add some kind of warning message during the build if this folder is missing. 

[faas-cli-test-binaries-v2.zip](https://github.com/openfaas/faas-cli/files/4050477/faas-cli-test-binaries-v2.zip)

alexellis:
Playing catch-up here, but In terms of typing stuff in, this feels overly verbose to me vs. the rest of the schema:

```yaml
configuration:
  copy_extra_paths:
   - ./etc
```

I'd rather see it as shorter names that are quick to type:

```yaml
config:
  copy:
    - ./ 
```

alexellis:
I would expect the contents of the common folder to be copied inside the execution context of the function.

The example I saw at ADP for instance was that the user didn't want to use npm for a library for data access, but instead had a folder called `lib`, inside it existed "lib/dataaccess/index.js".

So for his various functions, he'd just add `const dataaccess = require("./dataaccess")` at the top of the file instead of publishing to npm and then importing that back for every minor change.

In the example where the files are put in the entry point, he'd have to understand the underlying filesystem a bit more and reach out to pull the files up `const dataaccess = require("../dataaccess")`

What are the pros/cons we perceive to using the function/ context for the common files?

alexellis:
Regarding the template folder name being `function`, I think this is something we can and should start to parameterise in template.yml

Case in point: I was trying to help @dokie build an elixir template, which expects the folder name to be something like `src` or `lib`, and so we did some trickery to work around the `function/` name.

In the case of the elixir template we could override this value `template_folder: lib` and for the Dockerfile template we could simply set it to `./` for instance.

@LucasRoesler what do you think?

alexellis:
@LucasRoesler I've merged a PR, and now this needs a rebase, please can you also add your patch from the zip file.

builder/build.go
alexellis: +1

builder/build.go
```diff
@@ -209,9 +209,53 @@ func createBuildTemplate(functionName string, handler string, language string, u
...
+	if abs == scope {
+		return "", fmt.Errorf("Forbidden path appears to equal the entire project: %s (%s)", path, abs)
+	}
+
+	if strings.HasPrefix(abs, scope) {
+		return abs, nil
+	}
+
+	// default return is an error
+	return "", fmt.Errorf("Forbidden path appears to be outside of the build context: %s (%s)", path, abs)
```
alexellis: errors need to start lowercase

builder/build_test.go
alexellis: PTAL

builder/build.go
LucasRoesler: My latest pushes fixes this

LucasRoesler:
> Playing catch-up here, but In terms of typing stuff in, this feels overly verbose to me vs. the rest of the schema:
> 
> ```yaml
> configuration:
>   copy_extra_paths:
>    - ./etc
> ```
> 
> I'd rather see it as shorter names that are quick to type:
> 
> ```yaml
> config:
>   copy:
>     - ./etc
> ```

This is also addressed in the latest version

LucasRoesler:
Here is a new set of binaries 
[faas-cli-test-binaries.zip](https://github.com/openfaas/faas-cli/files/4077769/faas-cli-test-binaries.zip)

LucasRoesler:
@alexellis I just squashed to a single commit to cleanup the history

builder/build_test.go
alexellis: @LucasRoesler PTAL

commands/build.go
```diff
@@ -75,6 +77,7 @@ var buildCmd = &cobra.Command{
 				 [--parallel PARALLEL_DEPTH]
 				 [--build-arg KEY=VALUE]
 				 [--build-option VALUE]
+				 [--copy-extra PATH]
```
alexellis: What is the priority of `--copy-extra` when a stack file is given?

LucasRoesler: It is merged with the value from the stack file

Issue #754, opened 2020-01-15
[Feature] Customise the template "function" folder


alexellis:
## Description

Customise the template "function" folder #754

## Context and motivation

Templates consist of two parts - an entry point, and a handler. The folder name for the handler is hardcoded to "function", but there's some cases where this is obstructive.

For Elixir the name must be "lib" or "src", I forget which cc @dokie

Here's an imaginary programming language "ext":

```
./index.ext
./function/handler.ext
```

What if "ext" must, like Elixir have all sub-folders with a name "src", then we'd be stuck and have to resort to renaming things in the Dockerfile.

```
./index.ext
./src/handler.ext
```

Since "function" is hard-coded in the Go code, we can keep it as a default and offer to switch out for any value specified in template.yml, for instance `handler_folder: src`.

This is what we did to evolve the fprocess from being hard-coded in Go to being flexible for each template.

dokie:
@alexellis For Elixir the root source folder is called `lib`

alexellis:
Thanks for confirming. That's now supported if you're interested in trying it out with your template store submission?

PR #755, opened 2020-01-15
Add ability to specify handler folder for templates

alexellis:
Signed-off-by: Alex Ellis (OpenFaaS Ltd) <alexellis2@gmail.com>

## Description

Add ability to specify handler folder for a template as defined in 
issue #754

## Motivation and Context

Fixes: #754

* The destination folder for the user's function / handler was
hard-coded to ./function/ or ./ for the Dockerfile template, this
patch allows users to override the value in template.yml
* Set handler_folder to a custom value such as "src"
* Updated new command for new handler path - the new command
needs to decide whether to find its templated handler code in the 
function folder or a custom folder.

## How Has This Been Tested?

Tested by updating the node template to use a handler_folder of
src, the build failed, then the Dockerfile was updated to match
the src folder instead of function, this resolved the error
and once deployed, it worked as expected.

Tested "faas-cli new" with the old cli and a template modified to use "src".
The result was an empty directory, I then updated faas-cli with
this patch and got the files populated as expected: i.e.
handler.js and package.json

To try this out:

1) Download the templates
2) Edit a template where you see "function" and replace it with something else like "lib" or "src"
3) Edit template.yml for that template and add the `handler_folder` entry
4) Build a new CLI, run `faas-cli build --shrinkwrap` and look inside build/name/
5) See that your function code was put into lib or src
6) Run a build, test the function

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -25,6 +26,16 @@ const AdditionalPackageBuildArg = "ADDITIONAL_PACKAGE"
 func BuildImage(image string, handler string, functionName string, language string, nocache bool, squash bool, shrinkwrap bool, buildArgMap map[string]string, buildOptions []string, tagMode schema.BuildFormat, buildLabelMap map[string]string, quietBuild bool) error {
 
 	if stack.IsValidTemplate(language) {
+		pathToTemplateYAML := fmt.Sprintf("./template/%s/template.yml", language)
+		if _, err := os.Stat(pathToTemplateYAML); os.IsNotExist(err) {
+			return err
+		}
+
+		langTemplate, err := stack.ParseYAMLForLanguageTemplate(pathToTemplateYAML)
+		if err != nil {
+			return fmt.Errorf("error reading language template: %s", err.Error())
+		}
+
 		branch, version, err := GetImageTagValues(tagMode)
 		if err != nil {
 			return err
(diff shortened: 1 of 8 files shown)


alexellis:
I know folks are busy with their own stuff, so I wanted to ping a few reviewers to see if someone could give this a visual check before merging. Also, if you have any comments or concerns please feel free to raise them *now*.

alexellis:
Thank you Richard

builder/build.go
```diff
@@ -152,34 +161,41 @@ type dockerBuild struct {
...
 	}
 
 	functionPath := tempPath
+
 	if useFunction {
-		functionPath += "/function"
+		if handlerFolder == "" {
+			functionPath = path.Join(functionPath, defaultHandlerFolder)
+		} else {
+			functionPath = path.Join(functionPath, handlerFolder)
```
martindekov: not sure do we want to check for prepended `/` for `handlerFolder`?

builder/build.go
```diff
@@ -152,34 +161,41 @@ type dockerBuild struct {
...
+		fmt.Printf("Error clearing temporary build folder: %s\n", tempPath)
 		return tempPath, clearErr
 	}
 
 	functionPath := tempPath
+
 	if useFunction {
-		functionPath += "/function"
+		if handlerFolder == "" {
+			functionPath = path.Join(functionPath, defaultHandlerFolder)
```
martindekov: Not sure how this even worked before given that:
```
functionPath := tempPath    (./build/somename/)
                                             ??
functionPath += "/function" (./build/somename//function)
```
(talking about the old revision)

martindekov:
Also strange that I see that build failed, even though I see the latest commit's build succeeded

builder/build.go
alexellis: Yes, I've found this kind of code in a few places and I'm fixing it up to use `path.Join`, it was probably written by younger me.

alexellis:
Fixed the build error.

builder/build.go
LucasRoesler: Does it make sense to replace all of the string formatting paths with path.Join?   There are a couple more like this one https://github.com/openfaas/faas-cli/pull/755/files#diff-85b0a27de88442108d89cc278530678eR168 and this one https://github.com/openfaas/faas-cli/pull/755/files#diff-85b0a27de88442108d89cc278530678eR345

alexellis: I left those as they were a bit more involved with the sprint, I think path.Join takes a slice?

builder/build.go
alexellis: It should add it through path.Join

PR #784, opened 2020-03-23
Use an id_token for implicit flow

alexellis:
Signed-off-by: Alex Ellis (OpenFaaS Ltd) <alexellis2@gmail.com>

## Description

Allows an id_token to be used and the id_token flow instead
of the access_token which may have been incorrect.

## How Has This Been Tested?

Fixes an issue reported when used with Azure which was not
reproducible with Auth0/GitLab.

### DIFF
--- a/commands/auth.go
+++ b/commands/auth.go
@@ -42,7 +42,7 @@ func init() {
 	authCmd.Flags().BoolVar(&launchBrowser, "launch-browser", true, "Launch browser for OAuth2 redirect")
 
 	authCmd.Flags().StringVar(&scope, "scope", "openid profile", "scope for OAuth2 flow - i.e. \"openid profile\"")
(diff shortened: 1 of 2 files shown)


commands/auth.go
```diff
@@ -91,14 +93,16 @@ func checkValues(authURL, clientID string) error {
 
 func runAuth(cmd *cobra.Command, args []string) error {
 	if grant == "implicit" {
-		return authImplicit()
+		return authImplicit("token")
```
LucasRoesler: Maybe a switch here will be easier to read?

alexellis: I prefer if statements for smaller blogs, but I get the point. I think there's a sensor that fires off for developers if they see > 2 if statements that says: this must be a switch statement.

alexellis:
Thank you @LucasRoesler for taking a look.

PR #757, opened 2020-01-21
Feature go modules migration

LucasRoesler:
## Description

"Upgrade" to Go modules by replacing the dep lock/toml with go mod/sum.  Then, update vendor folder.

## Motivation and Context

- [ ] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

## How Has This Been Tested?

I have run `make build` and `make build_redist` to test the docker builds.  I have also use `make local-install` and then tested the mac and linux binaries

I have also run the unit tests via `make test-unit`

In all cases the binaries build correctly or tests pass.

I have also done a quick test of 
- the build and deploy for a function, 
- the create and delete for a secret, 
- and logs
all of these are working in my tests.

Most of the changes have to deal with updating the docker `term` package, which allows us to deal with the logrus name change and the fact that docker is not using semver/modules.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] I have signed-off my commits with `git commit -s`
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #808, opened 2020-05-10
Override the stack.yaml namespace field from faas-cli. Fixes #742

### LINKED ISSUE #742: namespace flag is ignored when stack.yml exists
## Expected Behaviour

When I run:

```sh
faas deploy --name my-fn-name --image doowb/my-fn-name --namespace my-namespace
```
I expect the namespace flag to be used and the output to look like:

```sh
Deploying: my-fn-name.

Deployed. 202 Accepted.
URL: https://gateway.example.com/function/my-fn-name.my-namespace
```

## Current Behaviour

The above works when there is not a `stack.yml` file in the directory.
The following is the result I get when there is a `stack.yml` file, but there is not a namespace specified for the function in the file:

```sh
Deploying: my-fn-name.

Deployed. 202 Accepted.
URL: https://gateway.example.com/function/my-fn-name
```

## Possible Solution

If it is not the desired behavior to override the namespace with the flag, then I think the documentation should be updated to specify that, otherwise, let the flag override the stack specification. Right now the documentation doesn't say either way, rather there are only examples on how to deploy with the namespace in the file, or using the flag when deploying from the store.

## Context

This affected me because I was automating part of my deployment and I wanted to deploy a function to multiple namespaces, but it only deployed to the one specified in the `stack.yml`.

## Your Environment

```
CLI:
 commit:  73004c23e5a4d3fdb7352f953247473477477a64
 version: 0.11.3

Gateway
 uri:     https://gateway.example.com
 version: 0.18.6
 sha:     04fead5eabc4e68fcd746b8c400ded5eeb41266b
 commit:  Hide the namespace selector when on swarm or <2 K8s ns


Provider
 name:          faas-netes
 orchestration: kubernetes
 version:       0.9.15
 sha:           41c33f9f7c29e8276bd01387f78d6f0cff847890
```

* Docker version ( Full output from: `docker version` ):

* Are you using Docker Swarm (FaaS-swarm ) or Kubernetes (FaaS-netes)?

* Operating System and version (e.g. Linux, Windows, MacOS):

* Link to your project or a code example to reproduce issue:

utsavanand2:

## Description
This commit fixes a namespace issue, where the value passed into
the namespace flag was being ignored if a stack.yaml had a namespace field.

## Motivation and Context

Issue #742 

## How Has This Been Tested?

The output of the newly added test:
```
Running tool: /usr/local/opt/go/libexec/bin/go test -timeout 30s github.com/openfaas/faas-cli/commands -run ^(Test_overrideFlag_deployToNamespace)$

ok  	github.com/openfaas/faas-cli/commands	(cached)
Success: Tests passed.
```

## Types of changes

## Checklist:

### DIFF
--- a/commands/priority_test.go
+++ b/commands/priority_test.go
@@ -49,3 +49,53 @@ func Test_getTemplateStoreURL(t *testing.T) {
 		})
 	}
 }
+
+func Test_getOverrideNamespace(t *testing.T) {
+	tests := []struct {
+		stack    string
+		flag     string
+		want     string
+		scenario string
+	}{
+		// Test cases
+		{
+			stack:    "",
+			flag:     "",
+			want:     "openfaas-fn",
+			scenario: "no namespace value set in flag and in namespace field of stack file",
+		},
+
+		{
+			stack:    "openfaas-fn",
+			flag:     "foo",
+			want:     "foo",
+			scenario: "both stack file and CLI flag provide namespace values",
+		},
+
+		{
+			stack:    "bar",
+			flag:     "",
+			want:     "bar",
+			scenario: "stack file provides namespace value whereas no namespace is provided by CLI",
+		},
+
+		{
+			stack:    "",
+			flag:     "foo",
+			want:     "foo",
+			scenario: "flag provides namespace value whereas no namespace is provided by stack file",
+		},
+	}
+
+	// Run the test for each test case defined in "tests"
+	for _, testCase := range tests {
+		testCase := testCase
+		functionNamespace := getNamespace(testCase.flag, testCase.stack)
+
+		t.Run(testCase.scenario, func(t *testing.T) {
+			if functionNamespace != testCase.want {
+				t.Fatalf("Namespace incorrect want: %q but got: %q\n", testCase.want, functionNamespace)
+			}
+		})
+	}
+}
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -14,7 +14,7 @@
 
 [[projects]]
   digest = "1:04457f9f6f3ffc5fea48e71d62f2ca256637dee0a04d710288e27e05c8b41976"
-  name = "github.com/Sirupsen/logrus"
+  name = "github.com/sirupsen/logrus"
   packages = ["."]
   pruneopts = "UT"
   revision = "839c75faf7f98a33d445d181f3018b5c3409a45e"
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -204,6 +204,10 @@ func runDeployCommand(args []string, image string, fprocess string, functionName
 				functionSecrets = mergeSlice(function.Secrets, functionSecrets)
 			}
 
+			// Check if there is a functionNamespace flag passed, if so, override the namespace value
+			// defined in the stack.yaml
+			function.Namespace = getNamespace(functionNamespace, function.Namespace)
+
 			if deployFlags.sendRegistryAuth {
 
 				dockerConfig := configFile{}
--- a/commands/priority.go
+++ b/commands/priority.go
@@ -58,3 +58,19 @@ func getTemplateStoreURL(argumentURL, environmentURL, defaultURL string) string 
 		return defaultURL
 	}
 }
+
+func getNamespace(flagNamespace, stackNamespace string) string {
+	// If the namespace flag is passed use it
+	if len(flagNamespace) > 0 {
+		return flagNamespace
+	}
+
+	// if both the namespace flag in stack.yaml and the namespace flag are ommitted
+	// return the defaultNamespace (openfaas-fn)
+	if len(stackNamespace) == 0 && len(flagNamespace) == 0 {
+		return defaultFunctionNamespace
+	}
+
+	// Else return the namespace in stack.yaml
+	return stackNamespace
+}


commands/deploy_test.go
```diff
@@ -42,6 +49,62 @@ func Test_deploy(t *testing.T) {
 	}
 }
 
+func Test_oveerrideFlag_deployToNamespace(t *testing.T) {
```
LucasRoesler, line 52: I was also thinking about how to test this, it might be easier to move `functionNamespace` to `DeployFlags.Namespace` (updating how the flag is bound) and then pulling out the code for creating the request into a 

```go
func PrepareDeployRequest(function stack.Function, flags DeployFlags) proxy.DeployFunctionSpec {}
```

which allows you to just test that the correct request is constructed from the flags, avoiding the need for the mock http server, capture stdout, and the tempfile.

The current test is still good but tests several things at once, flag binding, parsing, file reading, request construction, and the specific endpoint used

utsavanand2: Yes, @LucasRoesler I was wondering too about testing, because it is a bit challenging, also because you need to pull in the templates before you run the test, or it just fails. What you're suggesting seems really efficient, just check the faas-cli request!

commands/deploy_test.go
```diff
@@ -42,6 +52,125 @@ func Test_deploy(t *testing.T) {
...
+				}
+				w.WriteHeader(http.StatusAccepted)
+			}
+		}))
+
+		defer srv.Close()
+
+		test.CaptureStdout(func() {
+			faasCmd.SetArgs([]string{
+				"deploy",
```
LucasRoesler, line 165: I am surprised this works, is the test server on localhost:8080 by default? it seems like you need to pass a gateway value here

utsavanand2: @LucasRoesler Yes you're right! I should pass in a gateway flag but as I pasted the output above, it seems to work fine, I'll update the faas-cli command to add the gateway flag and see if this fixes the error in the CI

utsavanand2:
@LucasRoesler I've fixed a few things with the way we construct the faas-cli args and the stack yaml, to make sure --namespace field remains omitted from the command and the stack yaml when the test case is 
 ```
   {
			stackNamespace:    "",
			flagNamespace:     "",
			expectedNamespace: "openfaas-fn",
		}
```
But now I'm getting this error from this test case, 
```
--- FAIL: Test_overrideFlag_deployToNamespace (3.25s)
    deploy_test.go:166: Requested Namespace from faas-cli's request is: default
        Expected namespace: openfaas-fn
FAIL
FAIL	github.com/openfaas/faas-cli/commands	3.867s
FAIL
Error: Tests failed.
```
Do you think faas-cli requests a `default` in the namespace when there are no namespaces in the stack or the cli flag, if not is the `default` being converted into `openfaas-fn` internally? Because if you look at this [snippet](https://github.com/openfaas/faas-cli/issues/742#issuecomment-626286378) the fix is behaving fine when we see the effect it has on the kubernetes namespace

**Comment by LucasRoesler** [comment]
@utsavanand2 i have seen a similar issue before with cobra, because it uses a pointer to the command and then on this line https://github.com/openfaas/faas-cli/pull/808/files#diff-06bf51d838c827fdcb4149062cc12331R191

you set the args in each test and those will persist between each loop of the subtests, because of the pointer. 

If you invert the order of the tests, you will see it pass. 

But this is ... fragile.  This is part of why i recommended creating a new method `PrepareDeployRequest` and only testing that, in this comment https://github.com/openfaas/faas-cli/pull/808#discussion_r422628438 It avoids this statefulness between test cases. 

For this PR, try inverting the test cases and add a code comment about the order being important.

File: go.mod [hunk 3]
```diff
@@ -0,0 +1,25 @@
+module github.com/openfaas/faas-cli
```
LucasRoesler, line 1: @utsavanand2 i don't think you meant to commit these, perhaps sometime this year we can verify and merge this https://github.com/openfaas/faas-cli/pull/757

commands/deploy_test.go
```diff
@@ -42,6 +52,155 @@ func Test_deploy(t *testing.T) {
...
+			})
+		} else {
+			cliArgs := []string{
+				"deploy",
+				"--yaml=" + tempFile.Name(),
+				"--namespace=" + testCase.flagNamespace,
+				"--gateway=" + srv.URL,
+			}
+			test.CaptureStdout(func() {
+				faasCmd.SetArgs(cliArgs)
```
LucasRoesler, line 196: @utsavanand2 unfortunately, i don't this testing pattern is going to work 

```
			test.CaptureStdout(func() {
				faasCmd.SetArgs(cliArgs)
				faasCmd.Execute()
			})
```
the  value is persisting into other tests

go.mod
utsavanand2: Yes, these won't be included in the final commit.
I tried rearranging the tests but it was not the issue, as I did include 
`testCase := testCase` inside of the loop that iterates over the tests. I found out that if faas-cli doesn't get a namespace flag or one in the stack yaml, then the Namespace field in the request doesn't switch to `openfaas-fn`, I think that is taken care of on the server. So after handling that in the server, the test is running successfully irrespective of the order. But running all of the tests at once with `go test` is causing it to fail because of the dir in which I'm pulling and creating the stack.yaml, so working on it now.

commands/deploy_test.go
utsavanand2: I tested it out, the values are not persisting, nor it was the order, but it was something in the server that handled the request, I've fixed it 🙂

LucasRoesler: May doing `faasCmd.SetArgs([]string{})` at the end of the test will work, but i haven't verified yet

utsavanand2:
@LucasRoesler Can you pull in the latest changes and see if it works on VS Code, it's working on mine, but for some reason the test `Test_List` fails, looks like its getting a namespace="default" in the arg, I wonder if it has something to do with my generated stack.yaml
Test running on VS Code:
Screenshot 2020-05-11 at 8 51 53 PM

Running all the tests with:
`go test` inside the `commands` dir
Screenshot 2020-05-11 at 8 53 22 PM

LucasRoesler:
@utsavanand2 this fails for me 

```
Running tool: /usr/local/bin/go test -timeout 30s github.com/openfaas/faas-cli/commands -run ^(Test_overrideFlag_deployToNamespace)$

--- FAIL: Test_overrideFlag_deployToNamespace (0.23s)
    /Users/lucasroesler/Code/go/src/github.com/openfaas/faas-cli/commands/deploy_test.go:139: Cannot create a tmp stack yaml file:
        open /var/folders/st/lt0d5y4j0sn7qm0z2mrb63zm0000gn/T/openFaasTestTemplates266572575/nstest/test-stack.406024178.yaml: no such file or directory
FAIL
FAIL	github.com/openfaas/faas-cli/commands	0.362s
FAIL
Error: Tests failed.
```

utsavanand2:
@LucasRoesler This seems really interesting 😂
Screenshot 2020-05-11 at 10 13 12 PM

commands/deploy_test.go
alexellis: `srv` should have a URL you can use.

alexellis: This example may help https://github.com/alexellis/derek/blob/03cfe0757d50cc38c92f3823f4a53b5adff26581/auth/isCustomer_test.go#L41

commands/deploy_test.go
```diff
@@ -42,6 +49,152 @@ func Test_deploy(t *testing.T) {
 	}
 }
 
+func Test_overrideFlag_deployToNamespace(t *testing.T) {
```
alexellis, line 52: There's so much going on in this test, I have a feeling that it will be hard for newcomers (or old-timers) to maintain. Have you got any suggestions for increasing the granularity?

commands/deploy_test.go
```diff
@@ -42,6 +49,152 @@ func Test_deploy(t *testing.T) {
...
+	var request types.FunctionDeployment
+	const defaultNamespace = "openfaas-fn"
+	var cliArgs []string
+	envMap := make(map[string]string)
+	function := make(map[string]stack.Function)
+
+	tests := []struct {
```
alexellis, line 59: You should simply be able to unit-test the method that decides over priority when given a number of strings. That's how we did it for the gateway URL. A full end-to-end test of what's sent to the server seems unnecessary to me.

alexellis: Take a look at https://github.com/openfaas/faas-cli/blob/5b36c50280d961d3aa248bd17f901f4ef2774447/commands/priority_test.go

alexellis: Alternative example: https://github.com/openfaas/faas-cli/blob/5b36c50280d961d3aa248bd17f901f4ef2774447/commands/config_priority_test.go

utsavanand2:
@alexellis Thanks a lot Alex for the examples, I got easily lost with the amount of code and the packages, not knowing exactly what does what. Thanks a lot for the examples. I think it's much easier to implement than what I've done already. Thanks a lot

commands/deploy_test.go
LucasRoesler: that url was already added 
image

LucasRoesler:
@utsavanand2 i second @alexellis comment and it is part of what i meant in this comment https://github.com/openfaas/faas-cli/pull/808#discussion_r422628438

I thought the original test would be easy to just switch the order and fix, but in general, i would try to avoid testing the cmd by setting args and invoking it. Writing a smaller method that is simpler to unit test is going to be better in the long run.

utsavanand2:
Thanks a lot @LucasRoesler You guys are awesome! I think I'm a bit new to the code-base and there are a lot of moving parts involved that it was hard for me to comprehend a lot of it, I shamelessly asked you a lot of things, and you were prompt to quickly help me out. I realize that all of you have day jobs and family so, I thought to implement it in a way that made sense to me. I really appreciate all of the help!

commands/priority_test.go
```diff
@@ -49,3 +49,51 @@ func Test_getTemplateStoreURL(t *testing.T) {
...
+		// Test cases
+		{
+			stackNamespace:    "",
+			flagNamespace:     "",
+			expectedNamespace: defaultNamespace,
+		},
+
+		{
+			stackNamespace:    "openfaas-fn",
+			flagNamespace:     "default",
```
alexellis, line 70: The default is `openfaas-fn` so I find this example rather confusing.

Was there a reason why you didn't use the samples given by @LucasRoesler ? https://github.com/openfaas/faas-cli/issues/742#issuecomment-625746405

in place of `expectedNamespace` please use `want`

`Namespace` as a suffix is probably not needed in the test table at all, it could be simplified to:

```
		stack    string
 		flag     string
 		want string
```

```
			t.Errorf("Expected Namespace: %s\nGot: %s\n", testCase.expectedNamespace, functionNamespace)
```

To:

```
			t.Errorf("want %s, but got %s", case)
```

I'd also like to see a named test-cases as follows:

```
stack: ""
flag: ""
want: defaultNs
scenario: "no value in stack or flag"
```


```
stack: ""
flag: "staging"
want: "staging"
scenario: "not set in stack.yml, but set as flag"
```

Then in each iteration of the loop, you should run a sub-test as we do elsewhere, this will output the name of the test scenario in the logs, in the case of a failure.

```
        t.Run(testCase.scenario, func(t *testing.T) {

       }

```

commands/priority_test.go
```diff
@@ -49,3 +49,60 @@ func Test_getTemplateStoreURL(t *testing.T) {
 		})
 	}
 }
+
+func Test_getOverrideNamespace(t *testing.T) {
+	const defaultFunctionNamespace = defaultFunctionNamespace
```
alexellis, line 54: This looks like an error, can you fix it?

commands/priority_test.go
```diff
@@ -49,3 +49,60 @@ func Test_getTemplateStoreURL(t *testing.T) {
...
+			stack:    "",
+			flag:     "",
+			want:     defaultFunctionNamespace,
+			scenario: "no namespace value set in flag or in namespace field of stack file",
+		},
+
+		{
+			stack:    "openfaas-fn",
+			flag:     "default",
+			want:     "default",
```
alexellis, line 74: I did ask you to remove the text "default" since it isn't in @LucasRoesler testing grid and we have defaultFunctionNamespace as a default along with `openfaas-fn`.

It's all rather confusing.

If you don't want to use what was suggested, what about:

`dev`, `staging` and `openfaas-fn` (default)`

That would be far clearer for me as a maintainer.

commands/priority_test.go
```diff
@@ -49,3 +49,60 @@ func Test_getTemplateStoreURL(t *testing.T) {
...
+			flag:     "",
+			want:     "default",
+			scenario: "stack file provides namespace as 'default' whereas no namespace is provided by CLI",
+		},
+
+		{
+			stack:    "",
+			flag:     "default",
+			want:     "default",
+			scenario: "flag provides namespace as 'default' whereas no namespace is provided by stack file",
```
alexellis, line 89: The actual literal value shouldn't be in the scenario. Use an abstract name

commands/priority_test.go
```diff
@@ -49,3 +49,60 @@ func Test_getTemplateStoreURL(t *testing.T) {
...
+	}
+
+	// Run the test for each test case defined in "tests"
+	for _, testCase := range tests {
+		testCase := testCase
+		functionNamespace := getNamespace(testCase.flag, testCase.stack)
+
+		t.Run(testCase.scenario, func(t *testing.T) {
+			if functionNamespace != testCase.want {
+				t.Errorf("Expected Namespace: %s Got: %s\n", testCase.want, functionNamespace)
```
alexellis, line 100: It's fine if you want to fail immediately here, you can use `t.Fatalf`, no need to try to make it complicated.

The error also doesn't take into the account the feedback I gave you, it should be `namespace incorrect want: %s, but got %s`, even better would be `%q` to show any whitespace.

go.mod
alexellis: Can you remove the go.mod and go.sum please? We're still using dep.

utsavanand2:
@alexellis I've remove go mod and go sum files. Anything else required?

Issue #830, opened 2020-09-30
Add better support for CI/CD environments


utsavanand2:
faas-cli can be improved to better handle running in CI/CD environments

## Expected Behaviour

Running faas-cli in CI/CD environments should allow for more permissive permissions for the artifacts created by `faas-cli login` and `faas-cli build`, so these files can be shared across multiple users in a CI/CD environment.
The location of the `.openfaas/config.yml` dir should be the current dir rather than the $HOME dir in a CI environment.
These change in behaviour should be able to be toggled with a simple env var or a flag.

## Current Behaviour

faas-cli does not have the aforementioned behaviour making it really difficult to run the container image of faas-cli in a CI environment without making the workflow configuration too complex for the average user.

## Possible Solution

PR #831 

## Context

Running the containerised version of faas-cli in Github Actions is difficult without making the workflow action YAML too complicated.

## Your Environment

* FaaS-CLI version ( Full output from: `faas-cli version` ): 0.12.13

alexellis:
Linked PR was merged. @utsavanand2 in future please use closing keywords in your commit or PR text, i.e. `Closes: #` or `Fixes: #` - otherwise it leaves dangling issues like this one.

utsavanand2:
Sure @alexellis ! I left it without the keyword initally, so I don't spam a lot of `forced-pushed` messages here. Thought I would do that in the end when the changes are finalised after review. But forgot about it. Will keep this in mind.

PR #831, opened 2020-09-30
Add CI env check to change dir and file permissions of artifcats

utsavanand2:

## Description

This PR will enable users of faas-cli, especially the containerised version to have more permissive rules for artefacts created by the CLI to be shared across multiple users, which is typically the case in a CI/CD environment.

## Motivation and Context

Setting the env var CI to true or 1 will make the artefacts created by `faas-cli login` and `faas-cli build` to have 0777 permissions, with the files being created in the current directory, this will make sure the build and login artefacts can be shared across the host OS and user and the container in environments like GitHub Actions where only some host directories like the working dir are mounted to the container. But other environments like Google Cloud Build will benefit from the same.

- [x] I have raised an [issue](#830) to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

## How Has This Been Tested?

A running sample of the containerised binary is running and building OpenFaaS function here:
https://github.com/utsavanand2/faasd-hello/runs/1177059289?check_suite_focus=true
while maintaining the simplicity of the workflow action yaml
```yaml
name: CI/CD

on:
  push:
    branches: 
      - master
      - readable-workflow
jobs:
  func-build:
    runs-on: ubuntu-latest
    steps:
      - 
        name: Checkout
        uses: actions/checkout@v2
      -
        name: Pull template
        uses: docker://utsavanand2/faas-cli:latest-root
        with:
          args: template store pull golang-http
      - 
        name: Run shrinkwrap build
        uses: docker://utsavanand2/faas-cli:latest-root
        with:
          args: build -f faasd-hello.yml --shrinkwrap
      -
        name: Login to OpenFaaS Gateway
        uses: docker://utsavanand2/faas-cli:latest-root
        with:
          args: login -p ${{ secrets.OPENFAAS_GATEWAY_PASSWD }} -g ${{ secrets.OPENFAAS_GATEWAY }}
      -
        name: Set up QEMU
        uses: docker/setup-qemu-action@v1
      -
        name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v1
      -
        name: Login to DockerHub
        if: success()
        uses: docker/login-action@v1
        with:
          username: ${{ secrets.DOCKER_USERNAME }}
          password: ${{ secrets.DOCKER_PASSWORD }}
      -
        name: List dir contents
        run: |
          ls -lah
      - 
        name: Build and Push the OpenFaaS function
        uses: docker/build-push-action@v2
        with:
          context: ./build/faasd-hello/
          file: ./build/faasd-hello/Dockerfile
          push: true
          platforms: linux/amd64,linux/arm64,linux/arm/v7
          tags: utsavanand2/faasd-hello:latest
      - 
        name: Deploy the function
        uses: docker://utsavanand2/faas-cli:latest-root
        with:
          args: deploy -f faasd-hello.yml
```
## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,16 @@ Alternatively you can assign a secret to the node to allow it to pull from your 
 
 For Docker Swarm use the `--send-registry-auth` flag or its shorthand `-a` which will look up your registry credentials in your local credentials store and then transmit them over the wire to the deploy command on the API Gateway. Make sure HTTPS/TLS is enabled before attempting this.
 
+### Use faas-cli in CI environments
+
+If you're running faas-cli in a CI environment like [Github Actions](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables), [CircleCI](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables), or [Travis](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables), chances are you get the env var `CI` set to true.
+
+If the `CI` env var is set to `true` or `1`, faas-cli change the location of the OpenFaaS config from the default `~/.openfaas/config.yml` to `.openfaas/config.yml` with elevated permissions for the `config.yml` and the shrinkwrapped `build` dir (if there is one).
+
+This is really useful when running faas-cli as a container image. The recommended image type to use in a CI environment is the root variant, tagged with `-root` suffix.
+CI environments like Github Actions require you to use Docker images having a root user. Learn more about it [here](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/dockerfile-support-for-github-actions#user).
+
+
 ### Use a YAML stack file
 
 Read the [YAML reference guide in the OpenFaaS docs](https://docs.openfaas.com/reference/yaml/).
--- a/builder/build.go
+++ b/builder/build.go
@@ -161,7 +161,19 @@ type dockerBuild struct {
 	BuildLabelMap    map[string]string
 }
 
-const defaultHandlerFolder = "function"
+var defaultDirPermissions os.FileMode = 0700
+
+const defaultHandlerFolder string = "function"
+
+// isRunningInCI checks the ENV var CI and returns true if it's set to true or 1
+func isRunningInCI() bool {
+	if env, ok := os.LookupEnv("CI"); ok {
+		if env == "true" || env == "1" {
+			return true
+		}
+	}
+	return false
+}
 
 // createBuildContext creates temporary build folder to perform a Docker build with language template
 func createBuildContext(functionName string, handler string, language string, useFunction bool, handlerFolder string, copyExtraPaths []string) (string, error) {
@@ -186,7 +198,11 @@ func createBuildContext(functionName string, handler string, language string, us
 
 	fmt.Printf("Preparing: %s %s\n", handler+"/", functionPath)
 
-	mkdirErr := os.MkdirAll(functionPath, 0700)
+	if isRunningInCI() {
+		defaultDirPermissions = 0777
+	}
+
+	mkdirErr := os.MkdirAll(functionPath, defaultDirPermissions)
 	if mkdirErr != nil {
 		fmt.Printf("Error creating path: %s - %s.\n", functionPath, mkdirErr.Error())
 		return tempPath, mkdirErr
--- a/config/config_file.go
+++ b/config/config_file.go
@@ -19,8 +19,16 @@ import (
 )
 
 var (
-	DefaultDir  = "~/.openfaas"
-	DefaultFile = "config.yml"
+	DefaultDir         string      = "~/.openfaas"
+	DefaultFile        string      = "config.yml"
+	DefaultPermissions os.FileMode = 0700
+
+	// DefaultCIDir creates the 'openfaas' directory in the current directory
+	// if running in a CI environment.
+	DefaultCIDir string = ".openfaas"
+	// DefaultCIPermissions creates the config file with elevated permissions
+	// for it to be read by multiple users when running in a CI environment.
+	DefaultCIPermissions os.FileMode = 0744
 )
 
 //AuthType auth type
@@ -58,15 +66,31 @@ func New(filePath string) (*ConfigFile, error) {
 	return conf, nil
 }
 
+// isRunningInCI checks the ENV var CI and returns true if it's set to true or 1
+func isRunningInCI() bool {
+	if env, ok := os.LookupEnv("CI"); ok {
+		if env == "true" || env == "1" {
+			return true
+		}
+	}
+	return false
+}
+
 // EnsureFile creates the root dir and config file
 func EnsureFile() (string, error) {
-	dirPath, err := homedir.Expand(DefaultDir)
+	permission := DefaultPermissions
+	dir := DefaultDir
+	if isRunningInCI() {
+		permission = DefaultCIPermissions
+		dir = DefaultCIDir
+	}
+	dirPath, err := homedir.Expand(dir)
 	if err != nil {
 		return "", err
 	}
 
 	filePath := path.Clean(filepath.Join(dirPath, DefaultFile))
-	if err := os.MkdirAll(filepath.Dir(filePath), 0700); err != nil {
+	if err := os.MkdirAll(filepath.Dir(filePath), permission); err != nil {
 		return "", err
 	}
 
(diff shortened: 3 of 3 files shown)


builder/build.go
```diff
@@ -15,13 +15,22 @@ import (
...
 )
 
 // AdditionalPackageBuildArg holds the special build-arg keyname for use with build-opts.
 // Can also be passed as a build arg hence needs to be accessed from commands
 const AdditionalPackageBuildArg = "ADDITIONAL_PACKAGE"
 
+func init() {
+	// Check the ENV VAR "CI" to know if faas-cli is being invoked in a CI/CD env
+	// and change the default dir location & permissions for files to be accessible by all users
+	if utils.IsRunningInCI() {
```
LucasRoesler, line 29: I am not sure i am a fan of adding stuff to the `init`. As a general rule this should be avoided. I would rather see `permissions` passed as an argument to the method that needs it.  Not only is it easier to test but it is also easier to read/debug if there is an error.

LucasRoesler (changes requested):

builder/build.go
utsavanand2: @LucasRoesler makes sense for testability. I'll make the relevant changes.

utsavanand2: @LucasRoesler I've tried not to touch the existing vars as they're declared at the top. I don't know if my current commit makes a lot of sense for what you would like to see, but considering I don't want to change the location of these global vars to a function, the current code made sense to me.
I'm thinking about not to affect additional code that could be added in the future to these files, changing the way these vars are accessed or set. What do you think?

config/config_file.go
```diff
@@ -60,13 +73,14 @@ func New(filePath string) (*ConfigFile, error) {
 
 // EnsureFile creates the root dir and config file
 func EnsureFile() (string, error) {
+	setDefaults()
```
LucasRoesler, line 76: @utsavanand2 this is still mutating a global constant.  From a practical perspective in the CLI, this is probably _ok_ here, but not a general pattern we would want to use.   Something like this simply references the global value when needed, but doesn't modify it

```go
permissions := DefaultPermissions
if IsRunningInCI() {
    permissions = DefaultCIPermissions
}
// ...
if err := os.MkdirAll(filepath.Dir(filePath), permissions); err != nil {
// ...
}
```

additionally, having a value for `defaultCIPermissions` means you can add a doc string explaining why we need that specific value in CI

utsavanand2:
Understood. Adding an additional global var with a doc string makes a lot
more sense.

LucasRoesler:
@utsavanand2 two more things

1. can you document with and include these links to explain why we chose and look at the `CI` environment variable
    * https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
    * https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
    * https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
2. similarly we should document that we want these changes for the permissions because we _have_ to run as root per this documentation in the Github Actions https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/dockerfile-support-for-github-actions#user

LucasRoesler:
I have one other question about how `faas-cli login` works with this flow. The change should make it possible to read the config from the local working directory, but it isn't clear to me how or if that is happening now

utsavanand2:
> @utsavanand2 two more things
> 
> 1. can you document with and include these links to explain why we chose and look at the `CI` environment variable
>    
>    * https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
>    * https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
>    * https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
> 2. similarly we should document that we want these changes for the permissions because we _have_ to run as root per this documentation in the Github Actions https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/dockerfile-support-for-github-actions#user

Sure!

utsavanand2:
> I have one other question about how `faas-cli login` works with this flow. The change should make it possible to read the config from the local working directory, but it isn't clear to me how or if that is happening now

@LucasRoesler What is really happening right now, is when the env var `CI` is set to `true` or `1`, then in the `login` step of the action, we create the config file in the current dir as `./openfaas/config.yml` rather than `~/.openfaas/config.yml`. Since the current workspace dir is mounted, the artifact i.e. the config file is preserved.
When we're ready to deploy our function in the last step, then config file is again looked up for from `./openfaas/config.yml`. As the workspace dir is again mounted to the containers working dir, making the command successful.
Your question made me realise that we might actually not need to change the permissions of the `config.yml` file since we're accessing it again inside the faas-cli container itself after the workdir mount.
Though this is not true for the `build` dir since it needs to be accessed by the Github Actions user to read it when invoking `docker build` with the shrinkwrapped build.
Let me know if I'm incorrect.

@alexellis Do you think we should trim down the permission further for the build dir from 0777 to 0744? Is that going to break anything? I'll sure try it out to see if it breaks my CI build.

LucasRoesler:
> > I have one other question about how `faas-cli login` works with this flow. The change should make it possible to read the config from the local working directory, but it isn't clear to me how or if that is happening now
> 
> @LucasRoesler What is really happening right now, is when the env var `CI` is set to `true` or `1`, then in the `login` step of the action, we create the config file in the current dir as `./openfaas/config.yml` rather than `~/.openfaas/config.yml`. Since the current workspace dir is mounted, the artifact i.e. the config file is preserved.
> When we're ready to deploy our function in the last step, then config file is again looked up for from `./openfaas/config.yml`. As the workspace dir is again mounted to the containers working dir, making the command successful.
> Your question made me realise that we might actually not need to change the permissions of the `config.yml` file since we're accessing it again inside the faas-cli container itself after the workdir mount.
> Though this is not true for the `build` dir since it needs to be accessed by the Github Actions user to read it when invoking `docker build` with the shrinkwrapped build.
> Let me know if I'm incorrect.

@utsavanand2 that is the flow i was expecting. When i reread the code diff and thought about it again, i was surprised not to see any changes in a login related file, i guess this whole flow is determined by the change in `fileExists`

utsavanand2:
> @utsavanand2 that is the flow i was expecting. When i reread the code diff and thought about it again, i was surprised not to see any changes in a login related file, i guess this whole flow is determined by the change in `fileExists`

@LucasRoesler Yes I expected the same before approaching the code, but it's really well organised in the config.go file, as you said we're just concerned with `fileExists` and `EnsureFile`.

README.md
```diff
@@ -252,6 +252,16 @@ Alternatively you can assign a secret to the node to allow it to pull from your
 
 For Docker Swarm use the `--send-registry-auth` flag or its shorthand `-a` which will look up your registry credentials in your local credentials store and then transmit them over the wire to the deploy command on the API Gateway. Make sure HTTPS/TLS is enabled before attempting this.
 
+### Use faas-cli in CI environments
+
+If you're running faas-cli in a CI environment, chances are you get an env var as `CI=true` or `CI=1`, like [Github Actions](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables), [CircleCI](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables), or [Travis](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables).
```
alexellis, line 257: "if you're running in a CI environment chances are"

Can you be more direct?

When running faas-cli in a CI service such as X, Y, Z, the CI environment variable is set to `true` by the platform.

builder/build.go
```diff
@@ -15,6 +15,7 @@ import (
 	v1execute "github.com/alexellis/go-execute/pkg/v1"
 	"github.com/openfaas/faas-cli/schema"
 	"github.com/openfaas/faas-cli/stack"
+	"github.com/openfaas/faas-cli/utils"
```
alexellis, line 18: I can't see why a separate package is needed. I would suggest removing that package and putting the code where it is consumed.

Packages named utils are almost always a code smell.

alexellis (changes requested):
Changes requested.

README.md
```diff
@@ -252,6 +252,16 @@ Alternatively you can assign a secret to the node to allow it to pull from your
 
 For Docker Swarm use the `--send-registry-auth` flag or its shorthand `-a` which will look up your registry credentials in your local credentials store and then transmit them over the wire to the deploy command on the API Gateway. Make sure HTTPS/TLS is enabled before attempting this.
 
+### Use faas-cli in CI environments
```
alexellis, line 255: Several paragraphs shouldn't be needed to explain that an automatic env-var changes the file-mode. Can you have a go at writing this in < 3-5 brief sentences?

README.md
utsavanand2: Resolved

README.md
utsavanand2: I've shortened it now, while still adding references to the links as discussed here.
https://github.com/openfaas/faas-cli/pull/831#issuecomment-703123046

builder/build.go
utsavanand2: Resolved

alexellis (approved):
Approved.

PR #836, opened 2020-10-25
Use buildx for multiarch docker builds

LucasRoesler:
## Description

- Refactor the Makefile and build scripts to consolidate the multiarch
  builds into a single docker file by using `docker buildx`. This mimics
  the changes from inlets/inlets#204.
- The redist script has been removed and the makefile now contains the
  equivalent commands.
- The build command now calls buildx twice, to get the root and rootless
  image.
- Add a new constant `shortPlatform` so that we can map the `TARGETPLATFORM` 
  value that is exposed by `buildx` so the required value used by the template store.

## Motivation and Context

## How Has This Been Tested?

Wait and see how travis feels

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,28 @@
 sudo: required
 language: generic
 
+dist: bionic
+
 services:
 - docker
 
 before_script:
-- curl -sSLf https://get.docker.com | sed s/sleep\ 20/sleep\ 0/g | sudo -E sh
+  - ./ci/upgrade-docker.sh
+  - ./ci/install-buildx.sh
+  - docker buildx version
 
 script:
-  - make build
+  - make build || travis_terminate 1;
+  - make build_redist
   - make build_samples
   - ./test/version.sh ./faas-cli
-  - make build_redist
 
 after_success:
-    - if [ ! -z "$TRAVIS_TAG" ] ; then
-        if [ -z "$DOCKER_NS" ] ; then
-          export DOCKER_NS=openfaas;
-        fi;
+  - make docker-login
+  - make docker-login-ghcr
+  - make push
+  - make push-ghcr
 
-        docker tag $DOCKER_NS/faas-cli:latest-dev $DOCKER_NS/faas-cli:$TRAVIS_TAG;
-        docker tag $DOCKER_NS/faas-cli:latest-dev $DOCKER_NS/faas-cli:latest;
-        docker tag $DOCKER_NS/faas-cli:latest-dev-root $DOCKER_NS/faas-cli:${TRAVIS_TAG}-root;
-        docker tag $DOCKER_NS/faas-cli:latest-dev-root $DOCKER_NS/faas-cli:latest-root;
-        echo $DOCKER_PASSWORD | docker login -u=$DOCKER_USERNAME --password-stdin;
-        docker push $DOCKER_NS/faas-cli:$TRAVIS_TAG;
-        docker push $DOCKER_NS/faas-cli:latest;
-        docker push $DOCKER_NS/faas-cli:${TRAVIS_TAG}-root;
-        docker push $DOCKER_NS/faas-cli:latest-root;
-
-      fi;
 
 before_deploy:
   - make test-templating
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -1,43 +1,44 @@
...
 language: generic
 
+dist: bionic
+
 services:
 - docker
 
 before_script:
-- curl -sSLf https://get.docker.com | sed s/sleep\ 20/sleep\ 0/g | sudo -E sh
+  - sudo rm -rf /var/lib/apt/lists/*
```
alexellis, line 10: See how I extracted this into a shell script. I think it may be easier to maintain that way.

https://github.com/teamserverless/license-check/blob/master/.travis.yml#L14

.travis.yml
```diff
@@ -1,43 +1,44 @@
...
-        docker tag $DOCKER_NS/faas-cli:latest-dev-root quay.io/$DOCKER_NS/faas-cli:${TRAVIS_TAG}latest-root;
-        docker tag $DOCKER_NS/faas-cli:latest-dev-root quay.io/$DOCKER_NS/faas-cli:latest-root;
-        echo $QUAY_PASSWORD | docker login -u=$QUAY_USERNAME --password-stdin quay.io;
-        docker push quay.io/$DOCKER_NS/faas-cli:$TRAVIS_TAG;
-        docker push quay.io/$DOCKER_NS/faas-cli:latest;
-        docker push quay.io/$DOCKER_NS/faas-cli:${TRAVIS_TAG}-root;
-        docker push quay.io/$DOCKER_NS/faas-cli:latest-root;
+  - make docker-login
+  - make docker-login-ghcr
+  - if [ ! -z "$TRAVIS_TAG" ] ; then
```
alexellis, line 32: We shouldn't need this, or it can move to the Makefile

alexellis: See how I did it for [inlets/inlets](https://github.com/inlets/inlets/blob/master/Makefile)

Dockerfile
```diff
@@ -1,14 +1,23 @@
 FROM teamserverless/license-check:0.3.6 as license-check
```
alexellis, line 1: Can you bump to 0.3.9?

Dockerfile
```diff
@@ -21,31 +30,31 @@ RUN test -z "$(gofmt -l $(find . -type f -name '*.go' -not -path "./vendor/*"))"
...
 RUN /usr/bin/license-check -path ./ --verbose=false "Alex Ellis" "OpenFaaS Author(s)"
 
-RUN go test $(go list ./... | grep -v /vendor/ | grep -v /template/|grep -v /build/|grep -v /sample/) -cover
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
+    go test $(go list ./... | grep -v /vendor/ | grep -v /template/|grep -v /build/|grep -v /sample/) -cover
 
-RUN VERSION=$(git describe --all --exact-match `git rev-parse HEAD` | grep tags | sed 's/tags\///') \
-    && GIT_COMMIT=$(git rev-list -1 HEAD) \
-    && CGO_ENABLED=0 GOOS=linux go build --ldflags "-s -w \
+RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} CGO_ENABLED=0 \
```
alexellis, line 36: Do we need `-mod=vendor` here?

.travis.yml
LucasRoesler: i wasn't sure what you wanted to do with the quey stuff, so i left it alone

Dockerfile
LucasRoesler: it is set via `ENV GOFLAGS=-mod=vendor`

.travis.yml
alexellis: Sure. Can we do it like the inlets/inlets repo?

Basically build again with another tag, it just uses the cache.

Let's switch to GHCR for this repo from Quay.io, I just checked that nobody has downloaded it from there all year.

alexellis: (FYI I've added GHCR credentials and I've deleted the Quay.io repo for faas-cli)

alexellis (commented):
Sending comments over

Makefile
```diff
@@ -1,22 +1,55 @@
...
+		--build-arg http_proxy=${http_proxy} --build-arg https_proxy=${https_proxy} \
+		--build-arg VERSION=$(.GIT_VERSION) --build-arg GIT_COMMIT=$(.GIT_COMMIT) \
+		--platform $(.PLATFORMS) \
+		--output "$(.BUILDX_OUTPUT)" \
+		--target root \
+		--tag openfaas/faas-cli:latest-root \
+		--tag openfaas/faas-cli:$(.GIT_VERSION)-root .
 
 .PHONY: build_redist
 build_redist:
```
alexellis, line 38: Previously we had a build that might have run 2x as fast because we built in Docker, then pulled out the binaries.

How that we are both building in Docker, and on the host, how does the performance compare?

Should we adapt the Docker builds to pull in these binaries peharps?

LucasRoesler: I just copied the flow from inlets, i figured you preferred that

alexellis: What is the timing difference?

alexellis:
Screenshot 2020-10-30 at 09 14 25

```
build golang.org/x/sys/internal/unsafeheader: cannot find module for path golang.org/x/sys/internal/unsafeheader
933
```

Can you take a look at the error please?

Makefile
LucasRoesler: i'll just revert. I don't know the timing difference because i haven't gotten a successful build. My goal was consistency between projects to reduce the maintenance effort. But I  don't have strong feelings here.

LucasRoesler: I hav reverted that change

LucasRoesler: Using buildx results in build time: 8 min 51 sec
The last master build was : 4 min 46 sec

alexellis:
I've merged the newer PR from Utsav. Thank you for showing the way with this.

PR #839, opened 2020-11-06
Migrate builds from Travis to Github Actions

### LINKED ISSUE #838: Migrate from Travis to Github Actions
## Expected Behaviour

As proposed by @alexellis we should move all our repos to Github Actions because of it's generous free-tier as we're not able to get enough builds out of the free credits that Travis provides.

## Current Behaviour

We're running all our CI builds using Travis CI

## Possible Solution

PR in the works.

utsavanand2:
Fixes #838

Issue [#1585](https://github.com/openfaas/faas/issues/1585) for ref

Thanks to Lucas Roesler for his PR for buildx support #836

## Motivation and Context

## How Has This Been Tested?

Check out the `ci-only` [build](https://github.com/utsavanand2/faas-cli/runs/1364541475?check_suite_focus=true)
and the `publish` [build](https://github.com/utsavanand2/faas-cli/runs/1363227487?check_suite_focus=truehttps://github.com/utsavanand2/faas-cli/runs/1363227487?check_suite_focus=true)

## Types of changes

## Checklist:


.github/workflows/ci-only.yaml
```diff
@@ -0,0 +1,67 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - 
+        uses: actions/checkout@master
+        with:
+            fetch-depth: 1
+      -  
+        name: Get tags
```
alexellis, line 20: IMHO The tag isn't required for builds / ci.

.github/workflows/ci-only.yaml
```diff
@@ -0,0 +1,67 @@
...
+          build-args: |
+            VERSION=${{ steps.get_tag.outputs.TAG }}
+            GIT_COMMIT=${{ github.sha }}
+          load: true
+          tags: openfaas/faas-cli:${{ github.sha }}
+      - 
+        name: Local docker build (root image)
+        uses: docker/build-push-action@v2
+        with:
+          context: .
```
alexellis, line 47: Context + path can be dropped.

.github/workflows/ci-only.yaml
```diff
@@ -0,0 +1,67 @@
...
+        uses: docker/setup-qemu-action@v1
+      - 
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - 
+        name: Local docker build (non-root image)
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile
```
alexellis, line 34: Does one of the `file: ./Dockerfile` options need to be the root/non-root Dockerfile?

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - 
+        name: Create SHA of binaries
+        run: ./ci/hashgen.sh
+      - 
+        name: Upload binaries and their SHA to Github Release
+        uses: alexellis/upload-assets@0.2.2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          asset_paths: '["./faas-cli*"]'
```
alexellis, line 104: I would put the binaries into `./bin/`

alexellis:
Closes #838 from @LucasRoesler

.github/workflows/ci-only.yaml
```diff
@@ -0,0 +1,67 @@
+name: ci-only
+
+on:
+  push:
+    branches:
+      - "*"
+  pull_request:
+    branches:
+      - "master"
```
alexellis, line 9: Should this be on `*` also?

.github/workflows/ci-only.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - 
+        uses: actions/checkout@master
```
alexellis, line 16: See the setup go action from https://github.com/openfaas/faas-netes/pull/698

.github/workflows/ci-only.yaml
utsavanand2: We can target just the non-root or the root docker image in CI, I just kept it here to be consistent with what we had in travis.yml

utsavanand2:
@alexellis I've resolved the above issues. Does this look like ready to be merged?

.github/workflows/ci-only.yaml
alexellis: Isn't there a new Dockerfile for this like Dockerfile.nonroot?

OK you are controlling it via `--target release` here -> https://github.com/openfaas/faas-cli/blob/master/build.sh

utsavanand2: @alexellis I'm controlling it with
```yaml
with:
   context: .
   file: ./Dockerfile
   target: release
```
and 
```yaml
with:
   context: .
   file: ./Dockerfile
   target: root
```

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,103 @@
...
+      - 
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - 
+        name: Login to Github Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.CR_PAT }}
```
LucasRoesler, line 40: why can't we use the default `${{ secrets.GITHUB_TOKEN }}` i feel like we discussed it in the wednesday call, but I can't remember why

utsavanand2: For the record: we tried it and later found out it doesn't work for forked repos.
https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token

alexellis: So what is `CR_PAT`?

alexellis: If it's not convention then can we use `DOCKER_PASSWORD` as per all the other merged and released builds?

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,103 @@
...
+          platforms: linux/amd64,linux/arm/v7,linux/arm64
+          target: release
+          build-args: |
+            VERSION=${{ steps.get_tag.outputs.TAG }}
+            GIT_COMMIT=${{ github.sha }}
+            NS=${{ steps.get_owner_name.outputs.OWNER_NAME }}
+            REPO=${{ steps.get_repo_name.outputs.REPOSITORY_NAME }}
+          push: true
+          tags: |
+            ghcr.io/${{ steps.get_owner_name.outputs.OWNER_NAME }}/${{ steps.get_repo_name.outputs.REPOSITORY_NAME }}:${{ github.sha }}
```
alexellis, line 56: I'm seeing a lot of duplication. Seems like this could be put into a similar variable and used again?

Dockerfile
```diff
@@ -21,31 +30,41 @@ RUN test -z "$(gofmt -l $(find . -type f -name '*.go' -not -path "./vendor/*"))"
...
     -a -installsuffix cgo -o faas-cli
 
 # CICD stage
-FROM alpine:3.12 as root
+FROM --platform=${BUILDPLATFORM:-linux/amd64} alpine:3.12 as root
+
+ARG NS
+ARG REPO
+
+LABEL org.opencontainers.image.source https://github.com/$NS/$REPO
```
alexellis, line 49: Does that string need to be split out?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,79 @@
...
+        with:
+          context: .
+          file: ./Dockerfile
+          push: false
+          platforms: linux/amd64
+          target: release
+          build-args: |
+            VERSION=latest-dev
+            GIT_COMMIT=${{ github.sha }}
+            NS=openfaas
```
alexellis, line 45: Can't this be a single variable of `GITHUB_REPOSITORY`?

alexellis: Or whatever is the full path to the repo to inject in the special label?

.github/workflows/publish.yaml
utsavanand2: Its a secret for Container Registry.
I have seen CR_PAT in the Github Docs, thought it would be better if we go with it. What do you think?
https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/migrating-to-github-container-registry-for-docker-images#authenticating-with-the-container-registry

.github/workflows/publish.yaml
utsavanand2: Do you mean something like an ENV VAR?

Dockerfile
utsavanand2: No it doesn't need to be split out. I'll put a `=` in between.

alexellis: Can we not inject the complete URL instead?

.github/workflows/publish.yaml
alexellis: The same way you split the string in a step, construct the GHCR main image prefix once instead of 6+ times.

.github/workflows/publish.yaml
alexellis: How faas-netes worked https://github.com/openfaas/faas-netes/tree/master/.github/workflows

is CR_PAT something that works without me doing something on each repo?

alexellis: That link doesn't reference `CR_PAT`, is it the right link?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,79 @@
...
```
alexellis, line 61: Same comment as above

alexellis: REPO_URL=https://github.com/$GITHUB_REPOSITORY

.github/workflows/publish.yaml
alexellis: Sorry I can see where `CR_PAT` is mentioned now. It's not a special variable and has no meaning to Github Actions - can you change to `DOCKER_PASSWORD` please?

alexellis (changes requested):
Changed requested please. What do you say to aiming to have everything ready for a merge on Weds?

Dockerfile
utsavanand2: If you mean injecting the whole `org.opencontainers.image.source=https://github.com/$REPO` into the Dockerfile with build args, then I think it's not possible. The Dockerfile expects two args in LABEL, I get a red squiggly line saying so.
Screenshot 2020-11-17 at 1 54 01 PM

.github/workflows/build.yaml
utsavanand2: Yes we can get the GITHUB_URL into the build args. Sure!

.github/workflows/build.yaml
utsavanand2: Thanks a lot!

Dockerfile
utsavanand2: Okay I get it, by URL you mean this> https://github.com/openfaas/faas-cli/pull/839#discussion_r524514293

.github/workflows/publish.yaml
utsavanand2: Sure!

alexellis (approved):
Your `IMAGE_PREFIX` is much easier to follow than the duplicated strings. Thank you for working on this.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,72 @@
...
+        name: Set Username/Repo as ENV vars
+        run: echo "USER_REPO"=$(echo "$GITHUB_REPOSITORY" | awk '{print tolower($1)}' | sed -e "s/:refs//") >> $GITHUB_ENV
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - 
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - 
+        name: Local docker build (non-root image)
```
alexellis, line 29: Do this in a follow-up PR if you want, but we need to do a multi-arch build upon every commit, not just upon a release tag. This found issues in the past.

utsavanand2: I'll try to put up a PR by tonight to address this 👆

utsavanand2: @alexellis I've resolved it in a PR I've just put up
https://github.com/openfaas/faas-cli/pull/843

PR #840, opened 2020-11-08
Allow configuration of CLI config via env variables

### LINKED ISSUE #832: Proposal: Allow customizing the location that the cli config is stored
When working with @utsavanand2 on his upcoming blog post about Github Actions, it became clear that it would be really helpful to allow customizing the location of the CLI config file. Some initial work is done in #831 that will make the CLI much easier to use in CI environments. But i want to propose some larger changes

1) we could add support for a new env variable `OPENFAAS_CONFIG`. This would be the path to the config directory. If it is empty then it is equivalent to `$HOME/.openfaas` which is the current behavior. This makes the change backwards compatible. 
    Supporting this variable would allow the user to place the config in any location that is needed. In CI it could be preferable that this is in the current working directory. Or, a user might have a CI system running in k8s and want to store and then mount the secret with the configuration. The user can then mount anywhere, e.g. `/openfaas` and then set the `OPENFAAS_CONFIG` to for simplicity
2) Alternatively or additionally, we should consider if we want to support the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) We currently write to `~/.openfaas` but if we want to be compatible with XDG, it _should_ be `~/.config/openfaas`. To implement this, we would want to use a library like https://github.com/adrg/xdg 
    **Note:** Changing this on it's own would _not_ be backwards compatible, but i have a third proposal that would provide that compatibility.   
3) The CLI can look for configuration folder in multiple locations in order: 
    1. look for (hidden) directory `.openfaas` in the current directory 
    2. look for the config file in  `OPENFAAS_CONFIG`
    3. look for directory `openfaas` in `XDG_CONFIG_HOME`
    4. look for the config directory `~/.openfaas` 
        This is the current behavior and provides backwards compatibility for option (2) above.
    This proposal primarily adds the ability to have a project specific/local configuration file (3.i above). This behavior also has similar benefits as (1) in CI environments because the working directory is likely to be shared between different stages of a containerzed workflow while `$HOME` is not.
     This proposal will most likely require a flag that allows the user to specify where they want to write their config during login, `--local` would write to the working directory and when `--local=false` (the default setting) it would go the default user global config location, which is resolved as described above.

All three options can be considered independently but should also be compatible with each other.

Ultimately the goal is to provide more flexibility and control over how the CLI is working. The initial use cases I have considered deal with CI systems. But this could also allow for workflows like https://git-secret.io/

LucasRoesler:
## Description

- Support configuration the faas-cli configuration directory path via
  the `OPENFAAS_CONFIG` env variable. When this value is set, the path
  will be overridden. When empty, we will also check the CI flag, which
  defaults the location to the current working directory. When neither
  `OPENFAAS_CONFIG` or `CI` are set, then the default location is in
  the current HOME folder.
- Various tests have been updated to use this new env variable instead
  of modifying the global variables. Also, the variables are now
  constants to prevent this kind of modification in general.

## Motivation and Context

Resolves #832 

## How Has This Been Tested?

New unit tests verify the behavior. This is also verified by the modifications to the previous unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,6 +378,8 @@ The [scripts/export-sealed-secret-pubcert.sh](https://github.com/openfaas-incuba
 * `OPENFAAS_TEMPLATE_URL` - to set the default URL to pull templates from
 * `OPENFAAS_PREFIX` - for use with `faas-cli new` - this can act in place of `--prefix`
 * `OPENFAAS_URL` - to override the default gateway URL
+* `OPENFAAS_CONFIG` - to override the location of the configuration folder, which contains auth configuration.
+* `CI` - to override the location of the configuration folder, when true, the configuration folder is `.openfaas` in the current working directory. This value is ignored if `OPENFAAS_CONFIG` is set.
 
 ### FaaS-CLI Developers / Contributors
 
--- a/config/config_file.go
+++ b/config/config_file.go
@@ -18,7 +18,20 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-var (
+//AuthType auth type
+type AuthType string
+
+const (
+	//BasicAuthType basic authentication type
+	BasicAuthType = "basic"
+	//Oauth2AuthType oauth2 authentication type
+	Oauth2AuthType = "oauth2"
+
+	// ConfigLocationEnv is the name of he env variable used
+	// to configure the location of the faas-cli config folder.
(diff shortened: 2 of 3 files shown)


alexellis (approved):
Approved

Issue #853, opened 2020-12-17
Remove leaking logs from CLI proxy package 


viveksyngh:
As proxy package is used as SDK it should remove all logging and should return those outputs as results  from those API

## Expected Behaviour

It should not leak logs when SDK are called.

## Current Behaviour

It is having some print/log statements  like  

https://github.com/openfaas/faas-cli/blob/693a50d60c01974c2b5a7c6081f254f385e5b682/proxy/deploy.go#L75

## Possible Solution

Remove these print statements 

Standardised API results to provide more information to the callers  
TODO - Share API format results 

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* FaaS-CLI version ( Full output from: `faas-cli version` ):

* Docker version ( Full output from: `docker version` ):

* Are you using Docker Swarm (FaaS-swarm ) or Kubernetes (FaaS-netes)?

* Operating System and version (e.g. Linux, Windows, MacOS):

* Link to your project or a code example to reproduce issue:

alexellis:
@viveksyngh could you pick this up next please?

viveksyngh:
Yes

alexellis:
Today we have:

```
func (c *Client) DeployFunction(context context.Context, spec *DeployFunctionSpec) int {
```

But the example on the call (GitHub) looked like:

```
func (c *Client) DeployFunction(context context.Context, spec *DeployFunctionSpec) (DeployResponse, http.Response, error) {
```

viveksyngh:
I was looking into making CLI response consistent and I was thinking of following approach. Every API from SDK can return three output which are as following

1. Structured Output for that API for e.g. For list Functions it be a ListFunctionOutput
2. A Response struct which wraps HTTP response object as well as additional information from SDK for advance features like pagination, rate limiting
3. An error which will be basically error message

Response struct looks something as below
```
type Response struct {
	*http.Response
   // Additional information required
}
```

For errors I looked at few of the SDK and looks like if there in any response which is not in the range of 200 to 299, then they consider it as an error and respond with error message for example

https://github.com/google/go-github/blob/bf3472827069b02a1a19bab191b2b19471bf3edb/github/github.go#L807


One challenge I found is some of our API's like DeployFunction or DeleteFunction do not return anything in response body. We need to return something for these interfaces to make them consistent with other interfaces. I am not sure what we information we can return, may be we can return empty struct for now and update it later if required.

LucasRoesler:
[...]

PR #868, opened 2021-02-23
Remove leaking logs from SDK

viveksyngh:
This commit removes leaking logs from the SDK (proxy package) of the CLI by removing
custom print statements within proxy package. It updates method signature for
DeployFunction which now returns deployOutput text along with status code.

Fixes: #853

## Motivation and Context

## Types of changes

## Checklist:

- [ ] I have signed-off my commits with `git commit -s`

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -276,7 +276,9 @@ Error: %s`, fprocessErr.Error())
 			if msg := checkTLSInsecure(services.Provider.GatewayURL, deploySpec.TLSInsecure); len(msg) > 0 {
 				fmt.Println(msg)
 			}
-			statusCode := proxyClient.DeployFunction(ctx, deploySpec)
+			statusCode, deployOutputStr := proxyClient.DeployFunction(ctx, deploySpec)
+			fmt.Println()
+			fmt.Println(deployOutputStr)
 			if badStatusCode(statusCode) {
 				failedStatusCodes[k] = statusCode
 			}
@@ -375,8 +377,9 @@ func deployImage(
 		fmt.Println(msg)
 	}
 
-	statusCode = client.DeployFunction(ctx, deploySpec)
-
+	statusCode, deployOutputStr := client.DeployFunction(ctx, deploySpec)
+	fmt.Println()
+	fmt.Println(deployOutputStr)
 	return statusCode, nil
 }
 
--- a/proxy/deploy.go
+++ b/proxy/deploy.go
@@ -59,21 +59,15 @@ func generateFuncStr(spec *DeployFunctionSpec) string {
 
 // DeployFunction first tries to deploy a function and if it exists will then attempt
 // a rolling update. Warnings are suppressed for the second API call (if required.)
-func (c *Client) DeployFunction(context context.Context, spec *DeployFunctionSpec) int {
+func (c *Client) DeployFunction(context context.Context, spec *DeployFunctionSpec) (int, string) {
 
-	rollingUpdateInfo := fmt.Sprintf("Function %s already exists, attempting rolling-update.", spec.FunctionName)
 	statusCode, deployOutput := c.deploy(context, spec, spec.Update)
 
 	if spec.Update == true && statusCode == http.StatusNotFound {
 		// Re-run the function with update=false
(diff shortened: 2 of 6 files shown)


commands/deploy.go
```diff
@@ -276,7 +276,9 @@ Error: %s`, fprocessErr.Error())
 			if msg := checkTLSInsecure(services.Provider.GatewayURL, deploySpec.TLSInsecure); len(msg) > 0 {
 				fmt.Println(msg)
 			}
-			statusCode := proxyClient.DeployFunction(ctx, deploySpec)
+			statusCode, deployOutputStr := proxyClient.DeployFunction(ctx, deploySpec)
```
alexellis, line 279: I think this is better, but it's not what we discussed on last week's members call. We agreed to refactor to a pattern more similar to the GitHub go SDK where a Go struct, HTTP response,  and error are returned from the call.

proxy/deploy.go
```diff
@@ -59,21 +59,15 @@ func generateFuncStr(spec *DeployFunctionSpec) string {
 
 // DeployFunction first tries to deploy a function and if it exists will then attempt
 // a rolling update. Warnings are suppressed for the second API call (if required.)
-func (c *Client) DeployFunction(context context.Context, spec *DeployFunctionSpec) int {
+func (c *Client) DeployFunction(context context.Context, spec *DeployFunctionSpec) (int, string) {
 
-	rollingUpdateInfo := fmt.Sprintf("Function %s already exists, attempting rolling-update.", spec.FunctionName)
```
alexellis, line 64: Why was this message removed?

proxy/errors.go
```diff
@@ -0,0 +1,229 @@
...
+type StatusReason string
+
+const (
+	StatusReasonUnknown StatusReason = "Unknown"
+
+	// Status code 500
+	StatusReasonInternalError StatusReason = "InternalError"
+
+	// Status code 401
+	StatusReasonUnauthorized StatusReason = "Unauthorized"
```
alexellis, line 34: Could it be http.StatusUnauthorized? https://golang.org/src/net/http/status.go

viveksyngh: We can but we will have challenge in solving two problems, if we are not using `StatusReason`

1. When we will have same http status code being returned for different errors. I think we are not doing that right now but will be helpful in future. 

2. Since we are only having typed errors for only handful of http status codes, so if API server returns an HTTP response code for which we don't have typed error, we are returning that as `UnknownError` embedding HTTP Response code and reason. We are checking for UnknownError based on HTTP response code. Please go through this file, https://github.com/openfaas/faas-cli/pull/868/files#diff-25e38af31c2b62e171bb1b4e9a70546d19aa8fbd180d435435814962e2912600

If we remove StatusReason, to figure out UnknownError we will have to check for list of all HTTP status for which we have typed error.

proxy/deploy.go
viveksyngh: This was added back in subsequent updates.

proxy/deploy.go
```diff
@@ -57,29 +56,48 @@ func generateFuncStr(spec *DeployFunctionSpec) string {
...
+	}
+
+	var deployResponse DeployResponse
+	if err == nil {
+		deployResponse.Message = fmt.Sprintf("Deployed. %s.", res.Status)
+		deployResponse.URL = fmt.Sprintf("%s/function/%s", c.GatewayURL.String(), generateFuncStr(spec))
+	}
 
-	if spec.Update == true && statusCode == http.StatusNotFound {
+	if spec.Update == true && IsNotFound(err) {
```
viveksyngh, line 82: Example of checking types error.

commands/errors.go
```diff
@@ -24,3 +27,13 @@ func checkTLSInsecure(gateway string, tlsInsecure bool) string {
 	}
 	return ""
 }
+
+//actionableErrorMessage print actionable error message based on APIError check
+func actionableErrorMessage(err error) error {
```
viveksyngh, line 32: A generic error message functions which can be used in all commands to print error message for generic APIError like 403, 500 etc

PR #885, opened 2021-04-29
Refactor Proxy Secrets CRUD to new 3 value returns

Waterdrips:
## Description

This commit refactors the proxy secrets code to return 2 (or 3, for
list) values of (optional) data, *http.Response, error (either generic,
or OpenFaaS error type if its not a json unmarshal or other i/o type
error)

This can be used as showin in the tests, if you want to get status code
now you have access to the http response, or you can see what type the
returned error is, if its OpenFaaS error it should contain the http
status code and response (from upstream)

New unit tests added and faas-cli tested with 401, i/0 (no gw running)
and OK responses against faas-netes

## Also
I removed the error return on 
```go
type ClientAuth interface {
	Set(req *http.Request)
}
```
(Set used to optionally return an error, but we never returned one anywhere)

Signed-off-by: Alistair Hey <alistair@heyal.co.uk>

## Motivation and Context

-

## How Has This Been Tested?

see description, unit tests and locally against faas-netes

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
signature of proxy changes, anyone using it and upgrading will need to use new return signatures.
## Checklist:

### DIFF
--- a/proxy/errors.go
+++ b/proxy/errors.go
@@ -1,0 +1,33 @@
+package proxy
+
+import (
+	"fmt"
+	"net/http"
+)
+
+type OpenFaaSError struct {
+	StatusCode int
+	Message    string
+}
+
+func (e OpenFaaSError) Error() string {
+	switch e.StatusCode {
+	case http.StatusUnauthorized, http.StatusForbidden:
+		return fmt.Sprintf("unauthorized access, run \"faas-cli login\" to setup authentication for this server - http status %d", e.StatusCode)
+	case http.StatusNotFound:
+		return fmt.Sprintf("resource not found - http status %d", e.StatusCode)
+	case http.StatusInternalServerError:
+		return fmt.Sprintf("openfaas encountered an internal server error - http status: %d", e.StatusCode)
+	case http.StatusBadRequest:
+		return fmt.Sprintf("bad request - http status: %d", e.StatusCode)
+	default:
+		return fmt.Sprintf("%s - http status: %d", e.Message, e.StatusCode)
+	}
+}
+
+func NewOpenFaaSError(message string, statusCode int) error {
+	return OpenFaaSError{
+		StatusCode: statusCode,
+		Message:    message,
+	}
+}
--- a/proxy/secret.go
+++ b/proxy/secret.go
@@ -16,7 +16,7 @@ const (
 )
 
 // GetSecretList get secrets list
-func (c *Client) GetSecretList(ctx context.Context, namespace string) ([]types.Secret, error) {
+func (c *Client) GetSecretList(ctx context.Context, namespace string) ([]types.Secret, *http.Response, error) {
 	var (
 		results    []types.Secret
 		err        error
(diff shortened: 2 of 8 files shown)


proxy/auth.go
```diff
@@ -19,19 +19,17 @@ type BasicAuth struct {
 	password string
 }
 
-func (auth *BasicAuth) Set(req *http.Request) error {
+func (auth *BasicAuth) Set(req *http.Request) {
```
alexellis, line 22: I can't remember why we had `error` here, but there may have been a good reason. @viveksyngh can you speak to this?

proxy/secret.go
```diff
@@ -30,161 +30,136 @@ func (c *Client) GetSecretList(ctx context.Context, namespace string) ([]types.S
...
 
 	if res.Body != nil {
 		defer res.Body.Close()
 	}
+	bytesOut, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return res, err
+	}
 
-	switch res.StatusCode {
```
alexellis, line 93: I'm trying to find where this code moved to?

`unable to find secret` <- what happens for a 404 going forward, if we merge this?

proxy/secret.go
```diff
@@ -30,161 +30,136 @@ func (c *Client) GetSecretList(ctx context.Context, namespace string) ([]types.S
...
-	case http.StatusOK, http.StatusAccepted:
-		output += fmt.Sprintf("Updated: %s\n", res.Status)
-		break
-
-	case http.StatusNotFound:
-		output += fmt.Sprintf("unable to find secret: %s", secret.Name)
-
-	case http.StatusUnauthorized:
-		output += fmt.Sprintf("unauthorized access, run \"faas-cli login\" to setup authentication for this server")
```
alexellis, line 102: Same question for this one. Shouldn't this either be: part of the typed openfaas error (less repetition), or moved into the command itself?

alexellis (changes requested):
Couple of questions about where messages have gone, or what the plan is.

proxy/auth.go
viveksyngh: As I recall. Idea was to allow an interface with which we can use different authentication providers e.g the token can be obtained from a file or something or an API. In such case we might we might encounter some errors which can be returned by the method.

proxy/errors.go
```diff
@@ -0,0 +1,10 @@
+package proxy
+
+type OpenFaaSError struct {
+	StatusCode int
```
viveksyngh, line 4: Does it make sense to have StatusCode in here as we are returning HttpResponse object ?

proxy/auth.go
viveksyngh: If we are not checking for errors in where we are setting the auth then it might be bug

proxy/secret.go
Waterdrips: You get an error, which has status code field with 404 in it, which you switch on as a library consumer. The lib code doesnt really care what the status is if its not a "success"

proxy/errors.go
Waterdrips: yes. We can start to add logic onto the error later like `IsRetryable()` or `IsAuthinticationError()` to help people out.

proxy/auth.go
Waterdrips: for basic auth (req.SetBasicAuth()) AND barerAuth (header.set) the funcs dont return an error, so there's nothing to check

Waterdrips: Happy to put `error` back on, just spotted we didnt use it anywhere

Waterdrips:
Added `error` back onto auth interface to keep this PR single responsibility

LucasRoesler:
@Waterdrips I am planning to take a look tonight or tomorrow. Today is supposed to be a little bit sunny while tomorrow will be rainy, so we have a small road trip today.

proxy/secret.go
```diff
@@ -30,161 +30,123 @@ func (c *Client) GetSecretList(ctx context.Context, namespace string) ([]types.S
...
 	if err != nil {
-		return nil, fmt.Errorf("cannot connect to OpenFaaS on URL: %s", c.GatewayURL.String())
+		return nil, res, fmt.Errorf("cannot connect to OpenFaaS on URL: %s", c.GatewayURL.String())
 	}
 
 	if res.Body != nil {
 		defer res.Body.Close()
 	}
 
+	bytesOut, err := ioutil.ReadAll(res.Body)
```
LucasRoesler, line 45: have we upgraded to 1.16 yet? they moved most of the ioutil packages, this is now in `io.ReadAll`, how many Go versions do we want to support?

LucasRoesler: Also, in theory, Body could be nil here right? I don't think `ReadAll` handles a `nil` reader so in some odd edge cases we would get a panic here. We didn't hndle it before, but perhaps it something we want to improve here and in the other methods?

https://golang.org/src/io/io.go?s=19249:19287#L616

proxy/errors.go
```diff
@@ -0,0 +1,33 @@
+package proxy
+
+import (
+	"fmt"
+	"net/http"
+)
+
+type OpenFaaSError struct {
+	StatusCode int
+	Message    string
```
LucasRoesler, line 10: I think it would be good to include a copy of the response headers here as well, depending on the configuration of the system there could be useful debugging/tracing headers or other data that the caller would want to inspect when there is an error.  In that case it is probably a good idea to change 

```go
func NewOpenFaaSError(message string, statusCode int) error {}
```

to

```go
func NewOpenFaaSError(message string, resp *http.Response) error {}
```

PR #888, opened 2021-05-20
Add function YAML config to ignore paths during shrinkwrap

tatemz:
## Description

This change handles an `ignore_paths` configuration array that prevents
specified folders and files from being copied into the handler build
folder.

This is useful for node applications where the process of copying a
handler's node_modules folder can take a long time. In such case, the
node_modules folder is not needed as it is ignored during the Docker
build process anyway.

This change also applies to file paths that are copied from the
configurations.copy array.

## Motivation and Context

- [x] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md)) https://github.com/openfaas/faas-cli/issues/889

## How Has This Been Tested?

Go tests and example in `sample` folder and `stack.yml`. The following commands should show that the `sample/nodejs-echo/node_modules/foo` is not added to the `build` folder

```sh
$ ./bin/faas-cli-darwin build
$ ls -lha build/nodejs-echo/function/node_modules
```

## Types of changes

## Checklist:

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -24,7 +24,7 @@ const AdditionalPackageBuildArg = "ADDITIONAL_PACKAGE"
 
 // BuildImage construct Docker image from function parameters
 // TODO: refactor signature to a struct to simplify the length of the method header
-func BuildImage(image string, handler string, functionName string, language string, nocache bool, squash bool, shrinkwrap bool, buildArgMap map[string]string, buildOptions []string, tagMode schema.BuildFormat, buildLabelMap map[string]string, quietBuild bool, copyExtraPaths []string) error {
+func BuildImage(image string, handler string, functionName string, language string, nocache bool, squash bool, shrinkwrap bool, buildArgMap map[string]string, buildOptions []string, tagMode schema.BuildFormat, buildLabelMap map[string]string, quietBuild bool, copyExtraPaths []string, ignorePaths []string) error {
 
 	if stack.IsValidTemplate(language) {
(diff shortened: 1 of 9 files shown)


tatemz:
Hello, consider this feature as a proposal for new functionality to address https://github.com/openfaas/faas-cli/issues/889

A few dev notes:
  - I chose to use `ignore_paths` for lack of a better term. I'd love some better suggestions
  - I chose to ignore paths based on their destination rather than their source in order to handle files that may be copied into a function handler via the [`copy`](https://docs.openfaas.com/reference/yaml/#copy) configuration
  - I think this use case primarily addresses Node projects, but maybe there are more usecases
  
An alternative I thought through was to evaluate glob patterns and ignore patterns found in the `.dockerignore` file, which ideally should contain similar entries.

krakrjak:
This is a really important ergonomic and space saving feature. Is it possible to get some 👀 on this one with feedback. I've been using this version of the faas-cli for a while and it is very nice to not have to worry about cleaning up explicitly after local debugging.

alexellis:
> I chose to use ignore_paths for lack of a better term. I'd love some better suggestions

There is already a convention for this, placing a file in the upper directory:

.gitignore
.dockerignore

These cannot be executed recursively with Docker, so perhaps an implementation for the CLI should work the same, if you wanted to ignore `node_modules`.

By removing the folder from your builds, do you then cause the latency to be taken over into Docker or your builder and its npm step? What about private `npm` modules? How are you handling them?

PR #902, opened 2021-11-07
Fixes #852 optionally add last newline

### LINKED ISSUE #852: Append flag doesnt add newline if not present in file
## Expected Behaviour

The append flag shuld a function correctly if the file doesn't have a newline at the end

## Current Behaviour

When using `faas-cli new --append stack.yml` i have found that this doesnt append a newline if one is not present in the file

I therefore created a new function and it appended like this

```yaml
  route53:
    lang: golang-middleware
    handler: ./route53
    image: route53:latest  route54:       # <-- appended to here without newline
    lang: golang-middleware
    handler: ./route54
    image: route54:latest
```

## Possible Solution

check the stack file for a newline, if not present add one?

1. faas-cli new --lang golang new -f stack.yml
2.remove newlines from the stack.yml
3.faas-cli new --lang golang new2 --append stack.yml
4. check file, its not added a newline before the func so its squashed the the last line of the first definition

## Your Environment

```
  ___                   _____           ____
 / _ \ _ __   ___ _ __ |  ___|_ _  __ _/ ___|
| | | | '_ \ / _ \ '_ \| |_ / _` |/ _` \___ \
| |_| | |_) |  __/ | | |  _| (_| | (_| |___) |
 \___/| .__/ \___|_| |_|_|  \__,_|\__,_|____/
      |_|

CLI:
 commit:  598336a0cad38a79d5466e6a3a9aebab4fc61ba9
 version: 0.12.21
Error with YAML file
```

* Operating System and version (e.g. Linux, Windows, MacOS):
Linux

kylos101:
## Description
Fixes #852

## Motivation and Context

## How Has This Been Tested?

So far, just manually, but will write unit tests after getting feedback on the fix.
Tested using the stack.yml file in this repo, ran iterations where it did not have trailing new lines, and did - both behave the same now. 

Testing was done on Ubuntu; no Windows testing has been done yet.

I intentionally do this check early, [in the same block as the other append checks](https://github.com/openfaas/faas-cli/compare/master...kylos101:852?expand=1#diff-2b2b11ceabb35f054b2921b6cd82a3293df75820a3590be3c37ff4789c85076dR156), so that if we encounter an error while trying to fix the trailing new lines, we do not leave behind a handler folder.

## Types of changes

## Checklist:

- [ ] I have signed-off my commits with `git commit -s` (no, but I will, I look forward to Derek telling me abou tthis) :+1: 
- [ ] I have added tests to cover my changes. (not yet, I plan to, but first want to get feedback on my fix) :+1: 


kylos101:
@alexellis take a peek and let me know what you think? Once I have your :+1: on the fix itself, I'll proceed with unit tests. Thanks!

commands/new_function.go
```diff
@@ -346,3 +351,33 @@ Cannot have duplicate function names in same yaml file`, functionName, appendFil
 
 	return nil
 }
+
+func addLastNewline(fileName string) error {
```
alexellis, line 355: This seems to trying to optimise the problem by reading the last couple of bytes.

Perhaps it could be more paintable and much easier to read if the code read the whole file into memory and used string.HasSuffix()?

Out of interest, where did you source this technique, or did you come up with it yourself?

commands/new_function.go
```diff
@@ -346,3 +351,33 @@ Cannot have duplicate function names in same yaml file`, functionName, appendFil
...
+	} else if bytesRead != 1 {
+		return fmt.Errorf("read unexpected # of bytes in '%s'", fileName)
+	}
+
+	hasTrailingNewline := string(buffer) == "\n"
+	if !hasTrailingNewline {
+		// add 2 trailing newlines
+		// this is consistent with append behavior when last byte is already \n
+		file.Seek(0, 2)
+		file.Write([]byte("\n\n"))
```
alexellis, line 380: Some OSes may use `\r\n`https://en.wikipedia.org/wiki/Newline

I.e LF vs CR LF

commands/new_function.go
```diff
@@ -152,6 +152,11 @@ Download templates:
 		}
 
 		fileName = appendFile
+
+		newLineErr := addLastNewline("./" + fileName)
```
alexellis, line 156: I would inline this and use the word `err`

```
if err := ... ; err != nil {
    return err
}
```

alexellis (commented):
I didn't realise this would be as complicated as it's turned out. I'm leaving some comments.

commands/new_function.go
kylos101: I came up with it on my own. :shrug: Why? I figured it would be fast, regardless of file size. I'm not sure how large stack.yml files get in the real world.

I hadn't considered reading the whole file into a string and then using string.HasSuffix(). Thank you for the suggestion, I'll give that a whirl.

commands/new_function.go
kylos101: Thanks for the suggestion!

commands/new_function.go
kylos101: :+1:  I thought the same thing, and bumped [into this when searching](https://stackoverflow.com/a/14494187/326313), and if you [look at line 1171](https://golang.org/src/fmt/print.go) here, you'll see even the std lib doesn't write `\r\n` vs `\n`. :exploding_head: 

I'm not sure how to proceed. Where else do we write `\r\n` versus `\n`, or is there a library/standard method that does this for us?

kylos101: PS: Thanks again for the review! :100:

commands/new_function.go
kylos101: Resolved in 3a2cd9c

commands/new_function.go
kylos101: Resolved in 3a2cd9c

commands/new_function.go
kylos101: I'll setup a Windows VM and test with Git Bash to be safe. :) [Given this thread](https://github.com/golang/go/issues/28822#issuecomment-441430941), I assume Powershell and Git Bash can work with `\n`. Talk to you later.

kylos101: Works like a champ in Windows. In the below pic, **new2** and **new4** were appended when there were no trailing new lines. **new3** was appended when there were trailing new lines. 

image

I'm using Powershell here to test, and assume I'd get the same result wiht git bash.

kylos101: I'll test with git bash tomorrow, and push unit tests.

kylos101:
Hi @alexellis this should ready for review now (no rush, just whenever you get a chance).

FYI, while testing I found this issue #906 , let me know if you'd like me to draft a fix in a separate PR?

alexellis (dismissed):
Approved

commands/new_function_test.go
```diff
@@ -423,3 +428,58 @@ func Test_getPrefixValue_Flag(t *testing.T) {
 		t.Errorf("want %s, got %s", want, val)
 	}
 }
+
+func Test_addEofNewlines_works(t *testing.T) {
```
alexellis, line 432: Where is this getting tested?

I'd expect something like if wantedYAML == os.ReadFile("stack.yml")

kylos101: Nice catch, [I just added this](https://github.com/openfaas/faas-cli/pull/902/files#diff-d5819756d14b25e2e55fd81bf9ac16c54dcb28c81b67c66da0dab67e7cc5c71cR480-R488). Let me know if anything else is needed?

Issue #958, opened 2023-04-24
Openfaas might be incompatible for docker versions 23.0.0 and above


NikhilSharmaWe:
## Expected Behaviour

`faas-cli up -f pydict.yml` works fine and builds an image

## Current Behaviour

- But for docker versions 23.0.0 or above the faas-cli up command shows this error:

```
Errors received during build:
- [pydict] received non-zero exit code from build, error: failed to fetch metadata: fork/exec /usr/local/lib/docker/cli-plugins/docker-buildx: no such file or directory

ERROR: BuildKit is enabled but the buildx component is missing or broken.
       Install the buildx component to build images with BuildKit:
       https://docs.docker.com/go/buildx/ 
```
- Build works fine when docker version `20.10.24` is used.

## Who is this for?

Better user experience

What company is this for? Are you listed in the [ADOPTERS.md](https://github.com/openfaas/faas/blob/master/ADOPTERS.md) file?

NO

## Are you a GitHub Sponsor (Yes/No?)

Check at: https://github.com/sponsors/openfaas
- [ ] Yes
- [x] No

## List All Possible Solutions and Workarounds

- The issue is probably because of the changes made in docker `23.0.0` release regarding the docker build: https://docs.docker.com/engine/release-notes/23.0/#new
- So I think we can tackle this issue by setting env var [`DOCKER_BUILDKIT`](https://github.com/openfaas/faas-cli/blob/master/builder/build.go#L87) to `0` when docker version is `23.0.0` or above 
## Steps to Reproduce (for bugs)

Can reproduce the issue by trying to build image through openfaas with docker release version `23.0.0` or above.

## Your Environment

* FaaS-CLI version ( Full output from: `faas-cli version` ):
0.16.3
* Docker version ( Full output from: `docker version` ):
23.0.4 (when error occurs)
* Are you using Docker Swarm (FaaS-swarm ) or Kubernetes (FaaS-netes)?
Kubernetes
* Operating System and version (e.g. Linux, Windows, MacOS):
wsl: Ubuntu 20.04.5 LTS
* Link to your project or a code example to reproduce issue:

alexellis:
We'll take a look at this, thank you for sharing.

I'm not sure if you could run `docker buildx uninstall`?

alexellis:
Could you make a suggestion in Which Solution Do You Recommend?

Note that `faas-cli publish` requires `DOCKER_BUILDKIT=1`

PR #960, opened 2023-04-25
fix the bug with build operation in faas-cli for docker versions 23.0.0 and above

NikhilSharmaWe:
## Description

For docker versions `23.0.0` and above the env var `DOCKER_BUILDKIT` is equal to 1 by default for using the docker build kit, which used `docker buildx build` instead of `docker build`. This PR fixes the bug with the faas-cli build operation and docker build for docker version 23.0.0 and above. PTAL: #958 for more info.

## Motivation and Context

Fixes: #958
## How Has This Been Tested?

I have tested the changes. Now the `faas-cli build` works fine also with docker version `23.0.0` and above.
## Types of changes

## Checklist:

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -85,6 +85,8 @@ func BuildImage(image string, handler string, functionName string, language stri
 		envs := os.Environ()
 		if mountSSH {
 			envs = append(envs, "DOCKER_BUILDKIT=1")
+		} else {
+			envs = append(envs, "DOCKER_BUILDKIT=0")
 		}
 
 		task := v1execute.ExecTask{


NikhilSharmaWe:
I think then we can update the build to also use `buildx build` command as publish and set `DOCKER_BUILDKIT` to 1 in all cases.
WDYT?

alexellis:
The `buildx build` command cannot be used, since it requires a specific modern docker version, right?

But what is the issue with `DOCKER_BUILDKIT=1 docker build`?

NikhilSharmaWe:
If we set DOCKER_BUILDKIT=1 then `docker buildx build` is used instead of `docker build`, so that won't work for build in faas-cli, since it uses `docker build` unlike publish which uses `docker buildx build`.

alexellis:
> @alexellis If we set DOCKER_BUILDKIT=1 then docker buildx build is used instead of docker build, so that won't work for build in faas-cli, since it uses docker build unlike publish which uses docker buildx build.

I often set `DOCKER_BUILDKIT=1` and it does work for me - same for @welteki 

Perhaps you need to run `docker buildx install`? `uninstall` reverts that change.

NikhilSharmaWe:
@alexellis
- After `docker buildx uninstall`, it works. Could be that the `docker buildx` and docker engine I installed were installed separately. Not sure.

- Also, should we add a `--buildkit=true` flag or `DOCKER_BUILDKIT=1 faas-cli build` is fine for now.

Issue #952, opened 2023-02-08
feature: add watch mode to faas-cli up workflow


LucasRoesler:
## Expected Behaviour

We want to add a flag to the `faas-cli up` command so that it can run continuously and re-build and deploy when files are changed.

For example `faas-cli up --watch` will build and deploy the functions in the `stack.yaml` and then continue watching for file change notifications and rebuild and deploy the stack when an event is seen.

## Current Behaviour

`faas-cli up` is a one-off workflow right now.

## Why do you need this?

Make the dev-loop shorter. 

## Context

This was discussed in the community call as a feature that we would like to have and it would be a test ground for also adding the watch behavior to the `faas-cli local-run`

LucasRoesler:
For the file notifications, [skaffold is using](https://github.com/GoogleContainerTools/skaffold/blob/main/go.mod#LL179) the [fsnotify package](https://pkg.go.dev/github.com/fsnotify/fsnotify) and it looks like a solid package.

alexellis:
Here's the equivalent command with Google Skaffold which is well known for giving an "auto-build" workflow:

https://github.com/GoogleContainerTools/skaffold/issues/8396

Example function / config for Skaffold with faas-cli: https://github.com/alexellis/openfaas-skaffold-faas-cli

After we have a working prototype for faas-cli up, local-run would probably be a good place to go next.

alexellis:
An example with a library that doesn't need CGO and it's used by Skaffold:

https://github.com/alexellis/fstail/

NikhilSharmaWe:
I would like to work on this.

PR #763, opened 2020-02-09
File watch

johnmccabe:
## Description
Adds a `--watch` flag to the `up` command which will watch the filesystem for changes and trigger an `up` command, this includes a 500ms debounce interval to avoid unnecessary builds when many files change in quick succession.

Ignores changes some standard dirs, `build`, `template`, `.git`, and some temp files `~`, can be appended to with the `ignore-dir` and `ignore-suffix` args.

Currently does not honour filter argument when working in a multi function dir.

Changes are watched relative to the `pwd`.

## Motivation and Context

- [x] #472

## How Has This Been Tested?
Tested locally on Windows10.

Needs feedback from Linux/OSX users.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [ ] I have signed-off my commits with `git commit -s`
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #962, opened 2023-05-02
add new watch flag for up command

NikhilSharmaWe:
## Description

This PR works on adding a new `watch` flag for `up` command. When this flag is set to true then the cli is run continuously and re-build and deploy when files are changed.

**Demo** : https://asciinema.org/a/EYU177ju23V3XncUcxhxMGDRm

## Motivation and Context

Fixes: #952
## How Has This Been Tested?

Example used: https://www.openfaas.com/blog/golang-deep-dive/
- I run `faas-cli up -f homepage.yml --watch=true` locally and it works as expected.
- When I make changes in the `./homepage/static/index.html` file, the function is rebuilt and redeployed, and the deployment reflects the changes made in the html.

## Types of changes

## Checklist:


commands/up.go
```diff
@@ -67,14 +73,62 @@ func upHandler(cmd *cobra.Command, args []string) error {
...
+				return err
+			}
+			if info.IsDir() {
+				return watcher.Add(path)
+			}
+			return nil
+		})
+
+		logger := log.New(os.Stderr, "", log.LstdFlags)
+		logger.SetFlags(log.Ldate | log.Ltime | log.LUTC)
```
alexellis, line 105: What's happening here?

commands/up.go
```diff
@@ -67,14 +73,62 @@ func upHandler(cmd *cobra.Command, args []string) error {
...
+				if !ok {
+					return fmt.Errorf("watcher's Events channel is closed")
+				}
+
+				if event.Op == fsnotify.Write {
+					if err := runBuild(cmd, args); err != nil {
+						logger.Printf("%s %s", functionName, err)
+					}
+
+					if err := runDeploy(cmd, args); err != nil {
```
alexellis, line 119: If there was an error in the build, the deploy shouldn't happen at all.

Also, where is the push?

A deployment requires a remote image to be present - unless --skip-push has been given.

commands/up.go
```diff
@@ -67,14 +73,62 @@ func upHandler(cmd *cobra.Command, args []string) error {
...
+		logger.SetFlags(log.Ldate | log.Ltime | log.LUTC)
+
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return fmt.Errorf("watcher's Events channel is closed")
+				}
+
+				if event.Op == fsnotify.Write {
```
alexellis, line 114: I think you need to debounce these notifications to prevent two identical builds happening if two files changed at the same time.

See how @johnmccabe handled this - there is a very old PR that you should find from him.

commands/up.go
```diff
@@ -67,14 +73,62 @@ func upHandler(cmd *cobra.Command, args []string) error {
...
+			return err
+		}
+
+		watcher, err := fsnotify.NewWatcher()
+		if err != nil {
+			return err
+		}
+		defer watcher.Close()
+
+		_ = filepath.Walk("./", func(path string, info os.FileInfo, err error) error {
```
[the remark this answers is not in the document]
NikhilSharmaWe: Fixed it

commands/up.go
NikhilSharmaWe: Made changes for debouncing the notifications.

alexellis: https://github.com/openfaas/faas-cli/pull/763/files

commands/up.go
```diff
@@ -67,14 +73,62 @@ func upHandler(cmd *cobra.Command, args []string) error {
...
+			if info.IsDir() {
+				return watcher.Add(path)
+			}
+			return nil
+		})
+
+		logger := log.New(os.Stderr, "", log.LstdFlags)
+		logger.SetFlags(log.Ldate | log.Ltime | log.LUTC)
+
+		for {
```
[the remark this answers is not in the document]
NikhilSharmaWe: Added code for handling OS signals for interruption and termination. 
PTAL: https://github.com/NikhilSharmaWe/faas-cli/blob/up-watch/commands/up.go#L117-L128

commands/up.go
```diff
@@ -67,17 +77,88 @@ func upHandler(cmd *cobra.Command, args []string) error {
...
+		signalChannel := make(chan os.Signal, 1)
+		signal.Notify(signalChannel, os.Interrupt, syscall.SIGTERM)
+
+		go func() {
+			<-signalChannel
+
+			watcher.Close()
+
+			os.Exit(0)
+		}()
```
LucasRoesler, line 126: I don't think this exit is needed, after calling close this goroutine will complete and be cleaned up correct. I generally try to avoid calling Exit directly because people often forget where the exit is happening, instead of allowing the function to run to completion. 

Instead, the `for`-loop should properly stop when the `watcher` closes or the signal is submitted.

Looking at this again, this goroutine is not needed at all, this could simply be a `case` in the `switch` 

```go
		for {
			select {
			case event, ok := <-watcher.Events:
				if !ok {
					return fmt.Errorf("watcher's Events channel is closed")
				}

				if event.Op == fsnotify.Write {
					d(func() {
						if err := runBuild(cmd, args); err != nil {
							logger.Fatalf("%s %s", functionName, err)
						}

						if !skipPush {
							if err := runPush(cmd, args); err != nil {
								logger.Fatalf("%s %s", functionName, err)
							}
						}

						if err := runDeploy(cmd, args); err != nil {
							logger.Fatalf("%s %s", functionName, err)
						}
					})
				}

			case err, ok := <-watcher.Errors:
				if !ok {
					return fmt.Errorf("watcher's Errors channel is closed")
				}
				return err
			case <-signalChannel:
				watcher.Close()
				return nil
			}
		}
```

commands/up.go
```diff
@@ -67,17 +77,88 @@ func upHandler(cmd *cobra.Command, args []string) error {
...
+						if !skipPush {
+							if err := runPush(cmd, args); err != nil {
+								logger.Fatalf("%s %s", functionName, err)
+							}
+						}
+
+						if err := runDeploy(cmd, args); err != nil {
+							logger.Fatalf("%s %s", functionName, err)
+						}
+					})
```
LucasRoesler, line 152: Instead of duplicating this flow, i would consider defining a new function 

```go
// upRunner is constructs a handler function that runs the build, push and deploy commands one time.
func upRunner(cmd *cobra.Command, args []string) func() error {
	return func() error {
		if err := runBuild(cmd, args); err != nil {
			return err
		}

		if !skipPush {
			if err := runPush(cmd, args); err != nil {
				return err
			}
		}

		if !skipDeploy {
			if err := runDeploy(cmd, args); err != nil {
				return err
			}
		}
		return nil
	}
}
```

This is responsible for defining the actual up logic and can be called multiple times, specifically, inside the `upHandler` you can now do this 

```go
	handler := upRunner(cmd, args)

	err := handler()
	if err != nil {
		return err
	}
```

and 

```go
				if event.Op == fsnotify.Write {
					d(func() {
						if err := handler(); err != nil {
							logger.Printf("%s %s", functionName, err)
						}
					})
				}
```

I would avoid the `FatalF` here so that the watch can survive a typo in the code, etc. Let the watch run until the user explicitly stops it. 

You could also consider adding this helper function 

```go
func logError(f func() error) func() {
	return func() {
		if err := f(); err != nil {
			log.Println(err)
		}
	}
}
```

which then allows you to further simplify the debounce to look like this 

```go
				if event.Op == fsnotify.Write {
					d(logError(handler))
				}
```

LucasRoesler (changes requested):

commands/up.go
```diff
@@ -67,17 +77,88 @@ func upHandler(cmd *cobra.Command, args []string) error {
...
+		}
+		defer watcher.Close()
+
+		err = filepath.Walk("./", func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if info.IsDir() {
+				return watcher.Add(path)
+			}
```
LucasRoesler, line 109: I think it makes sense to have a list of non-watched paths, at a minimum, `.git` and the `build` folder should be omitted. 

```go
var skipWatch = map[string]struct{}{
    ".git",
    "build",
}
```

In fact, it is probably reasonable to omit _all_ hidden directories, for example `.vscode`  and `.pytest_cache` are not relevant. There are probably other paths that we should also reasonably ignore, in a  python project `__pycache__` could also be ignored and in a nodejs project, maybe `node_modules`. Perhaps we could read the `.gitignore` file and omit any path that matches one of the patterns. This would allow maximum flexibility and have a high probably of matching files that don't actually indicate a change in the functions. 

Alternatively, the stack file should be read and watches only created for each [`handler`](https://docs.openfaas.com/reference/yaml/#function-handler), but this is not enough, you would  also need to include any paths from the [`copy` field](https://docs.openfaas.com/reference/yaml/#copy), if it exists. 

For simplicity it is probably enough to skip `.git` and any path that matches a pattern in the `.gitignore`

commands/up.go
alexellis: @NikhilSharmaWe could you respond to the comments please and click "resolve" if you have resolved the comments?

NikhilSharmaWe: Thanks @LucasRoesler for your reviews

alexellis:
Comments from community call:

1) What do we do if a save happens during a build? See also compose/skaffold/webpack - https://docs.docker.com/compose/file-watch/ - https://skaffold.dev/docs/references/yaml/ - https://webpack.js.org/concepts/hot-module-replacement/ via @alexellis 
2) How do we want to work with ignore paths? i.e. ./build/ and ./.git/ via @LucasRoesler - might need to be able to add others via the command line (perhaps by parsing .gitignore?)

commands/up.go
```diff
@@ -64,20 +74,87 @@ func preRunUp(cmd *cobra.Command, args []string) error {
...
+		if err != nil {
 			return err
 		}
-		fmt.Println()
-	}
-	if !skipDeploy {
-		if err := runDeploy(cmd, args); err != nil {
+		defer watcher.Close()
+
+		err = filepath.Walk("./", func(path string, info os.FileInfo, err error) error {
```
alexellis, line 92: Perhaps also inline this:

if err:= ; err!=nil {

}

alexellis:
We need to show some examples of Control + C during a build.

commands/up.go
NikhilSharmaWe: @LucasRoesler Made changes for skipping `.git` file and files specified in the `.gitignore` file for the watch. 
PTAL: https://github.com/openfaas/faas-cli/pull/962/commits/5235a62b1b2599ae50ce5355915fe833d1e1a8b8

commands/up.go
NikhilSharmaWe: Resolved

commands/up.go
NikhilSharmaWe: Since `build` is not initially present in the dir, so I think we need to also go through the `filpath.Walk` after first build for skipping the build dir for skipping watch on that. Please give your thoughts @LucasRoesler @alexellis.

commands/up.go
```diff
@@ -64,20 +77,133 @@ func preRunUp(cmd *cobra.Command, args []string) error {
...
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return fmt.Errorf("watcher's Events channel is closed")
+				}
+
+				if event.Op == fsnotify.Write {
+					d(func() {
+						if err := handler(); err != nil {
```
LucasRoesler, line 138: @alexellis i have been thinking about command cancelling a bit more, in short, it will require a refactor in at least two project. 

To have the ability to cancel concurrent runs, we would need to pass and manage a Context through all of the build layers so that we can pass it to the actual build command. 

This obviously requires a lot of changes in this project, but that isn't enough. 

Back when you first wrote the builder you opted to wrap the `exec.Cmd` with https://github.com/alexellis/go-execute (which hasn't really seen or needed any changes in the last 3 years) 

Most importantly go-execute does not support passing a context or cancelling a command. The ideal situation would be to use https://pkg.go.dev/os/exec#CommandContext so that a context can be passed down to the command, allowing cancellation. 

Either go-execute needs an update, which can then be used here, or faas-cli needs to refactor back to the stdlib

LucasRoesler (changes requested):

commands/up.go
```diff
@@ -64,20 +77,133 @@ func preRunUp(cmd *cobra.Command, args []string) error {
...
+			return err
+		}
+
+		if err = filepath.Walk("./", func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+
+			for _, pattern := range patterns {
+				match, err := filepath.Match(pattern, path)
```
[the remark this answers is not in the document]
NikhilSharmaWe: If we will use it.

For example:

The **`.gitignore`** file contains **`build`**.

Suppose there is a function named **`gobuilder`** or something containing **`build`**, then the watcher will skip watching the files in the **`gobuilder`** folder too. This is what we do not want to happen. Therefore I think we should skip the folders exactly mentioned in the **`.gitignore`** because we can afford to watch a not required file but cannot skip watching a required file.

Please give your thoughts, there might be something I am missing here or unable to understand your suggestion.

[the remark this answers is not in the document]
LucasRoesler: hmm, i wonder how git handles this, maybe they do something like "if the pattern contains no stars, test if the path base == the pattern", something like this https://pkg.go.dev/path/filepath#Base

```go
match = match || filepath.Base(path) == pattern
```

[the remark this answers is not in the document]
NikhilSharmaWe: @LucasRoesler Could you please give an example of a pattern that we want to skip but not skipped by current filepath.Match(pattern, path) ?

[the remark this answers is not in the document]
NikhilSharmaWe: @LucasRoesler According to you "if the pattern contains no stars, test if the `path base == the pattern`"

Suppose pattern with no `*` is `example.txt` and path is `build/example.txt`. Path base for this case will be `example.txt`.

* If we use `filepath.Base(path) == pattern`, then path `build/example.txt` will match the pattern `example.txt`, which is the wrong behavior. 

* But if this case is handled by the `filepath.Match(pattern, path)`, in this case, the pattern will only be matched with the exact path given with the current implementation, which will be the correct behavior.

Could you please give an example of a pattern that we want to skip but is not taken into account by the current `filepath.Match(pattern, path)` implementation?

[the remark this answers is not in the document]
LucasRoesler: the reason i am thinking of this example 
```
.venv
__pycache__
*.pyc

logs/

.env
keyfile.json
.task
```

which is a mix of patterns and non-patterns. The `__pycache__` is something that could easily appear inside the template path and git would ignore it and i think we would want the watch to ignore it as well.

johnmccabe:
Have you looked at go-git - https://github.com/go-git/go-git/blob/dc2b346ed149080199d578f0190f3ac1156480c2/plumbing/format/gitignore/dir.go

commands/up.go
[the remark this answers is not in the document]
johnmccabe: Have you looked at go-git - https://github.com/go-git/go-git/blob/dc2b346ed149080199d578f0190f3ac1156480c2/plumbing/format/gitignore/dir.go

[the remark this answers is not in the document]
johnmccabe: Likewise, the git docs describe the format in detail - https://git-scm.com/docs/gitignore#_pattern_format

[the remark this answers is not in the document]
alexellis: Thanks for the comment John. I've sent you a Twitter DM if you still have an account and are interested in keeping in touch.

commands/up.go
alexellis: I think you're right about the context cancellation. We should definitely update go-execute for that. It hasn't seen changes because the API and requirements have been stable.

commands/up.go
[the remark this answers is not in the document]
NikhilSharmaWe: @LucasRoesler @johnmccabe Now I am using **"github.com/go-git/go-git/v5/plumbing/format/gitignore"** package for matching the patterns from `.gitignore`. Could you please take a look?

NikhilSharmaWe:
@alexellis I have resolved the conflicts that occurred due to the addition of `remote-builder`. 
Please inform me if this PR is what you expect or if something stills needs to be worked on.

commands/up.go
LucasRoesler: @NikhilSharmaWe and @alexellis i think we have to decide if this PR needs to include a change to allow context propagation to the build command or if we want to do that in a future change. 

go-execute is only really used once (that I can see) in this project. So, I would be in favor of just reverting to the stdlib here. Which would allow this PR to immediately update and add the required context propagation. Or updating once go-execute has context support. 

In either case, once we are building and cancelling a context based on the signal, the only other change i would like to see is on [L149](https://github.com/openfaas/faas-cli/pull/962/files#diff-5165889054d8170eac2603b4631a88d1addad1739d20c60d20a51949bffbaff5R149) when the watcher close can also watch the `ctx.Done()` channel for consistency.

LucasRoesler: Actually thinking about this more, we should just merge this and fix the context cancellation in a follow up.  The current state won't hurt anyone because it is an opt in feature

alexellis:
Thank you @NikhilSharmaWe 

Does this change also cover the `faas-cli local-run` feature, where we run the function in using docker installed on the host?

commands/up.go
alexellis: Are you volunteering for that @LucasRoesler ?

alexellis:
Hey guys, I think we may have a problem here.

I wanted to test this out and it just didn't work at all after editing a file, nothing happened?

https://asciinema.org/a/cHAZUrr3WPc95g7tjKrMN7xO6

Thoughts?

alexellis:
I can't see any `--watch` flag for faas-cli build either?

```
faas-cli build --watch
Unknown flag: --watch
```

One other thing we will need here is to generate a random or SHA-computed hash for the image tag.

In my stack.yml we have `latest`, and so the K8s cluster is never reloading the image when I do a manual push. Why ? It's set to IfNotLatest on the PullPolicy for lower cold-start times.

Issue #968, opened 2023-06-29
Issue with new feature faas-cli up --watch


alexellis:
I'm getting an issue with the new feature `faas-cli up --watch` feature by @NikhilSharmaWe and @LucasRoesler
 
## Expected Behaviour

Editing a file, whilst having run:

```
faas-cli up --watch
```

Should rebuild

## Current Behaviour

It didn't when editing handler.go with nano / vim.

## List All Possible Solutions and Workarounds

Do additional testing?

## Steps to Reproduce (for bugs)

See:

https://asciinema.org/a/cHAZUrr3WPc95g7tjKrMN7xO6

## Your Environment

Ubuntu Linux 
faas-cli 0.16.9

nitishkumar71:
I tried to reproduce this in my local with `faasd` on `multipass` using `ubuntu`. This scenario can only be produced if we are running `asciinema rec` to record  terminal output. If not, `watch` flag works correctly. I tried multiple times and each time without `asciinema rec` `watch` flag works correctly.

alexellis:
That's a really interesting finding. I'll take a quick look again now.

alexellis:
That's not working right now, or outside of tmux with vim. I did see it work once or twice then stopped.

alexellis:
It also seems like adding a new file isn't picked up because we are watching specific files instead of directories:

![image](https://github.com/openfaas/faas-cli/assets/6358735/3cddc39b-84d3-4fc6-82b2-91ddf6533403)

alexellis:
The code only looks for a Write:

But a Create should trigger this, as should a Delete/Remove.

Perhaps vim was doing a "trunctate" - i.e. Remove/Create and not the expected: "write"

```go

			select {
			case event, ok := <-watcher.Events:
				if !ok {
					return fmt.Errorf("watcher's Events channel is closed")
				}

				if event.Op == fsnotify.Write {
```

nitishkumar71:
> That's not working right now, or outside of tmux with vim. I did see it work once or twice then stopped.

Ahh, I was able to reproduce it now. There are 3 cases

1. with tmux using and edit in vim
2. open 2 terminal and run `watch` command in one and use  `vim` in another
3. use `asciinema rec` with tmux and edit in `vim`

In first case it works, but does not works for second and third. 

> But a Create should trigger this, as should a Delete/Remove.
Agree, these should be handled too.

alexellis:
This is what a save looks like from vim:

[...]

PR #969, opened 2023-06-30
Fixes for the --watch flag for faas-cli up

alexellis:
Fixes for the --watch flag for faas-cli up due to incomplete / defective feature PR.

## Description

* Fixes issue where all functions are rebuilt even if only one changed
* Fixes issue where unrelated files cause rebuilds
* Fixes issue where vim can't trigger and update due to using create
* Adds newly created folders into the watch when within a function's path
* Triggers rebuilds via stack.yml

## Motivation and Context

Fixes: #968

## How Has This Been Tested?

Tested with:

* creating a new folder within a function's handler
* editing a function's code multiple times with vim
* editing the stack.yml file
* `faas-cli new --append` during a watch
* two functions, only the changed one is rebuilt
* Out of scope for this PR - "copy extra directories" that can be specified for each function

## Types of changes

### DIFF
--- a/commands/up.go
+++ b/commands/up.go
@@ -9,12 +9,14 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"path"
 	"path/filepath"
 	"strings"
 	"syscall"
 	"time"
 
 	"github.com/bep/debounce"
+	"github.com/openfaas/faas-cli/stack"
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/go-git/go-git/v5/plumbing/format/gitignore"
@@ -82,8 +84,26 @@ func preRunUp(cmd *cobra.Command, args []string) error {
 
 func upHandler(cmd *cobra.Command, args []string) error {
 	handler := upRunner(cmd, args)
+
+	// Always run an initial build to freshen up
 	if err := handler(); err != nil {
 		return err
+	}
+
+	var services stack.Services
+	if len(yamlFile) > 0 {
+		parsedServices, err := stack.ParseYAMLFile(yamlFile, regex, filter, envsubst)
+		if err != nil {
+			return err
+		}
+
+		if parsedServices != nil {
+			services = *parsedServices
+		}
+	}
+
+	for _, service := range services.Functions {
+		fmt.Printf("Name: %s\n", service.Name)
 	}
 
 	if watch {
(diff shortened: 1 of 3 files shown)


alexellis:
cc @nitishkumar71 @nitishkumar71 @welteki

nitishkumar71 (changes requested):

commands/up.go
```diff
@@ -100,24 +123,30 @@ func upHandler(cmd *cobra.Command, args []string) error {
...
-				return watcher.Add(path)
+		for serviceName, service := range services.Functions {
+			handlerMap[serviceName] = path.Join(cwd, service.Handler)
+		}
+
+		for _, service := range services.Functions {
+			handlerPath := path.Join(cwd, service.Handler)
+
+			if err := addPath(watcher, handlerPath); err != nil {
+				return err
```
nitishkumar71, line 148: Although, I do understand these two loops are doing two different things, but as we are gonna exit on error in second loop. We should merge these two loops.

commands/up.go
```diff
@@ -82,10 +84,31 @@ func preRunUp(cmd *cobra.Command, args []string) error {
...
+		if err != nil {
+			return err
+		}
+
+		if parsedServices != nil {
+			services = *parsedServices
+		}
+	}
+
+	fnNames := []string{}
```
nitishkumar71, line 105: As, we need to replicate the capability into `local-run`. Possibly, we can wrap the entire watch code block and perform watch on the given list of functions.

nitishkumar71 (changes requested):

commands/up.go
```diff
@@ -131,13 +160,69 @@ func upHandler(cmd *cobra.Command, args []string) error {
 				if !ok {
 					return fmt.Errorf("watcher's Events channel is closed")
 				}
+				log.Printf("%s %s", event.Op, event.Name)
+
+				if strings.HasSuffix(event.Name, ".swp") || strings.HasSuffix(event.Name, "~") {
```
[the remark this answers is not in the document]
alexellis: I'll add it thanks

commands/up.go
alexellis: I don't think we can do them together because we have to exhaustively check all exact matches before looking for a fuzzy match

commands/up.go
alexellis: One note about local-run is that it only supports one function at a time right now.

commands/up.go
alexellis: You're right, that can be one loop.

commands/up.go
nitishkumar71: Yup. `fnNames` can be an array which can be used to filter functions which need to be watched.

alexellis:
I've added a note on why --watch won't work with localrun with the current design.

We need some kind of Go routine, cancellation/context approach so that the docker run which is blocking, doesn't block the watcher code - and the watcher code has a chance to kill the docker run for a prior version before starting a new build/run.

PR #1004, opened 2025-01-23
Pull templates from store when not present

alexellis:
## Description

Pull templates from store when not present

Implements a user-experience changes:

faas-cli new - when the templates folder is not present the language will be looked up in the template store using the default URL or an overriden one.

faas-cli publish - when no template folder is available then a pull will be carried out if a configuration section is present with a template given within in

Filtering - whilst there is no way to avoid pulling all the templates wihtin a repository, only the requested/needed ones are expanded by commands like faas-cli new.

## Motivation and Context

Improve user experience as discussed on community call Jan 22 2025

## How Has This Been Tested?

Tested mainly with existing-unit tests, faas-cli new / publish with and without a templates folder.

## Types of changes

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -5,7 +5,6 @@ package commands
 
 import (
 	"fmt"
-	"log"
 	"os"
 	"strings"
 	"sync"
@@ -155,6 +154,8 @@ func parseBuildArgs(args []string) (map[string]string, error) {
 
 func runBuild(cmd *cobra.Command, args []string) error {
 
+	templateName := "" // templateName may not be known at this point
+
 	var services stack.Services
 	if len(yamlFile) > 0 {
 		parsedServices, err := stack.ParseYAMLFile(yamlFile, regex, filter, envsubst)
@@ -178,7 +179,7 @@ func runBuild(cmd *cobra.Command, args []string) error {
 		}
 	} else {
 		templateAddress := getTemplateURL("", os.Getenv(templateURLEnvironment), DefaultTemplateRepository)
-		if pullErr := pullTemplates(templateAddress); pullErr != nil {
+		if pullErr := pullTemplates(templateAddress, templateName); pullErr != nil {
 			return fmt.Errorf("could not pull templates for OpenFaaS: %v", pullErr)
 		}
 	}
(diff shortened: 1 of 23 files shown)


commands/build.go
```diff
@@ -304,20 +305,19 @@ func build(services *stack.Services, queueDepth int, shrinkwrap, quietBuild bool
...
 
 // pullTemplates pulls templates from specified git remote. templateURL may be a pinned repository.
-func pullTemplates(templateURL string) error {
-	var err error
-	exists, err := os.Stat("./template")
-	if err != nil || exists == nil {
-		log.Println("No templates found in current directory.")
+func pullTemplates(templateURL, templateName string) error {
+
+	if _, err := os.Stat("./template"); err != nil && os.IsNotExist(err) {
```
alexellis, line 310: More idiomatic way to check for the folder.

commands/fetch_templates.go
```diff
@@ -20,21 +19,22 @@ const DefaultTemplateRepository = "https://github.com/openfaas/templates.git"
...
-func fetchTemplates(templateURL string, refName string, overwrite bool) error {
+func fetchTemplates(templateURL, refName, templateName string, overwrite bool) error {
 	if len(templateURL) == 0 {
 		return fmt.Errorf("pass valid templateURL")
 	}
 
 	dir, err := os.MkdirTemp("", "openfaas-templates-*")
 	if err != nil {
-		log.Fatal(err)
+		return err
```
alexellis, line 29: Better to return an error than use Fatal

commands/fetch_templates.go
```diff
@@ -56,7 +56,7 @@ func fetchTemplates(templateURL string, refName string, overwrite bool) error {
 		log.Printf("Cannot overwrite the following %d template(s): %v\n", len(preExistingLanguages), preExistingLanguages)
 	}
 
-	log.Printf("Fetched %d template(s) : %v from %s\n", len(fetchedLanguages), fetchedLanguages, templateURL)
+	fmt.Printf("Wrote %d template(s) : %v from %s\n", len(fetchedLanguages), fetchedLanguages, templateURL)
```
alexellis, line 59: No need to be using "log" package here in the CLI output

commands/template_pull_stack_test.go
```diff
@@ -3,49 +3,12 @@ package commands
...
 	"os"
 	"path/filepath"
-	"reflect"
 	"testing"
 
 	"github.com/openfaas/faas-cli/builder"
 	"github.com/openfaas/faas-cli/stack"
 )
 
-func Test_findTemplate(t *testing.T) {
```
alexellis, line 13: The findTemplate function was not used in any code path so was deleted

commands/build.go
rgee0: Agree.  

Would `os.Stat` return any errors other than  `os.IsNotExist`?

alexellis:
I'm going to merge this so it can get into the hands of a few people, we can make additional changes as we go along.

PR #1028, opened 2026-09-04
Set the default namespace with OPENFAAS_NS

welteki:
## Description

Add support for the `OPENFAAS_NS` environment variable as the default
namespace for namespace-aware commands (`up`, `deploy`, `describe`,
`diff`, `invoke`, `list`, `logs`, `ready`, `remove` and the `secret`
commands).

Namespace resolution precedence is unchanged: the `--namespace` flag
wins, then the stack.yaml namespace, then `OPENFAAS_NS`, then the
cluster default namespace.

`diff` previously listed deployed functions from a single namespace
only: the `--namespace` flag value, or the cluster default if the flag
was not set. For a stack that spans multiple namespaces this meant
deployed functions in other namespaces were invisible, so `diff` would
report them as not deployed. It now collects the set of namespaces
that the stack's functions resolve to (applying the same precedence)
and lists deployed functions from each of them, so every deployed
function that a stack function could match is found.

Also fixes `invoke` to use the resolved namespace on the 401 retry
path, which previously used the raw `--namespace` flag and dropped the
stack.yaml and `OPENFAAS_NS` values.

## Motivation and Context

Users with access to a non-default namespace, e.g. a per-user
namespace on a shared cluster, currently have to pass `--namespace`
to every command. Setting `OPENFAAS_NS` in the environment lets their
commands avoid implicitly targeting the cluster default namespace.

## How Has This Been Tested?

- Added unit tests for namespace precedence, env-substituted stack
  namespaces, logs resolution, and multi-namespace `diff` behavior.
- Covered same-named functions in different namespaces and API responses
  that omit the function namespace.
- Tested `OPENFAAS_NS` end to end with deploy, store deploy, list,
  describe, ready, invoke, logs, remove, and the secret lifecycle.
- Verified that `--namespace` overrides the environment and that existing
  stack namespace precedence is preserved.
- Tested `diff` with a stack spanning multiple namespaces. The old CLI
  missed namespaced functions, while the updated CLI matched each function
  to its queried namespace and reported only genuinely missing functions.

## Types of changes

## Checklist:

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -199,7 +199,7 @@ func runDeployCommand(args []string, image string, fprocess string, functionName
 
 			// Check if there is a functionNamespace flag passed, if so, override the namespace value
 			// defined in the stack.yaml
-			function.Namespace = getNamespace(functionNamespace, function.Namespace)
+			function.Namespace = getNamespace(functionNamespace, function.Namespace, os.Getenv(openFaaSNamespaceEnvironment))
 
 			fileEnvironment, err := readFiles(function.EnvironmentFile)
 			if err != nil {
@@ -319,7 +319,7 @@ Error: %s`, fprocessErr.Error())
 			tlsInsecure,
 			defaultReadOnlyRFS,
 			token,
-			functionNamespace,
+			getNamespace(functionNamespace, "", os.Getenv(openFaaSNamespaceEnvironment)),
 			cpuRequest,
 			cpuLimit,
 			memoryRequest,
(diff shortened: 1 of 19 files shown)


commands/deploy.go
```diff
@@ -199,7 +199,7 @@ func runDeployCommand(args []string, image string, fprocess string, functionName
 
 			// Check if there is a functionNamespace flag passed, if so, override the namespace value
 			// defined in the stack.yaml
-			function.Namespace = getNamespace(functionNamespace, function.Namespace)
+			function.Namespace = getNamespace(functionNamespace, function.Namespace, os.Getenv(openFaaSNamespaceEnvironment))
```
alexellis, line 202: I don't think I would have added `os.Getenv(openFaaSNamespaceEnvironment)` to every callsight - given it's consistent I would have perhaps not even have changed the signature at all and have had it self-contained.

Go unit tests can setenv dynamically when needed.

alexellis (approved):
Unsure on approach, but if it works, we'll ship it.

